PR #8, opened 2015-12-30
Adds aliases for function and return tags

topaxi:
See:
http://usejsdoc.org/tags-returns.html
http://usejsdoc.org/tags-function.html


gajus:
I have excluded aliases on purpose. I think it should be a configurable option. It could be a `prefer` map that allows to define parameter name preference over their alias, e.g.

```js
prefer: {
    'return': 'returns'
}
```

Aliases and the main name should not be used interchangeably in the same documentation.

topaxi:
Alright, makes sense to make this configurable :+1:

gajus:
@topaxi Please see https://github.com/gajus/eslint-plugin-jsdoc/issues/9

Issue #15, opened 2016-06-06
Allow passing additional valid tags to check-tag-name as an option


atakiel:
As jsdoc allows defining custom tags via plugins, it should be possible to give additional valid tag names to 'check-tag-name' rule as options.

Additionally, with such an option, users wouldn't have to wait for this plugin to be updated in order to be able to use tags added in a new version of jsdoc.

gajus:
I am Okay with this. Please raise a PR.

jensbodal:
Any chance you can take another look at the pull request which addresses this? 

https://github.com/gajus/eslint-plugin-jsdoc/pull/23

PR #26, opened 2016-11-09
Allow custom tag names jens

jensbodal:
This is in relation to this issue: https://github.com/gajus/eslint-plugin-jsdoc/issues/15

Somehow didn't notice you had already implemented an alternative solution by adding to the settings plus the additional unit tests.  I'll just leave this here if you're interested, and I can add test cases if you want to go this route instead.


jensbodal:
Leaving it to the other PR

PR #28, opened 2017-01-25
Support type name preference

LukasHechenberger:
The main aim of my fork is to support type name preference by setting `settings.jsdoc.typeNamePreference`.

E.g. if someone wants to document types as *Number* instead of *number* a `.eslintrc.json` could look like:

```json
{
  "settings": {
    "jsdoc": {
      "typeNamePreference": {
        "number": "Number"
      }
    }
  }
}
```

While implementing i stumbled across **various build issues**. I guess i fixed most of them, `npm run add-assertions` and `npm run generate-readme` now run without any errors.
The readme is still pretty messed up...


gajus:
> E.g. if someone wants to document types as Number instead of number a .eslintrc.json could look like:

The problem is that `Number` and `number` mean completely different things. It is very unlikely that you want to use `Number`.

You'd use number if you use [`Number` object wrapper](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number), e.g.

```js
/**
 * Thats `Number`.
 */
new Number('123');

/**
 * Thats `number`.
 */
123;

/**
 * Thats `number` too.
 */
Number('123');

```

I don't want to allow a preference setting, because it will most certainly going to be abused.

PR #32, opened 2017-04-14
fix: set node@4 as minimum

SimenB:
Also, don't test node@5, it's EOL

https://github.com/gajus/eslint-plugin-jsdoc/commit/8c9f89e5317f376a4f19c9c507ac10481d8c061c#commitcomment-21764339

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -8,7 +8,7 @@
       "env",
       {
         "targets": {
-          "node": 5
+          "node": 4
         }
       }
     ]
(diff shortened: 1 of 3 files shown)


SimenB:
@gajus thanks for the quick fix! (the tag isn't published, btw, but the new version is published to npm)

gajus:
https://github.com/gajus/eslint-plugin-jsdoc/tree/v3.0.1 ?

SimenB:
Ah! No update to package.json threw me 😄 

https://github.com/gajus/eslint-plugin-jsdoc/blob/master/package.json#L51

Issue #18, opened 2016-06-23
Error: Unsupported function signature format.


blake-newman:
> Error: Unsupported function signature format.
> 
> ```
> export default function testAction(
>   action, done, { args = [], mockThis = this, state = {}, mutations = [] } = {}
> ) {
> ```

This is actually valid es6 destructuring. I will try look into myself, but got a lot of work at the moment. Let me know if there is anything that i can help with 👍

risq:
Still having the issue on `3.0.2`. 
@gajus have you planned to support this syntax ?
Thanks !

Linked: https://github.com/gajus/canonical/issues/14, https://github.com/eslint/eslint/issues/2051 (`valid-jsdoc`)

brokentone:
Seems like this can be closed

PR #37, opened 2017-06-14
Fix the 'Unsupported function signature format.' error.

bennieswart:
This fixes the 'Unsupported function signature format.' error as per https://github.com/gajus/eslint-plugin-jsdoc/issues/18.

I cannot say that the testing for this has been rigorous, but it does seem to work for all basic cases. Then again, the fix is really simple and should always work, unless I'm missing something.

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,6 +277,13 @@ function quux (...args) {
 function quux ({a, b}) {
 
 }
+
+/**
+ * @param foo
+ */
+function quux ({a, b} = {}) {
+
+}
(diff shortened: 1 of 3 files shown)


sigorilla:
@gajus please approve this PR :)

bennieswart:
I've actually forgotten about this and I can provide an update on the situation.
I've been using this patch in a large project from the time the PR was made without issue.

gajus:
Thank you

PR #39, opened 2017-06-19
Correct check for non-ASCII descriptions

anton-rudeshko:
Simple regex check fails on non-ASCII descriptions.


anton-rudeshko:
ping

anton-rudeshko:
@gajus thanks!

PR #57, opened 2017-12-24
Fix issue with complete sentence fixer

bary12:
```
/**
 * foo
 */
```

was supposed to be fixed to 

```
/**
 * Foo.
 */
```

but actually was not fixed at all.
This PR solves that.


bary12:
@gajus it looks like the latest version of ESlint fails on src/rules/checkParamNames.js for some reason. I tried downgrading to 7.4.2 and it passed locally.

bary12:
The checks should pass after gajus/eslint-config-canonical#13 is merged.

PR #24, opened 2016-09-27
Support @inheritdoc in the require-param rule

theRobinator:
This fixes Issue #20.

See http://usejsdoc.org/tags-inheritdoc.html for the inheritdoc spec.

PR #74, opened 2018-05-16
Add an `allowOverrideWithoutParam` setting for `jsdoc/require-param`

### LINKED ISSUE #20: Rule 'require-param' does not respect @inheritdoc
According to the docs, the presence of @inheritdoc will cause jsdocs to ignore any other tags.  Since we are pulling in the docs from another node, then we should not have to re-declare the tags here as well.  This rule should check for the presence of the @inheritdoc and skip if found.  Same may also be true for @override, but I think that intended to work slightly different

### LINKED ISSUE #73: Need a way to get `require-param` to exempt `@override`
Hi @gajus, nice work in creating `eslint-plugin-jsdoc`! The [AMP project](https://www.ampproject.org/) just adopted it as our primary JSDoc checker, and it's working quite nicely for us. However, there's one really visible problem:

As described in https://github.com/ampproject/amphtml/issues/15297, the [Google JS style guide](https://google.github.io/styleguide/jsguide.html) prefers `@override` over `@inheritdoc`.

From [here](https://google.github.io/styleguide/jsguide.html#features-functions-parameters):

> Function parameters must be typed with JSDoc annotations in the JSDoc preceding the function’s definition, except in the case of same-signature `@override`s, where all types are omitted.

From [here](https://google.github.io/styleguide/jsguide.html#jsdoc-method-and-function-comments):

> Overridden methods must include all `@param` and `@return` annotations if any types are refined, but should omit them if the types are all the same.

And from [here](https://google.github.io/styleguide/jsguide.html#appendices-notes-about-standard-closure-compiler-annotations):

> `@inheritDoc` | Deprecated. Do not use. Use `@override` instead.

However, from looking at the source code of the `jsdoc/require-param` rule, it's clear that only `@inheritdoc` tags are exempted from documenting their `@param`s.

https://github.com/gajus/eslint-plugin-jsdoc/blob/52e3fcc46dd4087e50acf4a0c4f5683d51f5a699/src/rules/requireParam.js#L11-L14

I realize this is essentially a disparity between the JSDoc documentation and the Google style guide. However, we're seeing [hundreds of errors](https://github.com/ampproject/amphtml/search?l=JavaScript&q=%22%40override%22&type=Code) from using `@override` in our source code even though the Google style guide recommends it.

I was wondering if an exception could be made for `@override` in the `require-param` rule. Something gated by a setting would work too. Would you be willing to accept a PR that implements this?

Related to https://github.com/gajus/eslint-plugin-jsdoc/issues/20

rsimha:
This PR adds a new boolean setting called `allowOverrideWithoutParam` with a default value of `false`. 

- When the setting is `true`, the `@override` tag can be used without any accompanying `@param` tag(s).
- When the setting is omitted (or explicitly set to `false`), the original behavior remains, and `jsdoc/require-param` will flag an error when it finds an `@override` tag without the correct accompanying `@param` tags.

This was required due to the recommendations of the [Google JS style guide](https://google.github.io/styleguide/jsguide.html) ([here](https://google.github.io/styleguide/jsguide.html#features-functions-parameters), [here](https://google.github.io/styleguide/jsguide.html#jsdoc-method-and-function-comments), and [here](https://google.github.io/styleguide/jsguide.html#appendices-notes-about-standard-closure-compiler-annotations)). See #73 and https://github.com/ampproject/amphtml/issues/15297 for more context.

Fixes #73
Fixes #20 
Follow up to #24

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ JSDoc linting rules for ESLint.
     * [Settings](#eslint-plugin-jsdoc-settings)
         * [Alias Preference](#eslint-plugin-jsdoc-settings-alias-preference)
         * [Additional Tag Names](#eslint-plugin-jsdoc-settings-additional-tag-names)
+        * [Allow `@override` Without Accompanying `@param` Tags](#eslint-plugin-jsdoc-settings-allow-override-without-param)
     * [Rules](#eslint-plugin-jsdoc-rules)
         * [`check-param-names`](#eslint-plugin-jsdoc-rules-check-param-names)
         * [`check-tag-names`](#eslint-plugin-jsdoc-rules-check-tag-names)
@@ -143,6 +144,22 @@ Use `settings.jsdoc.additionalTagNames` to configure additional, allowed JSDoc t
             "additionalTagNames": {
                 "customTags": ["define", "extends", "record"]
             }
+        }
+    }
+}
+```
+
+<a name="eslint-plugin-jsdoc-settings-allow-override-without-param"></a>
+### Allow `@override` Without Accompanying `@param` Tags
+
+Use `settings.jsdoc.allowOverrideWithoutParam` to indicate whether the `@override` tag can be used without accompanying `@param` tag(s). The default value is `false`. The format of the configuration is as follows:
+
+```json
+{
+    "rules": {},
+    "settings": {
+        "jsdoc": {
+            "allowOverrideWithoutParam": true
         }
     }
 }
@@ -998,6 +1015,14 @@ function quux (foo, bar) {
 
 }
 // Message: Missing JSDoc @param "bar" declaration.
+
+/**
+ * @override
+ */
+function quux (foo) {
+
+}
+// Message: Missing JSDoc @arg "foo" declaration.
 ```
 
 The following patterns are not considered problems:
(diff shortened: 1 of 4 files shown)


rsimha:
Travis PR check is green.

I also tested the changes in this PR with [ampproject/amphtml: extensions/amp-hulu/0.1/amp-hulu.js](https://github.com/ampproject/amphtml/blob/master/extensions/amp-hulu/0.1/amp-hulu.js), which is supposed to have zero JSDoc lint errors.

**Before:**
```
[19:46:52] INFO: Running lint on extensions/amp-hulu/0.1/amp-hulu.js
[19:46:53] /usr/local/google/home/rsimha/src/amphtml/extensions/amp-hulu/0.1/amp-hulu.js
  38:3  warning  Missing JSDoc @param "layout" declaration  jsdoc/require-param

✖ 1 problem (0 errors, 1 warning)

[19:46:53] WARNING: Found 0 error(s) and 1 warning(s).
```

**After:**
```
[19:46:37] INFO: Running lint on extensions/amp-hulu/0.1/amp-hulu.js
[19:46:38] SUCCESS: No linter warnings or errors.
```

gajus:
Thank you

rsimha:
@gajus, thanks for the quick turnaround.

I was curious about 83ff68f6beeb9636424af570a4829c3d4a61a78d. Did I miss updating any files?

gajus:
> I was curious about 83ff68f. Did I miss updating any files?

The README.md document you've edited is generated using templates in:

https://github.com/gajus/eslint-plugin-jsdoc/tree/master/.README

Refer to https://github.com/gajus/gitdown for more information.

Issue #11, opened 2016-01-06
Support destructuring assignment (array, object, and array/object rest) on param and property checks


gajus:
Support destructuring assignment in function parameter.
- https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment
- https://github.com/codemix/babel-plugin-typecheck/issues/95

Example:

``` js
/**
 * @param {Object} coordinates
 * @param {number} coordinates.latitude
 * @param {number} coordinates.longitude
 * @returns {Promise}
 */
({latitude, longitude}) => {
}
```

gajus:
https://github.com/gajus/eslint-plugin-jsdoc#deconstructing-function-parameter

brokentone:
Is this still the case, that the AST does not provide sufficient hints for the names here? I don't see why this is fundamentally different than the various prop-type validation rules which consider destructures which have been added to the `eslint-plugin-react`.

gajus:
> Is this still the case, that the AST does not provide sufficient hints for the names here? I don't see why this is fundamentally different than the various prop-type validation rules which consider destructures which have been added to the eslint-plugin-react.

I would think that by now AST for this is available.

I am not using JSDoc myself anymore. Wouldn't investigate this myself, but will accept a PR.

brettz9:
I'd hope this could also support rest params...

Volem:
This is not the case anymore for below implementation.

```
/**
 * Verify TCKN from tckimlik.nvi.gov.tr
 * @param {object} userIdentity - Identity informations for user
 * @param {string} userIdentity.tckn - tckn of user
 * @param {string} userIdentity.name - name of user
 * @param {string} userIdentity.surname - surname of user
 * @param {number} userIdentity.birthYear - year of bearth
 * @returns {Promise<ResponseObject>} - Contains isVerified property on success. This is a boolen value 
 */
const verifyTCKN = async function ({ tckn, name, surname, birthYear }) {
```

This is not giving any error for me on version 8.4.2 of eslint-plugin-jsdoc

eslint parsing options

```
"parserOptions": {
        "sourceType": "module",
        "ecmaVersion": 8,
        "ecmaFeatures": {
            "experimentalObjectRestSpread": true
        }
    }
```
VSCode version
Version: 1.36.1 (user setup)
Commit: 2213894ea0415ee8c85c5eea0d0ff81ecc191529
Date: 2019-07-08T22:59:35.033Z
[...]

PR #83, opened 2018-08-21
[WIP] Improvement: Check param names at depth (destructure)

brokentone:
Fixes: https://github.com/gajus/eslint-plugin-jsdoc/issues/11
Prereq: https://github.com/gajus/eslint-plugin-jsdoc/pull/82

Meta: Certainly expect some conversation on this topic, not to be merged straight away. I am also still working on some edge cases.

This PR attempts to upgrade the "checkParamNames" rule to consider destructured props in a function argument, requiring the same level of jsdoc consistency that regular props require right now.

It accomplishes this by moving from the string versions of the jsdoc and function params to their most highly structured forms -- AST node for function params, and parsed array for jsdoc. We then build a pseudo AST for the latter, so that we can recurse into arbitrary depths of destructures.

One possible complaint at the moment would be that the AST building and the validation logic (and thus reporting) are comingled. It should be possible to separate these, but then there would need to be a step to compare the flat structure to the tree... seems not worth the effort.

I know there are currently issues with static definitions (though I don't know why...), and I haven't tested on greater depth or with combinations of destructures and either renames or defaults. All of which I'd consider requirements for this to land. Additionally, I would expect this to either need to hide behind an opt-in flag as a possible minor (though with this substantial of a rewrite, I can't guarantee perfect parity), OR a major release (still possibly flagged).

### DIFF
--- a/src/jsdocUtils.js
+++ b/src/jsdocUtils.js
@@ -1,26 +1,32 @@
 import _ from 'lodash';
 import tagNames from './tagNames';
 
+const getFunctionParameterName = (param : Object) : string => {
+  if (_.has(param, 'name')) {
+    return param.name;
+  }
+
+  if (_.has(param, 'left.name')) {
+    return param.left.name;
+  }
+
+  if (param.type === 'ObjectPattern' || _.get(param, 'left.type') === 'ObjectPattern') {
+    return '<ObjectPattern>';
+  }
+
+  if (param.type === 'RestElement') {
+    return param.argument.name;
+  }
+
+  if (param.type === 'ObjectPattern' || param.type === 'Property') {
+    return param.key.name;
+  }
+
+  throw new Error('Unsupported function signature format.');
+};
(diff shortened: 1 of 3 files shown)


gajus:
Ping me when this is ready to merge.

gajus:
@brokentone Hi, are you planning to finish this task?

brokentone:
@gajus indeed I am planning to -- I've been doing a few things:

1. Been living with this rule in my company's large codebases, which has been useful, I was able to find a few other tokens which needed to be considered.
2. This rule increase make this error in eslint much more annoying: https://github.com/eslint/eslint/pull/10792 Which... as you can see the comments on that PR, it's a pretty challenging solve actually.

I can wrap this as is and you can release as a major with caveats?

gajus:
@brokentone I am closing this PR. If you want to finish it and merge it, push changes and ping me.

Issue #84, opened 2018-08-21
Question: TemplateLiterals


brokentone:
Noticing in your [canonical eslint config](https://github.com/gajus/eslint-config-canonical/blob/master/configurations/eslintrc.json#L447-L450) a restriction on using `TemplateLiteral`s. Not here to debate JS style, but I feel that literals would provide cleaner error report messages. Have opinions changed on their usage, or still verboten?

gajus:
Can you provide a use case example?

I do use template literals often in specific cases, e.g. describing test cases or as a way to escape text (https://github.com/gajus/slonik#tagged-template-literals). However, I never use it for simple string interpolation.

PR #85, opened 2018-08-22
Streamline report messages

brokentone:
Just for discussion of: https://github.com/gajus/eslint-plugin-jsdoc/issues/84

Certainly unnecessary, but I do feel that the shorter phrases are every so little more ergonomic than the concatenated ones.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,6 +2,8 @@
   "extends": "canonical",
   "root": true,
   "rules": {
-    "filenames/match-regex": 0
+    "filenames/match-regex": 0,
+    "prefer-template": 2,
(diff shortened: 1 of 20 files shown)


gajus:
Sorry, you misunderstood me. As I said, I never use it for simple string interpolation.

I am only using template literals to create multi-line strings that are not for end-user consumption, e.g. such as in constructing the inputs in test cases (https://github.com/gajus/eslint-plugin-flowtype/blob/3cd70b83e009bfdff394d129c00dd5df9cbc712e/tests/rules/assertions/sortKeys.js#L50-L275).

PR #102, opened 2018-11-01
Rules config

brettz9:
Just thought I'd get my feet wet with some minor linting type changes before looking at adding/tweaking any rules...

### DIFF
--- a/README.md
+++ b/README.md
@@ -1196,6 +1196,13 @@ function quux () {
 function quux () {
 
 }
+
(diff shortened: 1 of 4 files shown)


brettz9:
Any ETA on when you might be able to take a look at these PRs? Thanks!

gajus:
I have. The reason I did not merge it is because it is a dirty PR.

Multiple changes, not related to one another.

PR #104, opened 2018-11-01
Without param

### LINKED ISSUE #100: Avoid flagging `require-param` with @implements present (and possibly also @augments/@extends)
For the `TransparentColor` function in this example:

```js
/**
 * @interface
 * @param {string} name The color name
 * @returns {object} A color object
 */
function Color (name) {}

/**
 * Function representing a transparent color.
 * @implements {Color}
 */
function TransparentColor (a) {}
```

...eslint-plugin-jsdoc's `jsdoc/require-param` rule complains with:

> Missing JSDoc @param "a" declaration

However, the [JSDoc (for @implements)](http://usejsdoc.org/tags-implements.html) states:

> If you do not document one of the symbols in the implementation, JSDoc will automatically use the interface's documentation for that symbol.

This would seem to suggest that with `@implements` present, it is expected that values need only be added if they differ from the interface they are implementing.

While JSDoc even states:

> You do not need to add the @implements tag to each member of the implementation (for example, the implementation's instance methods).

...it would be helpful if eslint-plugin-jsdoc could at least avoid the error when `@implements` is explicit on the block. In SVGEdit, we have a number of these real-world uses which are getting unduly flagged, and it would be a maintenance head-ache to have to duplicate parameter and return information on implementing objects as well as on the interface.

([\@augments](http://usejsdoc.org/tags-augments.html) (or its synonym `@extends`) should probably behave this way as well. I see you already support this optionally with [@override](http://usejsdoc.org/tags-override.html).)

----

Btw, I also filed an issue against jsdoctypeparser (at https://github.com/Kuniwak/jsdoctypeparser/issues/47 ) to allow "event:" within types as per the JSDoc spec--something I hope can be fixed so it can be surfaced through eslint-plugin-jsdoc (or make a fork if necessary if jsdoctypeparser is not being maintained).

And, FWIW, I've filed an issue at https://github.com/AtomLinter/linter-eslint/issues/1192 for errors reported specifically for eslint-plugin/jsdoc rules--not sure if this is actually your issue or not.

brettz9:
- Enhancement: Add new boolean options `allowImplementsWithoutParam` and `allowAugmentsExtendsWithoutParam` and apply to `requireParam` (fixes #100)

Builds on PR #102 

(FWIW, I logged a similar issue at https://github.com/eslint/doctrine/issues/222 but that project is no longer being maintained by ESLint.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1196,6 +1196,13 @@ function quux () {
 function quux () {
 
 }
+
+/**
+ * Hello: world.
+ */
+function quux () {
+
+}
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -1350,6 +1350,82 @@ function quux (foo) {
 
 }
 // Message: Missing JSDoc @param "foo" declaration.
+
+/**
+ * @implements
+ */
+function quux (foo) {
+
+}
+// Message: Missing JSDoc @param "foo" declaration.
+
+/**
+ * @augments
+ */
+function quux (foo) {
+
+}
+// Message: Missing JSDoc @param "foo" declaration.
+
+/**
+ * @extends
+ */
+function quux (foo) {
+
+}
+// Message: Missing JSDoc @param "foo" declaration.
+
+/**
+ * @override
+ */
+class A {
+  /**
+    *
+    */
+  quux (foo) {
+
+  }
+}
+// Message: Missing JSDoc @param "foo" declaration.
+
+/**
+ * @implements
+ */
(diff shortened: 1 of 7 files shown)


gajus:
Basically, all the PRs are dirty, because they extend one another and do not provide isolated changes.

gajus:
I think the easiest way to resolve this is to close all the existing PRs are start a new.

PR #105, opened 2018-11-01
Array pattern

brettz9:
Builds on PR #104 to fix #94 .

Issue #101, opened 2018-10-31
Full ESLint on JavaScript within @example


brettz9:
How easy would it be to add support for linting @example code (e.g., as with Github fenced blocks flagging "js" as the language)?

I see from the JSDoc source that "prettyprint" is added for @example. (Highlighting also applies for source and details (of methods and such), but this should not be needed again for user code; there is also the markdown highlighter though I think that code only applies to the tutorials and such that it can embed, and not fenced blocks or such.)

It'd be a bonus if there could also be the ability to indicate, either with an object with rules, a separate config file, or with the rules of the same file applied but with specific "overrides" applied (e.g., so that the examples within JSDoc comments may follow the same looser rules as say one's ".md" files as opposed to ".js" files)?

Would this be doable?

gajus:
It is definitely doable.

Seems very esoteric, though.

I would accept a PR, but thats not something I would attempt myself.

PR #106, opened 2018-11-14
Check examples

brettz9:
Builds on previous PRs (most recent: #105 ).

Implementation for #101.

I still need to remove logging and add documentation, but wanted to get your understanding on the final desired API.

Specifically, while it looked like you were favoring use of `settings`, I wasn't sure whether my changes really merited being project-wide settings as opposed to just offering options specific to my rule (`check-examples`).

I could probably do with some better config naming too.

And I'd still like to add some minor config (for `reportUnusedDisableDirectives` and `allowInlineConfig`), but thought I'd wait until hearing about where you wanted the options.

But the code does seem to be working, both in terms of passing tests and in my use of my fork on svg-edit, including the change to allow reporting of starting column info.

### DIFF
--- a/.README/rules/check-param-names.md
+++ b/.README/rules/check-param-names.md
@@ -25,4 +25,6 @@ function quux ({
 }
 ```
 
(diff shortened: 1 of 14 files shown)


gajus:
https://github.com/gajus/eslint-plugin-jsdoc/pull/104

PR #111, opened 2018-12-07
Rulescfg

brettz9:
- Refactoring: Remove outdated `rulesConfig`: https://eslint.org/docs/2.0.0/user-guide/migrating-to-2.0.0#plugins-no-longer-have-default-configurations

### DIFF
--- a/README.md
+++ b/README.md
@@ -1196,6 +1196,13 @@ function quux () {
 function quux () {
 
 }
+
+/**
+ * Hello: world.
+ */
+function quux () {
+
+}
 ```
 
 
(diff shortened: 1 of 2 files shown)


brettz9:
I see this is failing because of your having this line:

```js
import JSDOC_3_TAGS from './jsdoc3Tags';
```

...in `test/rules/assertions/checkTagNames.js` without its own `// eslint-disable-next-line id-match`.

Do you want me to add such a comment above the `import` as it is for the line after it, or ?

gajus:
https://github.com/gajus/eslint-plugin-jsdoc/commit/3e4c84f26f7e3306394d38fdd7ce06beca4713ba

Linting issues fixed in separate commits.

PR #112, opened 2018-12-07
package.json linting

brettz9:
- Linting (package.json): Add linter-recommended contributors, homepage, bugs (used on npmjs.com)

I'm not itching to get myself added as a contributor, but the linter linter-package-json-validator flags it as a recommended field. It probably helps though to advertise more contributors. If nothing else, it would be nice to have it as an empty array so it can stop flagging things...

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,9 @@
     "name": "Gajus Kuizinas",
     "url": "http://gajus.com"
   },
+  "contributors": [
+    "Brett Zamir"
+  ],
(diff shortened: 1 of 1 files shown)


gajus:
No intention of adding new fields to `package.json`.

Your contributions are visible in commits section of the project.

`contributors`, `homepage` and `bugs` are not required fields.

brettz9:
They are used on npmjs.com but your call obviously...

PR #113, opened 2018-12-07
Test expecting space

brettz9:
- Fix failing test by adding missing space (as added by rule along with newline)

### DIFF
--- a/src/rules/newlineAfterDescription.js
+++ b/src/rules/newlineAfterDescription.js
@@ -35,7 +35,7 @@ export default iterateJsdoc(({
         });
 
         // Add the new line
-        sourceLines.splice(lastDescriptionLine + 1, 0, indent + ' * ');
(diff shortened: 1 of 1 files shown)


gajus:
This extra space is going to be stripped by rules declared in `.editorconfig`.

The fix should happen in the rule, i.e. space should not be required.

brettz9:
I guess you mean the space should not be added in the rule rather than not be "required" in the rule. 

I've amended and force-pushed such a change.

gajus:
> I've amended and force-pushed such a change.

Whats the reason you are using force push?

gajus:
Thank you

brettz9:
Sure, my pleasure... I force-pushed as I recall getting rejected from pushing with amended commits (and you'd otherwise have my original commit, undoing that commit, and the second commit all together)...

PR #118, opened 2018-12-07
Peer (dev) deps

brettz9:
- Add expected peer dependencies for eslint-config-standard (through eslint-config-canonical) (warnings showing during local install)


gajus:
Fixed in https://github.com/gajus/eslint-config-canonical/commit/b6e3d096212e6edc6033a91c3a79509d01579232

PR #119, opened 2018-12-08
`check-examples` rule

### LINKED ISSUE #101: Full ESLint on JavaScript within @example

brettz9:
- feat: Add `check-examples` rule to lint JavaScript within `@example` (fixes #101);
    with settings to:
    1. require JSDoc-spec'd `<caption>` at beginning of `@example`
    2. regex settings to whitelist and blacklist lintable examples
    3. setting for a dummy file name to trigger specific rules defined in one's
         config; usable with ESLint `overrides`->`files` globs, to apply a desired
         subset of rules with `@example` (besides allowing for rules specific to
         examples, can be useful for using same rules within `@example` as
         with JavaScript Markdown lintable by other plugins)
    4. Other settings for specifying config (base config object or an `.eslintrc`
         config file; checks normal `.eslintrc` by default unless
         `eslintrcForExamples` setting is set to `false`)
    5. Provides some defaults for suppressing reporting of rules which are
         likely to be troublesome in example files unless
         `noDefaultExampleRules` setting is `true`
- Report column (for checkExamples and any that report it)

Regarding your [comment](https://github.com/gajus/eslint-plugin-jsdoc/issues/101#issuecomment-434608259) about being open to a PR but the need being esoteric, in my opinion, this could actually be the single most important doc item to lint for, in order to ensure the examples indicated for users of the library do not have inadvertent bugs.

Thanks!

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -14,6 +14,7 @@ This table maps the rules between `eslint-plugin-jsdoc` and `jscs-jsdoc`.
 
 | `eslint-plugin-jsdoc` | `jscs-jsdoc` |
 | --- | --- |
+| [`check-examples`](https://github.com/gajus/eslint-plugin-jsdoc#eslint-plugin-jsdoc-rules-check-examples) | N/A |
 | [`check-param-names`](https://github.com/gajus/eslint-plugin-jsdoc#eslint-plugin-jsdoc-rules-check-param-names) | [`checkParamNames`](https://github.com/jscs-dev/jscs-jsdoc#checkparamnames) |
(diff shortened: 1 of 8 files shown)


gajus:
Looks good. Can you please contribute the documentation for this rule?

brettz9:
I've made one small change in my amendment (in addition to the document additions) to ensure `reportUnusedDisableDirectives` always applies.

PR #126, opened 2018-12-09
- Add `allowInlineConfig` and `reportUnusedDisableDirectives`

brettz9:
Thanks very much for all the reviews!

I realized after the fact that I should probably add this config as well. One [can't currently use multi-line comments](https://github.com/jsdoc3/jsdoc/issues/821) in JSDoc, and I haven't added any handling for them, but these will at least work with single-line comment config and allows for disabling. Admittedly inline config would probably be pretty unusual within examples anyways, but figured it should be exposed.

Also has two very minor clarifications in the README.

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -158,7 +158,7 @@ The format of the configuration is as follows:
 ### Settings to Configure `check-examples`
 
 The settings below all impact the `check-examples` rule and default to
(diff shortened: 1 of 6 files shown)


gajus:
Thank you for your contributions.

PR #130, opened 2018-12-13
Fixed @examples with multiple lines breaking markdown

ibbignerd:
The \``` in the `@example`s that are multiline broke the markdown for the readme. Indenting to 4 spaces makes markdown see the \``` as code and just indents it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,8 +309,8 @@ function quux () {
 
 /**
  * @example ```js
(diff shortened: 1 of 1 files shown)


ibbignerd:
Resolved with https://github.com/gajus/eslint-plugin-jsdoc/pull/140

PR #131, opened 2018-12-13
Improve markdown links to use header tags

ibbignerd:
The empty `<a>` tags aren't needed for anchors. 

https://stackoverflow.com/a/45508928/10559113


gajus:
These links are generated using https://github.com/gajus/gitdown.

Issue #135, opened 2018-12-17
Defining the properties of values in an array warns when using an example from the JSDoc docs


matchai:
When implementing an example of JSDoc usage directly from the JSDoc docs, a `jsdoc/check-param-names` warning is reported.

<sub>The "Documenting properties of values in an array" heading from http://usejsdoc.org/tags-param.html</sub>
```js
/**
 * Assign the project to a list of employees.
 * @param {Object[]} employees - The employees who are responsible for the project.
 * @param {string} employees[].name - The name of an employee.
 * @param {string} employees[].department - The employee's department.
 */
Project.prototype.assign = function(employees) {
    // ...
};
```

The fully reported warning is as follows:
`@param path declaration ("employees[].name") root node name ("employees[]") does not match previous real parameter name ("employees"). [jsdoc/check-param-names]`

This warning is dismissed if the `[]` is removed from each of the param titles.

gajus:
PRs welcome. Seems like a bug.

matchai:
Looks like this is ready to be closed (looks like I used the "Closes" keyword incorrectly 😅)

thernstig:
@brettz9 @gajus This seems to have stopped working. I am using version 32.2.0.

brettz9:
What you are reporting is really a dupe of #596 .

I'd accept a PR to restore the behavior addressed back in #136 but the root of the issue is that this needs to be resolved in our type/namepath parser.

Note that I have eyes on https://github.com/simonseyock/jsdoc-type-pratt-parser for a potential replacement of our jsdoctypeparser, but it is not ready yet and it would need refactoring and testing.

We might want to ask if that parser has plans to support this type--or rather "namepath" in case we do end up incorporating it.

But please let's keep further discussions over on #596 .

PR #136, opened 2018-12-19
Allow for array value props to be defined

matchai:
Solves: #135.
This is by no means a perfect solution, but was a simple solution that wouldn't require considerable refactoring. This implementation allows for array value props to be defined.

A better solution would be one that checks that the `lastRealParameter` has the `Object[]` type.

Thank you for the excellent plugin! 👍


gajus:
Can you summarise the limitations of the current solution vs using `lastRealParameter`?

Issue #139, opened 2018-12-31
Add check-examples usage examples


gajus:
check-examples usage examples have been removed because the use of ``` is breaking the README.md markdown, see https://github.com/gajus/eslint-plugin-jsdoc/commit/dc5e015477a04d69f31d07ca09b3978d190d49ed.

MuhamadOmr:
hey @gajus this is my first pull request to open source
glad i could help here #140

kindly review , if there any changes

PR #140, opened 2019-01-04
fix: checkExample code with backticks, error messages, assertion link

MuhamadOmr:
fixes: #139


gajus:
Thank you @MuhamadOmr

Issue #145, opened 2019-01-10
Typescript types should be supported in jsdoc type positions


sandersn:
Typescript types should be supported. Most of the work needs to occur in ~~Kuniwak/jsdoctypeparser~~ (*now*, [jsdoc-type-pratt-parser](https://github.com/jsdoc-type-pratt-parser/jsdoc-type-pratt-parser))), but there will likely be some code needed here as well. At the very least, some tests need to be added to make sure that Typescript types work.

Related:

- ~~#133 - support for typeof, which is understood by every JS type checker, not just Typescript. Kuniwak/jsdoctypeparser#50 - the bug tracking the actual fix~~
- [#96](https://github.com/jsdoc-type-pratt-parser/jsdoc-type-pratt-parser/issues/96) - The bug tracking the fix for TS in general

gajus:
PRs welcome.

sandersn:
I have an initial PR, for typeof and import types, up at Kuniwak/jsdoctypeparser#51. I'll create a matching PR here, but it'll depend on a new version of jsdoctypeparser being available.

sandersn:
I like the jsdoctypeparser parser code itself, although it is quite old and hasn't been touched in 2.5 years or so. Maybe it just needs more maintainers. 

Its licence is MIT, which I think is compatible with that of eslint-plugin-jsdoc, so copying the source would just be a technical problem, not a legal one.

MaxMilton:
This would be amazing. I could finally get rid of all the `/* eslint-disable jsdoc/valid-types */` comments in files which contain TS types (which is a lot scattered throughout many projects).

sandersn:
I haven't heard back from my PR on jsdoctypeparser, so I think moving the code here is a good idea. I'll take a look when I have time, which may be a few weeks.

ExE-Boss:
Also, this should support allowing `object` over `Object`, which is&nbsp;recommended by&nbsp;TypeScript and&nbsp;enforced by&nbsp;my&nbsp;config.

sandersn:
@ExE-Boss I think that would be determined by `rules/validTypes.js`, since both types parse correctly in jsdoctypeparser.

Also, for Typescript, I don't recommend putting types in JSDoc since the types should be in the signature.

sandersn:
[...]

Issue #133, opened 2018-12-14
`typeof C` type expression not supported


arv:
Closure, TypeScript and Flow all support `typeof C` as a type expression.

```js
/**
 * @param {typeof C} constr
 * @return {C}
 */
function construct(constr) {
  return new constr();
}
```

sandersn:
eslint-plugin-jsdoc relies on jsdoctypeparser to parse jsdoc types. I just opened [an umbrella issue for parsing all Typescript types there](https://github.com/Kuniwak/jsdoctypeparser/issues/50).

I'll link the PR I have in the works once it's ready. It adds `typeof` and `import("./package")` types.

sandersn:
Fixed by #146

PR #146, opened 2019-01-10
Test typeof and import types from Typescript

sandersn:
This PR shows that #133 is fixed, and that #145 is starting to be addressed.

However, it depends on Kuniwak/jsdoctypeparser#51 being merged and a new version of that package being published.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1016,7 +1016,7 @@ function quux (foo) {
 }
 // Settings: {"jsdoc":{"additionalTagNames":{"customTags":["baz","bar"]}}}
 
-/**
+/** 
  * @abstract
  * @access
  * @alias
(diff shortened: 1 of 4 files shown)


sandersn:
OK, this PR is ready to go. There are a few things I don't understand:

1. A lot of edits in README are unrelated to my change. Maybe I made a mistake when merging from master?
2. checkIndentation failed lint until I added a blank line. Did master have lint on it when I merged?

The actual code change is 4 tests plus an update to package.json, so maybe it would be better just to create a new PR and close this one.

brettz9:
I won't voice my objections to this apparent break from current usejsdoc.org support if movement toward Typescript is going on there (as it is the original and I think still authority for "JSDoc").

However, I just wanted to mention that if it will take a while for decision on this PR for some reason, I'd hope for the jsdoctypeparser dep. to be updated anyways so its support for `event:` (or other improvements) can be surfaced.

Issue #142, opened 2019-01-07
"require-returns" requires useless return statements for return type "undefined"


lll000111:
Subject says it. If the return type is undefined I don't need a return statement in the code.

This plugin should concern itself with JSDoc &mdash; not with the code style! The return type when there is no `return` statement is `undefined`, so describing such a function as having `@returns {undefined}` is 100% correct. I should not get

    Present JSDoc @returns declaration but not available return expression in function  jsdoc/require-returns

gajus:
PR is welcome.

lll000111:
Sure, always. And so are big reports, and here is one.

ibbignerd:
[According to ESLint](https://eslint.org/docs/rules/valid-jsdoc), `@returns {void}` should be valid.

[According to Google Closure Compiler](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#return-type-description), omitting it completely or `@returns {undefined]` should be valid. The following should all be tests that satisfy these requirements.

```javascript
valid: [
    {
      code: `
          /**
           * @returns {Object}
           */
          function quux () {

            return {a: foo};
          }
      `
    },
    {
      code: `
          /**
           * @returns {Object}
           */
          const quux = () => ({a: foo});
      `
    },
    {
      code: `
          /**
           * @returns {Object}
           */
          const quux = () => {
            return {a: foo}
          };
      `
    },
    {
      code: `
          /**
           * @returns {void}
           */
          function quux () {
          }
      `
    },
    {
      code: `
          /**
           * @returns {void}
           */
          const quux = () => {
            
          }
      `
    },
    {
      code: `
          /**
           * @returns {undefined}
           */
          function quux () {
          }
      `
    },
    {
      code: `
          /**
           * @returns {undefined}
           */
          const quux = () => {
            
          }
      `
    },
    {
      code: `
          /**
           *
           */
          function quux () {
          }
      `
    },
    {
      code: `
          /**
           *
           */
          const quux = () => {
            
          }
      `
    }
  ]
```

ChrisHSandN:
It's no PR, but might be helpful. This is the change I made to work around this bug and get my code validating:

https://github.com/gajus/eslint-plugin-jsdoc/blob/master/src/rules/requireReturns.js

```javascript
// replace
if (JSON.stringify(jsdocTags) !== '[]' && sourcecode.indexOf('return') < 1) {
// with
[...]

PR #147, opened 2019-01-11
[Feature] [Fix] split require-returns in two rules and fix bug #142

ptkdev:
**Feature**: require-returns-check
**Fix**: require-returns
**Description**: require-returns-check: check if exist return expression in function and in comment. Need exist both.
**Test**: passed and writed
**Documentation**: writed

**Fix bug**: #142 (Thx @ChrisHSandN)


ptkdev:
Please explain me how send pull request without add README.MD file. 
pre-commit generate readme.md and git add automatically, i don't kown why.

PR #167, opened 2019-03-13
feat: make require-returns-description accept void functions (fixes #157)

### LINKED ISSUE #157: `require-returns-description` requires description even when type is `undefined`
Unlike ESLint's built-in `valid-jsdoc` rule, `require-returns-description` requires a description even when the type is `undefined`.

Please allow for no description if return type is `undefined`. Otherwise, what are we supposed to put there?

golopot:
Fixes  #157 .

If the return type is `{void}` or `{undefined}`, return description is not required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2146,6 +2146,20 @@ function quux () {
 function quux () {
 
 }
+
+/**
+ * @returns {undefined}
+ */
+function quux () {
+
+}
+
+/**
+ * @returns {void}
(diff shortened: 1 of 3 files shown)


eXigentCoder:
Hey there, just a quick update, it seems to still be an issue if your return type is `{Promise<undefined>}` or `{Promise<void>}`

PR #175, opened 2019-03-19
fix: requireReturnsCheck false positve on async functions (fixes #174)

### LINKED ISSUE #174: `require-returns-check` - async function always returns promise
So, following is false positive:

```js
/**
 * @returns {Promise}
 */
async foo() {
   await boo();
} 
```

golopot:


gajus:
Thank you

PR #183, opened 2019-03-31
Makes "Check Returns" ignore abstract methods.

thsmi:
Abstract methods are typically completely documented as their children inherit their documentation by default. But their implementation is normally a single thow statement.

A perfect example can be found in the JSDoc Documentation:
http://usejsdoc.org/tags-abstract.html

The current eslint-plugin-jsdoc implementation fails in this case. It complains about a missing return statement. 

I also reordered the code and removed code which looks to me obsolete. I could not think of any example where the check for undefined or void code could kick into action. There is also no unit test which would cover this branch.

Futhermore I added a test which ensures at most one returns parameter is specified.

### DIFF
--- a/src/rules/requireReturnsCheck.js
+++ b/src/rules/requireReturnsCheck.js
@@ -1,4 +1,3 @@
-import _ from 'lodash';
 import iterateJsdoc from '../iterateJsdoc';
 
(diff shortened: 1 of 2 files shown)


gajus:
Thank you

PR #184, opened 2019-04-04
feat: add fixer for check tag names

vulkd:


src/tagNames.js
```diff
@@ -71,7 +71,8 @@ export default {
   package: [],
   param: [
     'arg',
-    'argument'
+    'argument',
+    'parameter'
```
brettz9: `parameter` is not a recognized tag or alias of jsdoc. It is not at https://jsdoc.app/tags-param.html , and I can find no reference to it in the repository: https://github.com/jsdoc/jsdoc/search?utf8=%E2%9C%93&q=parameter&type= searching all files (except test files).

brettz9: I removed it--but just thought I'd mention it in case there was something I was missing, like it being in use in other syntaxes.

Issue #162, opened 2019-03-01
Rule to require JSdoc comment


philsch:
Like described in the deprecated package https://eslint.org/docs/rules/require-jsdoc, it should be possible to require that a JSdoc is existing at all.

gajus:
PR welcome. Simple first contribution.

thsmi:
I've implemented an eslint compatible requireJSDoc rule in my fork.
The change set is 54445b6d20fbc067e52cd522d896930a3432c2ec

The rule is re-engineered by running it against the ESLint Unit tests.

But before creating a pull request I need clarification on the following topic:

As this project does not uses any of the well known OSS licenses (Honestly it does not seem to me as a valid license at all) I do not know if it is permissible to include the copied unit test file. The rule file is reengineered, so it does not have any licensing issues.

Can you, as the maintainer, contact the eslint maintainer and ask for clarification? If they grant you the permission everything should be fine.

In case they cannot grant you the permission, the only way to include the unit test would be getting the permission of all the original authors to convert it from MIT license to the public domain or your homegrown license. Or including the Unittest file under the MIT License. Both is to my knowledge viable.  

So how to proceed? 

Should I create a pull request without the unit test file? Or contribute the unit test under MIT? Or wait until you clarified the situation with the eslint maintainer?

brettz9:
`package.json` lists this package as `BSD-3-Clause` which it appears to me to be... And with eslint/eslint being under MIT, if that's the code you included, then it should be compatible in this direction (to include within `eslint-plugin-jsdoc`).

But, FWIW, if including `eslint-plugin-jsdoc` into an MIT project, the project could maintain MIT for the rest of its files, but would need to follow the separate and slightly stricter BSD-3-Clause terms for the `eslint-plugin-jsdoc` code.

thsmi:
This raises a new question. Which is the actual license of this repository?

  * Github shows a custom license as defined in https://github.com/gajus/eslint-plugin-jsdoc/blob/master/LICENSE
  * NPM shows BDS as it is defined in the package.json

brettz9:
It is the same text as https://opensource.org/licenses/BSD-3-Clause , with two exceptions (the second of which is puzzling here, but I presume is due to a copy-paste error from another repo):

1. "Gajus Kuizinas (http://gajus.com/)" replaces "the copyright holder" as would be normal or expected.
[...]

PR #188, opened 2019-04-06
Implement a requireJSDoc compatible rule

thsmi:
Implements a requireJSDoc compatible rule as described in issue #162.

The unit test file is ported from eslint and therefore MIT licensed for more details refer also to issue #162

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,6 +18,7 @@ import requireReturnsCheck from './rules/requireReturnsCheck';
 import requireReturnsDescription from './rules/requireReturnsDescription';
 import requireReturnsType from './rules/requireReturnsType';
 import validTypes from './rules/validTypes';
+import requireJSDoc from './rules/requireJSDoc';
 
 export default {
   configs: {
@@ -33,6 +34,7 @@ export default {
         'jsdoc/require-description-complete-sentence': 'off',
         'jsdoc/require-example': 'off',
         'jsdoc/require-hyphen-before-param-description': 'off',
+        'jsdoc/require-jsdoc': 'warn',
         'jsdoc/require-param': 'warn',
         'jsdoc/require-param-description': 'warn',
         'jsdoc/require-param-name': 'warn',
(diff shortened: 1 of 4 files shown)


EdieLemoine:
Tests are failing, could you rename the rule to `require-js-doc` and rename the related files from `requireJSDoc` to `requireJsDoc`?

allthesignals:
I can't seem to rename the file. The requirement that each commit pass the test suite also prevents some of the [workarounds](https://stackoverflow.com/questions/17683458/how-do-i-commit-case-sensitive-only-filename-changes-in-git) for this.

How did you even commit this with failing tests? This is wild.

gajus:
> How did you even commit this with failing tests? This is wild.

`git commit ... --no-verify`

thsmi:
As you can see in the comment in @allthesignal pull request it works locally.

The issue quite obvious is in eslint-plugin-jsdoc/test/rules/index.js implementation.
The following line is causing this:

const assertions = require('./assertions/' + _.camelCase(ruleName));

If your are running the tests on a case insensitive filesystem like windows, macos they will succeed. 
But if you run them on a case sensitive filesystem like Linux normally is, the test will fail. 

Require was originally build for http, which use normally a case insensitive path. This is the reason why normally Node.js modules should use all lowercase and separate with dots or dashes. By doing so they avoid running into this camel case issue.

allthesignals:
@thsmi thanks, got it, I figured this was a filesystem issue. Hmm so what's the fix? I will give it another try, these workarounds for renaming with changes to case don't seem to work...

> git commit ... --no-verify
@gajus thanks - this seems like cheating! 😂

gajus:
Have you tried `git mv`?

allthesignals:
> Have you tried `git mv`?

@gajus Same issue =/ _however_ I did a simple temporary rename and it picked up the change. Passing tests on CI here: https://github.com/gajus/eslint-plugin-jsdoc/pull/205.

PR #193, opened 2019-04-07
Use estree information to check for returns.

thsmi:
The current implementation uses regex to detect return statements.

But this approach is error prone. It fails in case of nested return functions as well as strings which match the regex.

This pull request changes the rule to use the estree information. It walks through the tree and searches for "ReturnStatements".

By doing so it improves the quality dramatically and avoids false positives.
The estree is always provided by eslint internals. As it is already in memory the it is just walking the tree.

### DIFF
--- a/src/rules/requireReturns.js
+++ b/src/rules/requireReturns.js
@@ -1,51 +1,329 @@
-import _ from 'lodash';
 import iterateJsdoc from '../iterateJsdoc';
+
+const DEFAULT = '@default';
+const LOOP = '@loop';
+
+const RETURN_STATEMENT = 'ReturnStatement';
+const IF_STATEMENT = 'IfStatement';
+const EXPRESSION_STATEMENT = 'ExpressionStatement';
+const SWITCH_STATEMENT = 'SwitchStatement';
+const BLOCK_STATEMENT = 'BlockStatement';
+const FUNCTION_EXPRESSION = 'FunctionExpression';
+const ARROW_FUNCTION_EXPRESSION = 'ArrowFunctionExpression';
+const FUNCTION_DECLARATION = 'FunctionDeclaration';
+const TRY_STATEMENT = 'TryStatement';
+
+const LOOP_STATEMENTS = ['WhileStatement', 'DoWhileStatement', 'ForStatement', 'ForInStatement', 'ForOfStatement'];
+
+const STATEMENTS_WITH_CHILDREN = [
+  LOOP,
+  SWITCH_STATEMENT,
+  IF_STATEMENT,
+  BLOCK_STATEMENT,
+  TRY_STATEMENT
+];
+
+const RETURNFREE_STATEMENTS = [
+  'VariableDeclaration',
+  'ThrowStatement',
+  'FunctionDeclaration',
+  'BreakStatement',
+  'ContinueStatement',
+  'LabeledStatement',
+  'DebuggerStatement',
+  'EmptyStatement',
+  'WithStatement',
+  'ThrowStatement',
+  EXPRESSION_STATEMENT
+];
+
+const ENTRY_POINTS = [FUNCTION_DECLARATION, ARROW_FUNCTION_EXPRESSION, FUNCTION_EXPRESSION];
+
+/* eslint-disable sort-keys */
+const lookupTable = {
+  [RETURN_STATEMENT]: {
+    is: (node) => {
+      return node.type === RETURN_STATEMENT;
+    },
+    check: (node) => {
+      if (!lookupTable[RETURN_STATEMENT].is(node)) {
+        return false;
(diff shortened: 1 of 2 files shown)


src/rules/requireReturns.js
```diff
@@ -1,51 +1,329 @@
...
+      // throw new Error('Unknown node type: ' + node.type);
+      return false;
+    }
   }
+};
 
-  const targetTagName = utils.getPreferredTagName('returns');
+/**
+ * Checks if the source code returns a retrun value.
+ * It traverse the parsed source code an retruns as
```
brettz9: sp. nit for `retrun` in both lines above

brettz9:
Could we get a review of this, so I can rebase onto it if it is acceptable and get the remaining eslint-deprecated items merged?

src/rules/requireReturns.js
```diff
@@ -1,51 +1,329 @@
...
-  // inheritdoc implies that all documentation is inherited
-  // see http://usejsdoc.org/tags-inheritdoc.html
-  //
-  // As we do not know the parent method, we cannot perform any checks.
-  if (utils.hasTag('inheritdoc')) {
-    return;
+const DEFAULT = '@default';
+const LOOP = '@loop';
+
+const RETURN_STATEMENT = 'ReturnStatement';
```
gajus: I don't understand the stylistic purpose of using constants to access methods.

I would rather just use the methods as they are. If there is an ESLint rule that is complaining about variable names, then just disable that rule for this file, since we are dealing with 3rd party API.

gajus:
LGTM. There are a lot of grammar mistakes in the comments, but the code makes sense.

src/rules/requireReturns.js
thsmi: Javascript is missing an explicit forward declarations.

When walking a tree you need to have forward declaration but Javascript grammar is non linear. E.g. a function consists of a block. And blocks are allowed to have functions. So you there is no way to order the calls to that the "use before defined" rule would be satisfied. And I think disabling this rule is definitely not a good idea. It is very powerful protection against typos in function names.

The cleanest workaround is to wrap everything into a class. As class members are by definition explicit forward declarations. There is no "used before defined" for class members. But using a classes is outlawed by a linter rule. E.g. the Acorn parser as well as estree are using this approach. 

An other common way is using a lookup table as it is done here. You define all your entry points into an hashtable or an array and then use the index to resolve the binding at runtime. This aligns with the linter rule and is a common practice in e.g. in Mozilla's code base. 

So the options are basically:
  * Keep it as it is and be compliant to the rule set.
  * Wrap everything into a class, and disable this the no classes rule for this file. This would be my preference.
  * Use the function calls as the are, as you suggested. But this implies disabling the "use before defined" rule.

gajus: > So you there is no way to order the calls to that the "use before defined" rule would be satisfied.

I don't see where this is an issue, e.g. `EXPRESSION_STATEMENT` That is just used to inline 'ExpressionStatement' into `RETURNFREE_STATEMENTS` array.

Equally, something like `IF_STATEMENT` could be just replaced with:

```diff
-[IF_STATEMENT]: {
+IfStatement: {}
```suggestion
const RETURN_STATEMENT = 'ReturnStatement';
```

```diff
- return node.type === IF_STATEMENT;
+ return node.type === 'ReturnStatement';

```

etc.

thsmi: In order to avoid confusion, I think two stories are mixed up. 
  * The lookup table is there to bypass the "use before defined" warnings, this is unrelated to the constants 
  * The constants are used to externalize strings and thus avoid runtime errors and improve the code stability and quality.

The rational behind the constants is: use the compilers builtin protection against typos.

 Just take the following example. It was changed as you suggested but I added a tiny typo.

```
const lookupTable = {
  ReturnStatement: {
    is: (node) => {
      return node.type === 'ReturnStatement';
    },
    check: (node) => {
      if (!lookupTable.ReturStatement.is(node)) {

```

This would crash at runtime which because an "n" is missing in the last line.  Which makes the dynamic binding fail. There is no chance in this case to detect this upfront.

This cannot happen with the constants. Because in this case the array lookup uses the "[]" operator instead of the "." operator. They are almost same, with a tiny exception. The [] operator requires that the identifier is an existing variable or constant. This enables the es-lints "no-undef" rule to detect such typos and show error as there would be neither a variable nor a constant with a matching name.

gajus: I would buy the argument for these constants if they were exported by the AST iterator, but when they are defined in the codebase itself, then it is only a matter of where the typo can be made. Other than this, such style is alien to the rest of the codebase and should be discussed separately.

Please inline the constants used for mapping and I will merge the PR.

PR #194, opened 2019-04-08
`forceRequireReturn`

brettz9:
- Add setting `forceRequireReturn` to require checks regardless of whether a non-nested return exists in the function

The advantage is for the sake of flagging to documenters that the `undefined` return value was considered and documented with no need for further documentation rather than being indeterminate and the function's return in need of examination.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2705,6 +2705,28 @@ function quux (foo) {
  */
 function quux (foo) {
 }
+
+/**
+ * @override
+ */
+function quux (foo) {
+
+  return foo;
+}
+
+/**
+ * @class
+ */
+function quux (foo) {
+
+}
+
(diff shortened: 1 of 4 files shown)


gajus:
Did you see this PR?

https://github.com/gajus/eslint-plugin-jsdoc/pull/193

Some features seem to be clashing.

brettz9:
No, I hadn't seen it. Such an approach should in theory indeed be superior. I only moved the old code out into a utility (ESLint was complaining of complexity there too). My own addition is just some config. If it's easier, I can apply my changes on top of #193 if that PR is good by you (haven't studied it myself).

brettz9:
Closing in favor of #217.

PR #205, opened 2019-04-17
Implement a requireJSDoc compatible rule with working tests

allthesignals:
Forked from the proposed pull request here: https://github.com/gajus/eslint-plugin-jsdoc/pull/188.

Original text from @thsmi: 
> Implements a requireJSDoc compatible rule as described in issue #162.
>
> The unit test file is ported from eslint and therefore MIT licensed for more details refer also to issue #162


allthesignals:
Actually, the tests pass locally, so I'm not sure what's happening.

allthesignals:
Should be good to go! The latest commit here is a cleaner approach using `git mv` and `git rm`.

src/rules/requireJsdoc.js
```diff
@@ -0,0 +1,166 @@
+//
```
gajus: Unnecessary comment.

src/rules/requireJsdoc.js
```diff
@@ -0,0 +1,166 @@
...
+// The requireJDDoc rule at eslint is deprecated and won't get any bugfixes anymore.
+// Thus this method implements a compatible requireJSDoc rule.
+//
+
+/* eslint-disable sort-keys */
```
gajus: Unnecessary eslint-disable/ comment.

Sorted property style is enforced across the entire library. This rule implementation is not an exception.

gajus:
Thank you for cleaning this up. I've added to style comments. Can you please address them? Otherwise, I am happy to jump in and amend the the PR in the next 24 hours.

allthesignals:
@gajus I will be happy to - @thsmi, is this okay? I feel I am co-opting your original PR. Otherwise, will continue with cleanup.

allthesignals:
@gajus I committed some changes that address your comments. Let me know if there's anything else. 

Thanks!

gajus:
Noticed after merging that this was submitted without the accompanying documentation. That would be a nice follow up PR.

PR #207, opened 2019-04-18
Prefix the tag name with an @ sign to avoid replacements in non-tag-names

TuckerWhitehouse:
#184 added support for auto-fixing the `check-tag-names`, but it looks like the replacement is on the entire string, and not tag names specifically. This PR is to prefix the replacement with an `@` symbol to try and target tag names.

I'm not sure if it would be better to do a global find and replace so all tag names are updated, not just the first match? e.g. `new RegExp('@' + jsdocTag.tag, 'g')`

I was also looking to add tests, but I didn't see anything for the autofixes - if you have some guidance on how I might go about adding them, I'd be more than happy to.

### DIFF
--- a/src/rules/checkTagNames.js
+++ b/src/rules/checkTagNames.js
@@ -14,7 +14,7 @@ export default iterateJsdoc(({
 
       if (preferredTagName !== jsdocTag.tag) {
         report('Invalid JSDoc tag (preference). Replace "' + jsdocTag.tag + '" JSDoc tag with "' + preferredTagName + '".', (fixer) => {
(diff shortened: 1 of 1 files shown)


TuckerWhitehouse:
Sorry, I think this is actually a bug. Give a comment like this:
```js
/**
 * Some description that talks about the return value...

 * @return {Object}
 */
```
The current behaviour replaces `return` with `returns` in the description and leaves the tag name unmodified. Because the tag name is unchanged, the fix will run on every execution and continue to replace `return` with `returns` which ends up with something like this:
```js
/**
 * Some description that talks about the returnssssssssssssssss value...

 * @return {Object}
 */
```

gajus:
Thank you for the explanation.

Issue #200, opened 2019-04-16
Avoid `require-returns` when `@class` present


brettz9:
While we detect for `class` now, we don't detect for a normal function labeled with `@class` to suppress the need for a `return`.

brettz9:
Fixed by #217.

Issue #201, opened 2019-04-16
Suppress need for `require` in `require-returns` when `@implements` present  


brettz9:
While normally the issue of a function serving as a class may be handled by #200 , the fact that http://usejsdoc.org/tags-implements.html does not exclude the possibility of a normal function using `@implements` to point to a function signature that it implements, I would think this case should be suppressed as well. @sandersn

brettz9:
And if instead, TypeScript disallows this, one may wish to instead *prevent* `@implements` on a function unless that function is marked with `@class`.

sandersn:
+1 for adding the option to prevent `@implements` on functions. This is exactly the kind of jsdoc error we can't put in Typescript itself because we don't want to be noisy, but it's perfect for eslint.

brettz9:
Ok, I agree that is a good option, but until jsdoc may implement `typeof` (which IIUC, is the correct replacement in Typescript; see also https://github.com/jsdoc3/jsdoc/issues/1349 and https://github.com/jsdoctypeparser/jsdoctypeparser/issues/50 ), I think jsdoc users should have the ability to suppress the error.

brettz9:
Also, does Typescript have any way to easily indicate a signature of an options object which has say several properties and callbacks?

`typeof` seems to assume that there is already an implementation in code, no?

`@typedef` might work, e.g., with a type of "function" for callback properties, but this wouldn't allow a full function signature.

And `@interface` seems it might be an approach to use, as, per http://usejsdoc.org/tags-interface.html#virtual-comments , but I'm not sure on whether it is intended to be used for the likes of object literals.

sandersn:
Is this what you mean for "indicate a signature of an options object which has several properties and callbacks"?

```js
/** @typedef {{
  (n: number): number
  p: number
  opt?: number
  m: (n: number) => number
}} Foo */

`Foo` is callable, has a method `m` and two other properties `p` and `opt`. `opt` is optional. You can use `@callback` if you want to provide documentation for `m`'s signature separately. I don't think it's possible for the signature of `Foo` itself.

brettz9:
Ah, ok... Very nice, thanks...

brettz9:
Fixed by #217.

Issue #158, opened 2019-02-04
`require-returns` requires `@return` tags for empty return statements


pelepelin:
Using 4.1.0

For the purpose of `require-returns`, the function body that contains only return statements without expressions should be considered equivalent to a function without return statements.
The following examples should be considered valid:
```js
/**
 *
 */
function foo() {
  return;
}

/**
 *
 */
function foo() {
  if (1) {
    return;
  } else {
    return;
  }
}
```

pelepelin:
This is the reverse of #142

pelepelin:
Oops. A duplicate of #149 which is closed. Does #142 cover both cases?

mattlyons0:
I am using `4.8.2` (which is supposed to have the fix for #142) and still seeing ```function foo() { return }``` say it is missing the return function

brettz9:
Fixed by #217.

PR #217, opened 2019-05-13
Rework require returns

### LINKED ISSUE #151: Inner return calls fail the require-returns rule
The following function has no returned value, but it is marked as failing the `require-returns` rule because of the internal `return` call from the array's `filter()` method:

```
removeItem(itemForRemoval) {
    const newItems = this.Items.filter(
      ({ label, index }) => {
        return label !== itemForRemoval.label;
      }
    );
    this.saveItems(newItems);
 }
```

brettz9:
This build's on @thsmi 's great work with #194, refactoring to use string literals in place of the string constants as requested.

I also switched arrow functions to object shorthand since there was no advantage to using arrow functions here; besides taking up slightly more space, they are less performant. (See https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1 )

Note that this also fixes #151, #158, #200 and #201.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2721,6 +2721,28 @@ function quux (foo) {
  */
 function quux (foo) {
 }
+
+/**
+ * @override
+ */
+function quux (foo) {
+
+  return foo;
+}
+
+/**
+ * @class
+ */
+function quux (foo) {
+
+}
+
+/**
+ * @constructor
+ */
+function quux (foo) {
+
+}
 ````
 
 
--- a/README.md
+++ b/README.md
@@ -2515,7 +2515,7 @@ The following patterns are considered problems:
 function quux (foo) {
 
 }
-// Message: Present JSDoc @returns declaration but not available return expression in function.
+// Message: JSDoc @returns declaration present but return expression not available in function.
 
 /**
  * @return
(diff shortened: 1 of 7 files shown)


brettz9:
I can look to rebase my changes from #194 after this may be approved.

brettz9:
I've also added the examples from https://github.com/gajus/eslint-plugin-jsdoc/issues/142#issuecomment-452566272 as tests and expanded the code a little to handle checking of arrow function bodies (and ensuring context is passed on).

brettz9:
The changes really apply (or ought to apply) to require-return-checks, so I factored out code to be shared in common, and applied to both. Now both rules are getting the advantage of the AST checking, while allowing for choice between the rules. 

I also added my forceRequireReturns check on top of this as it was only a minor addition.

brettz9:
And for both rules, I am treating async functions as returning a value, since they return Promises.

I think this PR is now adequate to replace both #193 and #194

brettz9:
As it was a negligible addition, I also tacked on a fix to #201 (and the original PR also fixes #200).

brettz9:
As further additions--while on the subject--and reviewing `valid-jsdoc`'s behavior:
 
- `require-returns-check`: If a return is documented, allow `@class` and `@interface` (as with `@constructor`) to ignore reporting a missing return in body as constructors implicitly return the class; for testing, add getter check
- `require-returns`: Allow presence of `@interface` to avoid need for `@returns` (as `@interface implies a class`); Allow presence of ES setter to avoid need for `@returns` as setters don't return

The one deviation from ESLint's `valid-jsdoc` behavior is that while with this PR we no longer report setters without returns, we do throw if an ES getter is missing a return, as it is reasonable that these would be documented (but for some reason valid-jsdoc exempts getters too).

PR #220, opened 2019-05-14
Update to later ESLint rule format

### LINKED ISSUE #19: Support ESLint 3 and new rule format
The new rule format, introduced in the ESLint version 3, is explained here http://eslint.org/blog/2016/07/eslint-new-rule-format

Any chance of converting the rules in that format and making a new major release which support ESLint 3 and up?

brettz9:
- Change rule format (require-jsdoc already in this format and all other rules use iterateJsdoc.js) per issue #19

Fixes #19.

### DIFF
--- a/src/iterateJsdoc.js
+++ b/src/iterateJsdoc.js
@@ -178,113 +178,116 @@ export {
   parseComment
 };
 
-export default (iterator) => {
-  return (context) => {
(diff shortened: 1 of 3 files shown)


brettz9:
FYI, this is almost exclusively just whitespace changing as you can confirm.

Issue #107, opened 2018-11-28
Incorporating now deprecated `valid-jsdoc` ESLint jsdoc rules and options which we do not already support


brettz9:
If you weren't aware, ESLint has end-of-lifed their jsdoc support: https://eslint.org/blog/2018/11/jsdoc-end-of-life (and are recommending eslint-plugin-jsdoc as a replacement).

~Besides this it would be very nice to incorporate the functionality of ~`require-jsdoc` and~ `valid-jsdoc` which are not part of eslint-plugin-jsdoc and document how to transition...~  *See https://github.com/gajus/eslint-plugin-jsdoc/wiki/Comparison-with-deprecated-JSdoc-related-ESLint-rules for already existing equivalents or near-equivalents and below for the remaining.*

~For `require-jsdoc`, see #162. ([original source](https://github.com/eslint/eslint/blob/0fb5fd402334098dc44cbfbb8ab25919da04843d/lib/rules/require-jsdoc.js) and [tests](https://github.com/eslint/eslint/blob/caeb223c4f7b0b6fe35e5348ae0df4c6446b5bed/tests/lib/rules/require-jsdoc.js))~ *Merged*

~For our need to deal with the deprecation of `sourceCode.getJSDocComment`, see #189.~ *Merged*

~We have no equivalents yet for the following behaviors:~

- [x] - ~[`require-returns`](https://github.com/gajus/eslint-plugin-jsdoc#require-returns) - "`@constructor` tag allows missing `@returns` tag"~ [v4.8.1](https://github.com/gajus/eslint-plugin-jsdoc/releases/tag/v4.8.1)
- [x] - ~[`require-returns`](https://github.com/gajus/eslint-plugin-jsdoc#require-returns) -  "class constructor allows missing `@returns` tag"~ [v4.8.1](https://github.com/gajus/eslint-plugin-jsdoc/releases/tag/v4.8.1)
- [x] - ~[`require-returns`](https://github.com/gajus/eslint-plugin-jsdoc#require-returns) -  "`@override` tag allows missing ... `@returns` tags"~ [v4.8.1](https://github.com/gajus/eslint-plugin-jsdoc/releases/tag/v4.8.1)

We have no equally powerful/granular replacements for the following options:

[...]

PR #225, opened 2019-05-15
- Add matchDescription for #107

brettz9:
Fixes part of #107.

This is working in my local Node environment, but not with Node 4 which doesn't support Unicode regexes. Can we bump to Node >= 6?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -6,6 +6,7 @@ import checkParamNames from './rules/checkParamNames';
 import checkSyntax from './rules/checkSyntax';
 import checkTagNames from './rules/checkTagNames';
 import checkTypes from './rules/checkTypes';
+import matchDescription from './rules/matchDescription';
 import newlineAfterDescription from './rules/newlineAfterDescription';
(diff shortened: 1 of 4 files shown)


gajus:
Done

brettz9:
My apologies, it seems that while Unicode regexes (with the 'u' flag) are supported as of 6.0, the property escapes I needed here are not yet available until 10.3. Not a big deal from that angle, as I just wanted it in the default regex to allow for default upper-casing independent of language. Hope you don't mind having made the bump though, as it should make for a lot cleaner code at least, such as allowing destructuring and defaults. I will look to refactor.

brettz9:
Ok, I've refactored to just support English upper-case by default (the config allows one to add one's own regex anyways).

I came across another regex error in another rule (`require-hyphen-before-param-description`) due to it also using a feature not yet supported in Node 6; that is also fixed in this PR.

brettz9:
Btw, thanks for the commit access! Would you like me to do merges and approvals for any I feel comfortable accepting, or reserve my access for minor fixes/tweaks?

PR #237, opened 2019-05-17
npm updates

brettz9:
- npm: Update devDeps and peerDep (breaking change?)
- Linting: Disable two new rules; prefer `includes`

I ran `npm-check-updates` and found the devDep items in the PR ready for updates.

It also updated the peerDependency, eslint. I don't know if you'd consider that a breaking change, since it will probably only alert users rather than require them to use eslint 5, but I figure it may be time for a change?

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,6 +2,8 @@
   "extends": "canonical",
   "root": true,
   "rules": {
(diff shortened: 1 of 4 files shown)


gajus:
Not a breaking change. Good to proceed.

PR #243, opened 2019-05-18
Prevent `@implements` on non-constructor functions

### LINKED ISSUE #218: Rule to discourage @implements on regular (non-class) functions
Per discussion in #201 , it may be preferable to discourage @implements on non-class functions (i.e., those without a `@class` or `@constructor` tag or without being an ES6 `class`).

brettz9:
- Add rule `implements-on-classes` to prevent @implements on non-constructor functions; fixes #218

@sandersn

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -82,6 +82,7 @@ Finally, enable all of the rules that you would like to use.
         "jsdoc/check-syntax": 1,
         "jsdoc/check-tag-names": 1,
         "jsdoc/check-types": 1,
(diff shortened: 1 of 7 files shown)


mrfelton:
This rule seems to flag every single method in my codebase whilst I have not used `@implements` anywhere.

brettz9:
Ugh, my apologies... Fixed in 6.0.2...

brettz9:
And it was added to recommended, which is probably the reason it is showing up...

PR #244, opened 2019-05-21
- Add `meta.type`

brettz9:
As per https://eslint.org/docs/developer-guide/working-with-rules#rule-basics .

Not sure how this may be surfaced, but figure it doesn't hurt to document it.

### DIFF
--- a/src/rules/checkAlignment.js
+++ b/src/rules/checkAlignment.js
@@ -23,4 +23,8 @@ export default iterateJsdoc(({
       break;
     }
   }
+}, {
+  meta: {
+    type: 'layout'
+  }
 });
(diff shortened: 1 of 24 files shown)


brettz9:
And I see now that it does already have practical utility; per https://eslint.org/docs/user-guide/command-line-interface , one can call `eslint` with `--fix-type Array` to choose whether to apply fixes to "problem", "suggestion", and/or "layout".

Note too that I didn't label any of our issues as "problem" (even though bad jsdoc's could potentially signal a problem), since I wasn't sure you'd want say invalid types to be, as per https://eslint.org/docs/developer-guide/working-with-rules#rule-basics ,  "identifying code that either will cause an error or may cause a confusing behavior". I can change some of the rules to "problem" though if you think the criteria is met.

Issue #215, opened 2019-05-12
Make `check-types` configurable


golopot:
Users should be able to set the preference of types. The option could looks like this:
```js
{
  '@typescript-eslint/ban-types': [
    'error',
    {
      types: {
        // ban type `Foo` using the default error message
        Foo: null,

        // ban type `Bar` with a custom message.
        Bar: "Don't use Bar!",

        // ban type `String` and tell the plugin how to fix it.
        String: {
          message: 'Use string instead',
          fixWith: 'string',
        },
      },
    },
  ],
};
```

See [@typescript-eslint/ban-types](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md).

As a side note, `ban-types` is a better name than `check-types`. The name `check-types` could mean "check type is valid", "check type is consistent to how it is used" or "check the style of types".

brettz9:
See my [comment](https://github.com/gajus/eslint-plugin-jsdoc/issues/166#issuecomment-480778864) on a related issue, i.e., per issue #107 I think we might look at something like this if reimplementing valid-jsdoc's `preferType`.

brettz9:
Closed by #247. Thanks for the great idea! I tweaked it a little in the implementation, but it is similar as you can see in the docs.

Issue #166, opened 2019-03-13
How to disallow '*' type?


synthecypher:
How can I disallow the `*` type?
```javascript
/**
 * Do something
 * @param {*} a
 * @param {*} b
 * @returns {undefined}
 */
function doSomething(a, b) {
  return undefined;
}
```

brettz9:
This looks like a dupe of #134.

brettz9:
If `preferType` from ESLint's `valid-jsdoc` (#107) is carried over, I'd suggest piggy-backing on this feature so that when keys (i.e., the types) of this object are mapped to values which are `false` (instead of a string indicating a preferred name), then this could indicate a disallowed type.

brettz9:
Closed by #247. See https://github.com/gajus/eslint-plugin-jsdoc/tree/master/.README#settings-to-configure-check-types-and-no-undefined-types for use with `check-types`.

i.e., you'll be able to just add:

```js
{
  settings: {
    jsdoc: {
      preferredTypes: {
        '*': false
     }
  }
}
```

Issue #164, opened 2019-03-04
`check‑types`: Allow lower‑case `object`


ExE-Boss:
TypeScript recommends `object` over `Object`, which&nbsp;is enforced by&nbsp;my&nbsp;ESLint config for&nbsp;the&nbsp;deprecated `valid‑jsdoc` rule, and would require changes to&nbsp;a&nbsp;lot&nbsp;of&nbsp;my&nbsp;code.

brettz9:
If the `preferType` option from `valid-jsdoc` is carried over from ESLint (#107), one could define this rule for oneself (unless you *also* wanted to change the current behavior of `check-types`).

brettz9:
Closed by #247. See https://github.com/gajus/eslint-plugin-jsdoc/tree/master/.README#settings-to-configure-check-types-and-no-undefined-types for use with `check-types`.

You can get this behavior now (in `master`) by:

```js
{
  settings: {
    jsdoc: {
      preferredTypes: {
        'Object': 'object'
     }
  }
}
```

However, I've asked @gajus if we might change the default, as I agree it makes more sense given that objects made with `Object.create(null)` are not `instanceof Object` and as mentioned, Typescript recommends "object".

gajus:
> However, I've asked @gajus if we might change the default, as I agree it makes more sense given that objects made with Object.create(null) are not instanceof Object and as mentioned, Typescript recommends "object".

Go for it.

When merging, the commit message must include "BREAKING CHANGE:" and a description of the breaking change. This will cause new major version release. See https://github.com/semantic-release/semantic-release#commit-message-format

dietergeerts:
So because TS is using `object`, the default is now `object`? What happened to the reasoning behind having it `Object` in the first place? I liked the reasoning because it made sense towards the JS language....

brettz9:
Yes, in part because of TS which has become very popular. But there was a case missed by the original logic. Objects made with `Object.create(null)` are not `instanceof Object`. So if accepting an arbitrary object, it indeed may not be an instance of `Object`.

dietergeerts:
Who would ever do that? I mean yes, technically it's possible, but why?

brettz9:
Using `Object.create(null)` allows you to create objects which aren't at risk for inheriting from `Object.prototype` and for which `for...in` guards are not needed.

So if some code library adds this to a project you or your users are using:

```js
/**
* @example
* ({a: 1, b: 2}).getKeys()
*/
[...]

PR #247, opened 2019-05-21
`preferredTypes`

### LINKED ISSUE #107: Incorporating now deprecated `valid-jsdoc` ESLint jsdoc rules and options which we do not already support

brettz9:
- Change: `check-types` will now include the preferred type in its report message
- Change: `no-undefined-types` now automatically allows `NaN` and `Infinity` types
- Enhancement: `check-types` and `no-undefined-types`: utilize `settings.jsdoc.preferredTypes` map
- Enhancement: `check-types`: Add option object with `noDefaults`
- Enhancement: `no-undefined-types`: Add option object with `preferredTypesAreDefined` and `definedTypes`
- Docs: `no-undefined-types`: Mention role of globals

Closes #107, #164, #166, #215 .

Also unblocks the one item I think is remaining for #1 , namely to correlate our new `preferredTypes` to eslint's deprecated `valid-jsdoc`.

Regarding #164, @gajus, I wonder if you want to also make a breaking change to the default behavior and require `object` instead of `Object`, given that valid-jsdoc did so and TypeScript [recommends so](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html). I think the rationale for this is that `Object.create(null)` will create an object whose `typeof` is `object` but which is *not* `instanceof Object`. While this PR allows overriding the default behavior, I think this makes sense as a new default as well, especially considering the `Object.create(null)` aspect.

There is also one issue in this PR I have not been able to figure out, and wonder if it is an issue with eslint itself. Though I changed this to get the test to passed, what I would expect in line 477 at https://github.com/brettz9/eslint-plugin-jsdoc/blob/a571c68bb988732b45308f3f6ae0353ce71678b3/test/rules/assertions/checkTypes.js#L477  really would instead be "number" lower-cased, as I'd think that both reported errors for this test would be fixed, but only one is. This seems not to be tied to this specific problem but it appears to me only one fix can get made. Any idea why that is?

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -198,6 +198,31 @@ The format of the configuration is as follows:
 }
 ```
 
+### Settings to Configure `check-types` and `no-undefined-types`
+
+- `settings.jsdoc.preferredTypes` An option map to indicate preferred
+  or forbidden types (if default types are indicated here, these will
+  have precedence over the default recommendations for `check-types`).
+  The keys of this map are the types to be replaced (or forbidden) and
+  can include the "ANY" type, `*`.
+  The values can be:
+  - `false` to forbid the type
+  - a string to indicate the type that should be preferred in its place
+    (and which `fix` mode can replace)
+  - an object with the key `message` to provide a specific error message
+    when encountering the discouraged type and, if a type is to be preferred
+    in its place, the key `replacement` to indicate the type that should be
+    used in its place (and which `fix` mode can replace) or `false` if
+    forbidding the type. The message string will have the following
+    substrings with special meaning replaced with their corresponding
+    value (`{{tagName}}`, `{{tagValue}}`, `{{badType}}`, and
+    `{{preferredType}}`, noting that the latter is of no use when one is
+    merely forbidding a type).
+
+If `no-undefined-types` has the option key `preferredTypesDefined` set to
+`true`, the preferred types indicated in the `settings.jsdoc.preferredTypes`
+map will be assumed to be defined.
+
 ### Settings to Configure `valid-types`
 
 * `settings.jsdoc.allowEmptyNamepaths` - Set to `false` to disallow
--- a/.README/rules/check-types.md
+++ b/.README/rules/check-types.md
@@ -2,7 +2,7 @@
 
 Reports invalid types.
 
-Ensures that case of native types is the same as in this list:
+By default, ensures that case of native types is the same as in this list:
 
 ```
 undefined
@@ -15,6 +15,17 @@ Array
 Date
 RegExp
 ```
+
+#### Options
+
+`check-types` allows one option:
+
+- An option object with the key `noDefaults` to insist that only the
+  supplied option type map is to be used, and that the default preferences
+  (such as "string" over "String") will not be enforced.
+
+See also the documentation on `settings.jsdoc.preferredTypes` which impacts
+the behavior of `check-types`.
 
 #### Why not capital case everything?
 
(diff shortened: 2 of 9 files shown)


brettz9:
Thanks for the reviews! Just wondering though about changing that default--are you inclined to keep it as it is or change it for `object`?

Issue #25, opened 2016-10-31
check-types should accept capitalized primitive types


ZacBrownBand:
Eslint and this plugin conflict.
Eslint suggest that `{string}` should be `{String}`
This plugin suggests that `{String}` should be `{string}`

Jscs setting would be `"checkTypes": "capitalizedNativeCase",`

gajus:
This has been discussed a number of times, e.g. https://github.com/gajus/eslint-plugin-jsdoc/issues/22

ESLint is wrong to recommend `String`. 99% of the time you mean `string`, e.g.

```js
// Thats 'string'
const foo = 'FOO';

// Thats 'String'
const bar = new String('bar');
```

ZacBrownBand:
Right, so `{String}` is a valid type, regardless of if its not the common case.
Could there be an option to allow for these types? Like a `nativeTypes` vs [`strictNativeTypes`](https://github.com/gajus/eslint-plugin-jsdoc/blob/master/src/rules/checkTypes.js#L32)

gajus:
It is such a rare edge case to use these types that I would like to avoid adding support for it. I have not seen a single valid use of a native type constructor in the past 5 years.

Redsandro:
Ah I get it. `{String}` and `{string}` are technically both valid, but they are not the same.

```js
new String('lard') === 'lard' // false

new String('lard') // String {0: "l", 1: "a", 2: "r", 3: "d", length: 4}
'lard' // "lard"
```

​So it's not about consistency, rather about probability of use.

type name | `typeof` | check-types | testcase
--|--|--|--
**Object** | object | **Object** | `({}) instanceof Object` -> `true`
**Array** | object | **Array** | `([]) instanceof Array` -> `true`
**Date** | object | **Date** | `(new Date()) instanceof Date` -> `true`
**RegExp** | object | **RegExp** | `(new RegExp(/.+/)) instanceof RegExp` -> `true`
Boolean | **boolean** | **boolean** | `(true) instanceof Boolean` -> **`false`**
Number | **number** | **number** | `(41) instanceof Number` -> **`false`**
String | **string** | **string** | `("test") instanceof String` -> **`false`**

Confusing. But makes sense.

gajus:
@Redsandro Please contribute a variation of this to the documentation. This is the most often asked question.

Redsandro:
@gajus https://github.com/gajus/eslint-plugin-jsdoc/pull/30

PR #248, opened 2019-05-21
`check-types` now expects "object" instead of "Object"

### LINKED ISSUE #164: `check‑types`: Allow lower‑case `object`

brettz9:
BREAKING CHANGE: `check-types` now expects "object" instead of "Object"

Makes consistent with Typescript recommendation and fact that `Object.create(null)` is not `instanceof Object` but its `typeof` is `object`.

Can restore old behavior through `settings: {jsdoc: {preferredTypes: {object: 'Object'}}}}` (and `check-types` still enabled). Closes #164

### DIFF
--- a/.README/rules/check-types.md
+++ b/.README/rules/check-types.md
@@ -10,7 +10,7 @@ null
 boolean
 number
 string
-Object
+object
 Array
 Date
 RegExp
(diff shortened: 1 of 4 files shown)


gajus:
Just in case, the original reasoning for Object is "documented" here https://github.com/gajus/eslint-plugin-jsdoc/issues/25

PR #257, opened 2019-05-24
enhancement(`no-undefined-types`): allow `function`

brettz9:


brettz9:
I'm thinking this rule is independent of `check-types`, so thinking we should rather be more liberal in accepting more common types, e.g., "array" perhaps ought to be allowed lower-cased too... What about "Boolean", "String", "Number", "Any", "regexp", "date", "nan", "infinity"?

My preference over how check-types works is to use the `typeof` value (i.e., lower-cased) whereever sufficiently distinctive (making exceptions therefore for `null`, `Date`, etc.), so in this scheme, the only ones to add would be `function`.

However, I would think "Boolean", "String", and "Number" should probably be added regardless as they may sometimes be used even in their object form.

PR #259, opened 2019-05-24
Change `preferredTypes` to require "<>" be appended to types by default in order to apply to parent types and otherwise apply them to child types only

### LINKED ISSUE #87: New rule to ensure that arrays, objects, and Promises specify the type they contain
Please add a new rule, which would disallow `@param {Array}`, and instead require `@param {string[]}`, so that the type of items in the array is explicitly specified. The default would therefore be `@param {any[]}`.

And for objects, rather than `@param {Object}`, require `@param {Object<string, number>}` instead.

Thanks!

brettz9:
This PR fixes #87 - allowing one to forbid or rename types which are too generic, e.g., `Array`, `Promise`, or `Object` and have no child types specified.

----
BREAKING CHANGE: Cause `preferredTypes` setting not to apply by default to parent types; need to target type name with "<>" appended to target parent types (and only parent types); can alternatively; also adds option `unifyParentAndChildTypeChecks` to restore the old behavior

Allows for warning against overly generic types only (e.g., `Array` without its own children) or against the parent type only (to disallow `Array.<someType>`)

To restore old behavior, also target type names with "<>" or set the `unifyParentAndChildTypeChecks` option.

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -243,6 +243,11 @@ If `no-undefined-types` has the option key `preferredTypesDefined` set to
 `true`, the preferred types indicated in the `settings.jsdoc.preferredTypes`
 map will be assumed to be defined.
 
+See the option of `check-types`, `unifyParentAndChildTypeChecks`, for
+how the keys of `preferredTypes` may have `<>` appended and its bearing
+on whether types are checked as parents/children only (e.g., to match `Array`
+if the type is `Array` vs. `Array.<string>`).
+
 ### Settings to Configure `valid-types`
 
 * `settings.jsdoc.allowEmptyNamepaths` - Set to `false` to disallow
--- a/.README/rules/check-types.md
+++ b/.README/rules/check-types.md
@@ -21,12 +21,20 @@ RegExp
 
 `check-types` allows one option:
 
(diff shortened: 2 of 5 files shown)


garyking:
This change doesn't seem to work. `{Array}` and `{Object}` aren't reported as errors, with the latest version, whether I set `unifyParentAndChildTypeChecks` to `true`, `false`, or unset it.

brettz9:
I added some more test cases / examples to `master`. You need `preferredTypes` set to either disable `Array` and `Object` (by setting them to `false`):

```js
      settings: {
        jsdoc: {
          preferredTypes: {
            object: false
          }
        }
      }
```

and:

```js
      // Errs with 'Invalid JSDoc @param "foo" type "object".'
      /**
       * @param {object} foo
       */
      function quux (foo) {

      }
```

While this works with `unifyParentAndChildTypeChecks`, you probably don't want that setting as it will also report for `Array.<string>`, `string[]`, etc. The purpose of that (non-default) setting is to allow you to indicate you don't want that type or want it remapped no matter whether it has children or not.

FWIW, there is currently no way to enforce a particular format, e.g., `string[]` over `Array.<string>` or vice versa.

Issue #216, opened 2019-05-12
Move settings to options


golopot:
It is hard to find out which setting adheres to which rule. It is better to move settings specific to one rule to the option of that rule.

brettz9:
Could you change the issue title to not be specific to `require-param`? I think it would be better in the case of most rules to allow for options instead of settings. (There may be a few though which interact across rules.)

golopot:
It seems that only `require-param` and `check-examples` have accompanied settings.

brettz9:
There are others; e.g., `valid-types` invokes `utils.isNamepathType` which, passes on the `checkSeesForNamepaths` setting, etc. It is a little harder to follow in the source because the settings are in iterateJsdoc and the rule pages typically interact with these settings through `utils`.

brettz9:
Also `check-tag-names` has the setting `additionalTagNames` which only it uses, and `valid-types` has two settings (`allowEmptyNamepaths` and `checkSeesForNamepaths`) that only it uses.

The only setting that really is used across rules is `tagNamePreference` and I actually think we might change it so that it only works with `check-tag-names`, as the various require and param rules which use it are, because of it, failing to check for violations in non-preferred (or non-default) tags (e.g., if I prefer `@param`, but have a problem in `@arg`, it won't be caught). While this might be seen as a minor optimization if one knows one has no need to check for other synonyms, I think this is not worth it. But the option does make sense in `check-tag-names`.

golopot:
What is the process for moving settings to options? I suggest we do it one-step in a major release: add options, stop reading settings, and warns to stderr in case a removed setting is set.

brettz9:
Sounds like a fine approach to me... I have my hands pretty full now, but a PR would be welcome...

Bear in mind though re: breaking changes... As much as one doesn't wish to make breaking changes in any project, in a linting project, it can be as little total pain for a project to keep up with more frequent breaking changes and make a few incremental changes as it is to "rip off the bandage" and do it all at once, and with linting, breaking changes are far more likely (especially if one were to interpret every minor loosening or tightening as a breaking change).

With linting, particularly ESLint, one has good feedback about what is failing and why, so it doesn't generally take a lot of debugging and deep refactoring work that might be better done all at once as can be the case for non-linting API changes.

But doing it all at once makes sense, especially as the changes are related, and it is better to increase consistency with the API.

golopot:
[...]

PR #262, opened 2019-05-24
Add exported functions option to require-jsdoc rule

### LINKED ISSUE #192: Require JSDoc on exported methods
Support rule that requires JSDoc comment on publicly exported functions. This was originally proposed for now deprecated ESLint rule https://eslint.org/docs/rules/require-jsdoc.

Original enhancement suggestion:
https://github.com/eslint/eslint/issues/7812

The rule could support ES6 exports and possibly other interfaces such as CommonJS exports.

Extersky:
This PR adds support for configuring require-jsdoc rule to require JSDoc comments for exported function bodies only. There are options for configuring the export checks for CommonJS and ES6 exports.
Fixes #192

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -122,6 +122,7 @@ Finally, enable all of the rules that you would like to use.
 
 - `settings.jsdoc.publicFunctionsOnly` - Missing jsdoc blocks
   are only reported for function bodies that are exported from the module.
+  May be a boolean or object.
 
   This setting object supports the following keys:
 
--- a/.README/README.md
+++ b/.README/README.md
@@ -120,8 +120,8 @@ Finally, enable all of the rules that you would like to use.
 
 ### Requiring JSDoc comments for exported functions in <code>require-jsdoc</code>
 
-- `settings.jsdoc.publicFunctionsOnly` - Missing jsdoc blocks
-  are only reported for function bodies that are exported from the module.
+- `settings.jsdoc.publicOnly` - Missing jsdoc blocks
+  are only reported for function bodies / class declarations that are exported from the module.
   May be a boolean or object.
 
   This setting object supports the following keys:
--- a/.README/README.md
+++ b/.README/README.md
@@ -126,6 +126,7 @@ Finally, enable all of the rules that you would like to use.
 
   This setting object supports the following keys:
 
+  - `ancestorsOnly` - Only check node ancestors to check if node is exported
   - `exports` - ESM exports are checked for JSDoc comments
   - `modules` - CommonJS exports are checked for JSDoc comments
   - `browserEnv` - Populates window variable
--- a/.README/README.md
+++ b/.README/README.md
@@ -126,9 +126,9 @@ Finally, enable all of the rules that you would like to use.
 
   This setting object supports the following keys:
 
-  - `exports`
-  - `modules`
-  - `browserEnv`
+  - `exports` - ESM exports are checked for JSDoc comments
+  - `modules` - CommonJS exports are checked for JSDoc comments
+  - `browserEnv` - Populates window variable
 
 ### Alias Preference
 
--- a/.README/README.md
+++ b/.README/README.md
@@ -118,19 +118,6 @@ Finally, enable all of the rules that you would like to use.
   above functions/methods with no parameters or return values (intended where
   variable names are sufficient for themselves as documentation).
 
-### Requiring JSDoc comments for exported functions in <code>require-jsdoc</code>
-
-- `settings.jsdoc.publicOnly` - Missing jsdoc blocks
-  are only reported for function bodies / class declarations that are exported from the module.
-  May be a boolean or object.
-
-  This setting object supports the following keys:
-
-  - `ancestorsOnly` - Only check node ancestors to check if node is exported
-  - `exports` - ESM exports are checked for JSDoc comments
-  - `modules` - CommonJS exports are checked for JSDoc comments
-  - `browserEnv` - Populates window variable
-
 ### Alias Preference
 
 Use `settings.jsdoc.tagNamePreference` to configure a preferred alias name for a JSDoc tag. The format of the configuration is: `<primary tag name>: <preferred alias name>`, e.g.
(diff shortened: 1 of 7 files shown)


Extersky:
There are at least following limitations:
1) Scopes and branches are not supported. This may not be an issue, but may require significant refactoring if desired to be added later
2) The performance would likely improve significantly if a code file wasn't parsed again for each processed node

brettz9:
I'm not sure what "branches" are in this context (unless it refers to blocks within a scope?), but `export` isn't allowed anywhere but in the main module scope, so IIUC, I think that is not an issue.

From my eyeballing, it looks like this will be quite useful for many projects...

brettz9:
I've amended my comment, as I was forgetting this is for Node modules, not ESM. But speaking of which, how about amending to have an optional `publicFunctionsOnly` check for ESM exports?

Extersky:
With branches I meant to refer conditional branching, such as statically analyzable if blocks, which are possible in CommonJS. One such rare case would be a polyfill module where a shim would only be returned if the class/method is not already available. I would consider it ok that for such classes or methods the comments are not enforced, since it can still offer implementation with no JSDoc comment.

Extersky:
Noticed some more cases where referencing unknown variables would cause errors to be thrown in exportParser.js. The "browserEnv" option also modifies the window variable when the option is not enabled. I'll submit fixes for these today.

brettz9:
As nice as it is to be able to analyze the whole file, if it wouldn't be too hard to add an option to only check ancestors, I think that would be a nice optimization and possibly the only needed use case for some usage scenarios.

LukeSkyw:
Very useful PR.

Sorry, I don't know how eslint and its plugin works internally, but I see you are using a custom parser, will that be a problem for Typescript? Would your rule include dangling functions, interfaces, enums, etc

Besides, would your solution worked has it used to on tslint [completed-docs rule](https://palantir.github.io/tslint/rules/completed-docs/) ? I always got frustrated with this rule because it does not allow precise control on where you do expect documentation and where you don't. For example, how to require documentation on everything on interfaces but to restrict the check to the top level doc for classes (no doc required on methods).

Extersky:
It is definitely useful to consider how the same rule could be made available for both JavaScript and TypeScript. The typescript-eslint project aims at being able to use many of the same ESLint rules with TypeScript. One obvious option is to implement the same rule in that project separately. Another option is to modify this implementation to make extendable, but I'm not sure if that is reasonable approach.
Does TypeScript have a language service planned or already existing that could provide info about source code exports? In that case the custom parser/resolver implemented here could be replaced.

brettz9:
Are we ok to merge this as it is now, @Extersky ?

brettz9:
It'd be nice to get this merged, as I think it should be ready. Anything else @Extersky before sending it on to gajus?

Extersky:
Sorry, I missed this. From my side merging this is ok in case the optimisation option for only checking through ancestors isn't needed yet.

brettz9:
Since this is a preexisting rule, a guard on possible performance problems would be of greater assurance. Do you have an estimate on how long the option might take to implement?

Extersky:
I have now added an option for only checking the ancestor nodes when the nodes are checked if they are exported or not. This means that only one top level statement is checked at most when a node is checked for exports. From my side, this can now be reviewed again.

brettz9:
I noticed that `publicOnly` was implemented as a setting, and as discussed in #216 , we've been moving toward options when only one rule is impacted, so I switched that. An additional benefit that using an option gets us is that eslint has a `schema` mechanism for it, so it can do early validation of options structure.

I also added some additional documentation (a summary of options, defaulting behavior, and that the keys are booleans), including for the `require` option which had been missing docs.

brettz9:
If you want to take a further look, I think we should then be ready for merging.

Extersky:
The added documentation mentions browser window global export as an option, but this is not yet supported by browserEnv option. Should this be added? This is a simple modification to also check the window variable. I'll start preparing this in case it is included.

Extersky:
browserEnv option now causes exports to be checked from window global variable. This doesn't yet populate window variable from GeneratorDeclarations, but this isn't currently supported in the rule at all.

brettz9:
Yeah, sorry, I meant "checks" in the docs, but your docs for it are even better.

This looks great now. I think it is ready to merge. Though it may be helpful if you could offer a summary of any remaining to-dos for the lesser features that are not yet implemented, e.g., checking generator declarations.

@gajus , as a new feature with changes since your approval, do you still want notice? Don't mean to be annoying, just presume it is better to err on the side of caution until I have a better idea of what to pass on, as I took your earlier comment as being about smaller fixes and this is part of a new feature.

gajus:
> @gajus , as a new feature with changes since your approval, do you still want notice? Don't mean to be annoying, just presume it is better to err on the side of caution until I have a better idea of what to pass on, as I took your earlier comment as being about smaller fixes and this is part of a new feature.

I have no issues with you taking the lead. All changes you've pushed that I have reviewed did not ask for a single comment. Just take the lead and ping me if you need any help. I only ask not to do major code structure changes without first raising an issue (such as rewrite in TS, changing ESLint preset, etc).

brettz9:
Great, sounds good, thanks!

brettz9:
@Extersky : I've made one other change (discussed off list), re: switching to "esm", "cjs", and "window" for greater clarity on how the export types apply, as "exports" can, as you suggest, be confusable with `exports` in "cjs". I also thought `browserEnv` was misleading as well, since we were not checking for the environment; nor do I think we should given that some polyfills may work in a Node environment yet still expect docs on `window` if they set that option to `true`.

brettz9:
@Extersky : I just went through the codebase to try to get us to 100% coverage, and though I got through everything else, including some changes to `hasReturnValue` which would impact `require-jsdoc`, I was wondering if you might be able to help add tests for those lines in `exportParser.js` not yet covered in our tests? 

Toward this end, I added an npm script, `npm run test-cov` which gives a printout showing a summary of which lines are still uncovered across all files. (The uncovered lines column (for `exportParser.js`) is truncated at the beginning, but one can start from the line numbers listed at the end.) If there really is some code which will not be covered now (and which has no public interface), we can disable (with `/* istanbul ignore next */`), as I did for a handful of lines in `jsdocUtils.js` since they are private and are not currently possibly to be visited, but which are nevertheless probably wise to keep there as guards, e.g., in case expanded in the future for other types.

PR #264, opened 2019-05-29
enhancement(check-alignment): add fixer

jasminexie:
* add fixer for rule `check-alignment`.


gajus:
PR is not automatically released because the commit does not follow semantic-release conventions.

brettz9:
Thanks, forgot it was "feat"...

brettz9:
I've reversed the commit and reapplied with the semantic-release convention...

PR #266, opened 2019-05-30
Fix fixer for require-hyphen-before-param-description

jasminexie:
Thank you for the quick response to #265, but it appears I haven't tested thoroughly. Sincerest apologies!

This PR fixes hyphen insertion for the following cases:

* Parameter name and description are equal
```
/*
 *@param foo foo
 */
function () {}
```
by replacing the *last* occurrence in a line, not the *first*.
* Multiline descriptions
```
/*
 * @param foo single line
 * another line
 */
 function () {}
```
by replacing the (last occurrence of (the first line of the description)) in a line.

That's a mouthful :/
Hopefully this will have covered all cases. Please let me know if I've missed anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3033,8 +3033,19 @@ function quux () {
 // Message: There must be no hyphen before @param description.
 
 /**
- * @param foo - Foo.
- * @param bar Foo.
+ * @param foo - foo
+ * @param foo foo
+ */
+function quux () {
+
+}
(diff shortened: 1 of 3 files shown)


brettz9:
Thanks for the fixes!

Btw, if you're ever in Shenzhen, we've had a kind of informal meeting for open source... brettz9 is my WeChat ID...

PR #269, opened 2019-05-30
fix: Stop warning missing param and returns tags if `@type` is present

### LINKED ISSUE #268: require-return, require-param: don't warn if @type is present
If function is annotated by `@type` tag, then it does not need `@param` and `@returns`. Annotating a function by `@types` tag is handy when you need to write a long function type many times, for example, an express request handler. 

**code**: 
```js
/**
 * @typedef {(req: Request, res: Response, next: NextFunction) => void} RequestHandler
 */

/** @type {RequestHandler} */
function foo(req, res, next) {

}
```
**current behavior**:
Warns about missing params tag and missing returns tag 
**expected behavior**:
No warnings.

golopot:
fixes #268

### DIFF
--- a/README.md
+++ b/README.md
@@ -4252,6 +4252,9 @@ function quux (foo) {
 // issue 182: optional chaining
 /** @const {boolean} test */
 const test = something?.find(_ => _)
+
+/** @type {RequestHandler} */
+function foo(req, res, next) {}
 ````
 
 
@@ -4856,6 +4859,11 @@ function quux () {
   return;
 }
 // Settings: {"jsdoc":{"forceRequireReturn":true}}
+
+/** @type {RequestHandler} */
+function quux (req, res , next) {
+  return;
+}
 
 /** foo class */
 class foo {
--- a/src/rules/requireParam.js
+++ b/src/rules/requireParam.js
@@ -12,6 +12,11 @@ export default iterateJsdoc(({
     return;
   }
 
+  // Param type is specified by type in @type
+  if (utils.hasTag('type')) {
+    return;
+  }
+
   functionParameterNames.some((functionParameterName, index) => {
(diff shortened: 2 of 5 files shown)


brettz9:
I'm thinking maybe the code for this should be moved to `utils.avoidDocs`, which also add the behavior for `require-description` and `require-example` which I think should be exemptable by similar logic. But as with other items in `utils.avoidDocs`, it may be best to have (effectively on-by-default) config (like `implementsReplacesDocs`) to disable the behavior. Take a look inside (the end of) `utils.avoidDocs`. And as with those items, perhaps `@type` should have a check with `utils.classHasTag('type')` in case it is used on a class, thus working as a kind of `@implements`.

golopot:
These things can happen in another pr.

brettz9:
While it would lead to more features, this is also a suggestion on how to refactor so that it avoids adding code in more places than we actually want.

golopot:
I don't see value in putting it into utils since the logic is one-line.

brettz9:
The value is that there is already a method for this in utils--no need to add a new method. Your logic just needs to be added inside that method (i.e., `avoidDocs`). This will also give the benefit of your improvements working with `require-description` and `require-example`.

golopot:
But the presence of types tag should not disable require descriptions and require example. Because types tag block does not includes information of examples and descriptions. While types implies params and return types.

brettz9:
I guess it may be different enough, especially for examples as a different method name would be needed. I'd think a description would be similar, even if the API is different, but I guess I see your point.

brettz9:
@gajus: I think this is good to go as is, if you're ok with it...

gajus:
> @gajus: I think this is good to go as is, if you're ok with it...

You are doing an excellent job. No need to add me to review every change.

PR #270, opened 2019-06-01
`check-types`: Distinguish `.<>` from `<>`; support `[]`; check/fix nested array/object types

brettz9:
BREAKING CHANGE: change `<>` to no longer apply to `SomeType.<>` but instead to now support `SomeType<>`
feat: pseudo-type `[]` to catch parent array types of form `string[]`, `number[]`, etc.
feat: support `.<>` as separate type
fix: ensure working with nested type arrays/objects
fix: ensure fixer preserves angle-bracket-dot or square bracket notation for arrays/objects
chore: Update husky, semantic-release devDeps; add `typescript` devDep for `type-fest` peer dep. (used indirectly by husky and semantic-release)

Breaking change through `jsdoctypeparser` update.

----

Also includes a minor adjustment to `no-undefined-types` to be able to recognize types suffixed with the pseudo-types (e.g., `<>`).

~This is currently being blocked by https://github.com/jsdoctypeparser/jsdoctypeparser/pull/60 but once that may be merged and released, this PR should only need to target the new version as its dependency. (And tests failing here should be due to this need.)~ *Now incorporated*

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -226,12 +226,37 @@ but restricted to `@param`. These settings are now deprecated.
 - `settings.jsdoc.preferredTypes` An option map to indicate preferred
   or forbidden types (if default types are indicated here, these will
   have precedence over the default recommendations for `check-types`).
-  The keys of this map are the types to be replaced (or forbidden) and
-  can include the "ANY" type, `*`.
+  The keys of this map are the types to be replaced (or forbidden).
+  These keys may include:
+  1. The "ANY" type, `*`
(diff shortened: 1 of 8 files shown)


brettz9:
Not sure if semantic release can handle multiple angular-type statements in a single commit... I didn't see docs about this...

gajus:
> Not sure if semantic release can handle multiple angular-type statements in a single commit... I didn't see docs about this...

It will just use whichever has the greatest power. In this case, the BREAKING CHANGE will just force a new major version (if I understand the question correctly).

Issue #276, opened 2019-06-11
require-returns not working for async functions that don't return anything


AndrewLeedham:
Seems like https://github.com/gajus/eslint-plugin-jsdoc/issues/174 has resurfaced in later versions. I created a minimal repro that has the same issue when running with both TS and JS, just run `yarn lint:ts` and `yarn lint:js` to see the issue: https://github.com/AndrewLeedham/eslint-plugin-jsdoc-repro

**ESLint version**: 5.16.0
**eslint-plugin-jsdoc version**: 8.0.1
**Node version**: v10.15.3

gajus:
CC @golopot

golopot:
#174 address the rule `require-returns-check` while this issue addresses `require-returns`, so this is a new issue, not a regression.

I agree with you this is a false positive, however there are test cases saying your code is considered warning. So this issue might needs decision from the maintainers. @gajus @brettz9

https://github.com/gajus/eslint-plugin-jsdoc/blob/a9490947bb85960fbe41dae968a5875ff2085c0d/test/rules/assertions/requireReturns.js#L88-L104

I consider this a false positive:
```js
// Error: Missing JSDoc @returns declaration.
/**
 *
 */
async function quux() {}
```

brettz9:
I personally like to use the force return setting, and even use the new settings to reject a plain child `Promise` (requiring it be of the parent form, i.e., `Promise<...>`, instead).

But I think others closer to my preferences may, even if they don't want reminders for plain `undefined` returns, still wish to be reminded that a Promise is being returned there, even if it is admittedly redundant with the `async`. I think some config to allow either would be desirable, though I'd think we could loosen the default to avoid insisting on a return for a mere async.

AndrewLeedham:
So default would not flag an empty async function, but setting this new option to true or setting `forceRequireReturn` to true would flag it?

What would this new option be called `forcePromiseReturn` perhaps?

brettz9:
Right, that was my thinking... But maybe it should be`forceAsyncReturn`, or more clearly, `forceReturnsWithAsync` to avoid giving the impression it has anything to do with normal `return Promise(...)`.

AndrewLeedham:
Would it make sense to add the same setting to `require-returns-check`. As the test assertion
```
          /**
           * @returns {Promise<void>}
           */
          const quux = async () => {}
```
would fail, when defaulting an async function to not return an empty Promise.

PR #277, opened 2019-06-12
feat(require-returns): add forceReturnsWithAsync

AndrewLeedham:
Async functions implicitly returning `Promise<void>` is now behind the new `forceReturnsWithAsync` (or `forceRequireReturn`) setting.

fixes: #276

### DIFF
--- a/test/rules/assertions/requireReturns.js
+++ b/test/rules/assertions/requireReturns.js
@@ -157,6 +157,28 @@ export default {
           /**
            *
            */
+           async function quux () {}
+      `,
+      errors: [
+        {
+          line: 2,
+          message: 'Missing JSDoc @returns declaration.'
+        }
+      ],
+      parserOptions: {
+        ecmaVersion: 8
+      },
+      settings: {
+        jsdoc: {
+          forceRequireReturn: true
+        }
+      }
+    },
+    {
+      code: `
+          /**
+           *
+           */
           function quux () {
           }
       `,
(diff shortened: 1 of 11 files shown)


AndrewLeedham:
@brettz9 Thoughts?

brettz9:
It looked good at my bird's eye view. When I may have a chance, I would like to have a closer look at the tests and the tests that were changed. Just those nits for the moment...

brettz9:
I went ahead and fixed the nits and added back the tests that were removed. While some functionality may cover similar ground, I think it is better to be explicit with different tests, with `forceRequireReturn` and `forceReturnsWithAsync` being of course somewhat different. (And in the case of async arrow functions, we do want to cover that is added back along with the few others.)

golopot:
Since `forceReturnsWithAsync` only affects this rule, it should be moved to options?

AndrewLeedham:
> Since `forceReturnsWithAsync` only affects this rule, it should be moved to options?

Makes sense to me. Implemented.

test/rules/assertions/requireReturns.js
```diff
@@ -134,6 +145,33 @@ export default {
...
+      errors: [
+        {
+          line: 2,
+          message: 'Missing JSDoc @returns declaration.'
+        }
+      ],
+      parserOptions: {
+        ecmaVersion: 8
+      },
+      settings: {
```
golopot: This should be options. (and other places.)

brettz9: `forceRequireReturn` wasn't a part of this PR, so those are still there are settings, but my recent commit did add one bad `forceRequiresWithAsync` which I've now amended, thanks!

PR #282, opened 2019-06-20
require-returns-check: Skip interface class methods

l1bbcsg:
As per jsdoc `@interface` tag [description](https://devdocs.io/jsdoc/tags-interface), class marked with this tag should denote that all members of the class are interface symbols and therefore functions can have `@returns` tag, but no function body with `return` statement.

This pull request makes the following code valid:

```js
/**
 * @interface
 */
class Foo {
    /**
     * @returns {string}
     */
    bar () {}
}
```

While previously `require-returns-check` reported `bar` as missing return value;

### DIFF
--- a/test/rules/assertions/requireReturnsCheck.js
+++ b/test/rules/assertions/requireReturnsCheck.js
@@ -89,6 +89,23 @@ export default {
           message: 'JSDoc @returns declaration present but return expression not available in function.'
         }
       ]
+    },
+    {
+      code: `
+          class Foo {
+            /**
+             * @returns {string}
+             */
+            bar () {
+            }
+          }
+      `,
+      errors: [
+        {
+          line: 3,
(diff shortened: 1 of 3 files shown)


test/rules/assertions/requireReturnsCheck.js
```diff
@@ -223,6 +223,20 @@ export default {
...
+           */
+          class Foo {
+            /**
+             * @returns {string}
+             */
+            bar () {
+            }
+          }
+      `
+    },
```
brettz9: If you can also add a corresponding invalid example (where there is no `@interface`), so we can see the contrast, I think this will be ready to merge.

l1bbcsg: Makes sense, added an invalid test case.

PR #283, opened 2019-06-20
no-undefined-types: Treat GCC generic types as defined

l1bbcsg:
This pull request adds support for Google Closure Compiler [Generic (aka Template) types](https://github.com/google/closure-compiler/wiki/Generic-Types) to `no-undefined-types`.

Makes the following code valid:

```js
/**
 * @template TEMPLATE_TYPE
 */
class Foo {
    /**
     * @return {TEMPLATE_TYPE}
     */
    bar () {
    }
}
```
While previously `no-undefined-types` reported `TEMPLATE_TYPE` in `bar` as undefined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2784,6 +2784,35 @@ function quux(foo, bar, baz) {
 // Settings: {"jsdoc":{"preferredTypes":{"hertype":{"replacement":false},"histype":"HisType"}}}
 // Options: [{"definedTypes":["MyType"],"preferredTypesDefined":true}]
 // Message: The type 'HerType' is undefined.
+
+class Foo {
+  /**
+   * @return {TEMPLATE_TYPE}
+   */
+  bar () {
+  }
+}
+// Message: The type 'TEMPLATE_TYPE' is undefined.
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -2839,6 +2839,29 @@ function quux(foo, bar, baz) {
...
+
+/**
+ * @template TEMPLATE_TYPE_A, TEMPLATE_TYPE_B
+ */
+class Foo {
+  /**
+   * @param {TEMPLATE_TYPE_A} baz
+   * @return {TEMPLATE_TYPE_B}
+   */
+  bar (baz) {
```
brettz9: Please add an invalid example corresponding to this one as well, though without `@template` (and with the appropriate error).

l1bbcsg: I originally thought this might be redundant since all invalid types can essentially be treated as generic types with missing `@template` tag and thus covered by existing tests.

Still, I added such a test case and also one that check that generic types are scoped correctly to their class and not treated as globals as a potential invalid implementation could do.

PR #284, opened 2019-06-21
fix: make options schemas effective

golopot:
This pr moves misplaced `schema`s to inside of `meta`. Otherwise these schemas are ineffective, that is, eslint engine does not validate options.


brettz9:
🤦 Thanks!

PR #179, opened 2019-03-22
Support TypeScript parameter properties

rasenplanscher:
#177

PR #285, opened 2019-06-21
Support TypeScript parameter properties

### LINKED ISSUE #177: [bug] TypeScript "Parameter properties" unsupported
Linting this file:
```
export class SomeClass {
    /**
     * @param property
     */
    constructor(private property: string) {}
}
```

`eslint` aborts with this Error:
> Error: Unsupported function signature format.

The stack trace points to at `eslint-plugin-jsdoc\dist\jsdocUtils.js:36:11`

Versions used:
+  eslint: 5.15.3
+  eslint-plugin-jsdoc: 4.4.3

For the feature's docs, see https://www.typescriptlang.org/docs/handbook/classes.html#parameter-properties

jgoz:
Fixes #177 
Closes #179 

This is the fix from #179 without the infrastructure changes.

### DIFF
--- a/src/jsdocUtils.js
+++ b/src/jsdocUtils.js
@@ -2,7 +2,7 @@ import _ from 'lodash';
 import tagNames from './tagNames';
 
 const getFunctionParameterNames = (functionNode : Object) : Array<string> => {
-  const functionParameterNameMapper = (param) => {
+  const getParamName = (param) => {
     if (_.has(param, 'name')) {
       return param.name;
     }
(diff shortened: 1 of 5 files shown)


src/jsdocUtils.js
```diff
@@ -23,8 +23,14 @@ const getFunctionParameterNames = (functionNode : Object) : Array<string> => {
...
 
+    if (param.type === 'TSParameterProperty') {
+      return functionParameterNameMapper(param.parameter);
+    }
+
     throw new Error('Unsupported function signature format.');
-  });
+  };
+
+  return _.map(functionNode.params, functionParameterNameMapper);
```
golopot:
```suggestion
  return functionNode.params.map(getParamName)
```
We are moving away from lodash.

src/jsdocUtils.js
```diff
@@ -23,8 +23,14 @@ const getFunctionParameterNames = (functionNode : Object) : Array<string> => {
       return param.argument.name;
     }
 
+    if (param.type === 'TSParameterProperty') {
+      return functionParameterNameMapper(param.parameter);
```
golopot:
```suggestion
      return getParamName(param.parameter);
```

PR #287, opened 2019-06-22
ESLint 6

brettz9:
- fix(require-jsdoc): with eslint 6, we can't use schema for defaults, so revert to old approach
- chore(linting): update eslint devDep
- testing: avoid timeout (not an issue anymore with eslint 6)
- testing(check-examples): use absolute paths with `matchingFileName` (used with `getConfigForFile`/`Linter`) per eslint 6
- testing(check-examples): use a simple `matchingFileName` file which doesn't have problem with loading `babel-eslint` in eslint 6 as does our config
- testing(require-param): use absolute paths with `parser` per eslint 6

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ package-lock.json
 !.npmignore
 !.README
 !.travis.yml
+!.ncurc.js
--- a/.ncurc.js
+++ b/.ncurc.js
@@ -1,0 +1,6 @@
+module.exports = {
+  // Whitelist all for checking besides `peer` which indicates
+  //   somewhat older versions of `eslint` we still support even
+  //   while our devDeps point to a more recent version
+  "dep": "prod,dev,optional,bundle"
+};
--- a/.ncurc.js
+++ b/.ncurc.js
@@ -1,6 +1,0 @@
-module.exports = {
-  // Whitelist all for checking besides `peer` which indicates
-  //   somewhat older versions of `eslint` we still support even
-  //   while our devDeps point to a more recent version
-  "dep": "prod,dev,optional,bundle"
-};
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: node_js
 node_js:
   - node
+  - 10
   - 8
-  - 6
 before_install:
   - npm config set depth 0
 notifications:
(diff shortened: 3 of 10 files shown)


brettz9:
I haven't addressed the Travis builds. (The npm update should be fine though as it is just a devDep). However, if we do eventually insist on users upgrading to eslint 6, we'll want to update `peerDependencies`.

Speaking of which, I just added a commit here with `.ncurc.js` file to help avoid alerting of the need to update `peerDependencies` since we're deliberately supporting some previous `eslint` versions (minimizes distractions when running `npm-check-updates` (or `ncu`) to find dependencies in need of update).

brettz9:
I bumped `.travis.yml` to use node 10 over 6. If you want to play with conditional downgrading of eslint (using `matrix`?), if that's all right, I'll leave it to you, @golopot , to do as you have time/inclination as a separate commit?

golopot:
There would be not need to test node 6 if we decide to drop node 6 support. Do you have strong opinion about node 6 support? I am ok either way.

brettz9:
No need as I see it...So I'll just go ahead and merge.

brettz9:
I'll update `peerDependencies` too then... It isn't strictly obligatory, but users should be warned if we aren't confirming against that version.

golopot:
The peerDependancy  would be `"eslint": ">=5"` ?

brettz9:
I was thinking bumping it to 6.0.0. Since our devDep has been bumped to 6.0.0 (the latest stable version), travis will only be testing that version unless we add some routines to install a lower version and run the tests against that. The `peerDependencies` won't block use of eslint < 6, but it will give a warning.

brettz9:
Btw, if we support eslint < 6, we don't need to bump Node from 6 (though requiring Node 8 has the  advantage that we can use more up to date ES features as Babel doesn't polyfill globals and we're not importing `core-js` or such to do so).

golopot:
I pull together a `.travis.yml` that will test `eslint 5 + node 6~12` and `eslint 6 + node 8~12`. You can cherry-pick it and worry about the version support decision later.

https://github.com/golopot/eslint-plugin-jsdoc/commit/56fa3d6cb216f54c0b9cdddb14c1c882d0207a32

https://travis-ci.org/golopot/eslint-plugin-jsdoc/builds/549088533

brettz9:
This is great... I just added back the `.ncurc` file since we have lowered peer deps again... So *now* I plan to merge...

PR #291, opened 2019-06-24
Add @template tag to no-undefined-types docs

l1bbcsg:
Essentially a follow up to #283, I noticed I forgot to update documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2706,7 +2706,7 @@ In addition to considering globals found in code (or in ESLint-indicated
 name(path) definitions to also serve as a potential "type" for checking
 the tag types in the table below:
 
-`@callback`, `@class` (or `@constructor`), `@constant` (or `@const`), `@event`, `@external` (or `@host`), `@function` (or `@func` or `@method`), `@interface`, `@member` (or `@var`), `@mixin`, `@name`, `@namespace`, `@typedef`.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -678,7 +678,7 @@ function quux () {}
 function quux2 () {
 
 }
-// Settings: {"jsdoc":{"matchingFileName":"/Users/brett/eslint-plugin-jsdoc/test/rules/data/test.js"}}
```
l1bbcsg: Oh, looks like a local path snuck in here. Not sure what I can do about it in this PR, this has to be updated in upstream.

l1bbcsg: Actually, I did the nasty thing and dropped out the commit with a force push. This way it's not in the diff at least and won't get in the way once the path reference is removed in upstream.

brettz9: Yeah, ESLint 6 requires absolute paths, and it seems the documentation generator resolves the path to the final result. For now, I can just rebase your PR and regenerate the docs on my machine to avoid your own path getting in there.

brettz9: Your force push solution is fine. Could you add this in parentheses though after `@template`: " (for Closure)"?

brettz9: or " (for Closure/TypeScript)"... It is not a standard jsdoc tag, so I think that should be called out for those who might otherwise try to search for the tag on https://jsdoc.app and not find it...

brettz9: Nevermind, I handled it (and changed the commit message to "docs:"). Thanks for the docs!

brettz9:
Sorry for the disconnected rebase, but this is now included in `master` through 4383525b1b03ab95200d68eb1bd136a11ef2d5b9

Issue #293, opened 2019-06-24
Validation for description-like comments?


adi-zz:
I am able to validate the description of a `function` using [`match-description`](https://github.com/gajus/eslint-plugin-jsdoc#match-description).

```
/** This function description is checked by eslint-plugin-jsdoc. */
function id(i) { return i }

```
But is there a way to validate other (sort of) descriptions using the same reg exp I declared in `match-description`?
```
/** Can eslint-plugin-jsdoc check this class description ? */
class MyClass {
  /** Can eslint-plugin-jsdoc check this field description? */
  myClassField = 1
}

const myObject = {
  /** Can eslint-plugin-jsdoc check this property description? */
  myProp: true
}

/** Can eslint-plugin-jsdoc check this interface description? */
interface MyTypescriptInterface {
}
```

brettz9:
I've submitted #294 which would allow you to add an option to this rule, `{contexts: ['ClassDeclaration']}` and get linting for classes.

From my testing, ESLint does not appear to allow the context to be `Property` or `VariableDeclaration`; it seems to only handle the likes of functions and class declarations, so we cannot apparently support it for property descriptions at this time.

As far as TypeScript, when I tried the following (using the types referencing interfaces at https://github.com/typescript-eslint/typescript-eslint/blob/861844d727a2ba7f75cdbba2571498078aee2671/packages/typescript-estree/src/ts-estree/ast-node-types.ts and testing at https://runkit.com/brettz9/5d11892d0ed30f0013311bbe to confirm `TSInterfaceDeclaration` was indeed the type needed), I wasn't able to get it working there either.

```js
{
      code: `
          /**
           * foo.
           */
          interface quux {

          }
      `,
      errors: [
        {
          line: 3,
          message: 'JSDoc description does not satisfy the regex pattern.'
        }
      ],
      options: [
        {
          contexts: [
            'TSInterfaceBody',
            'TSInterfaceDeclaration',
            'TSInterfaceHeritage'
          ],
          noDefaults: true
        }
      ],
      parser: require.resolve('@typescript-eslint/parser')
    }
```

Likewise no success with `ClassProperty` testing against your `myClassField = 1`.

[...]

PR #294, opened 2019-06-25
match-description options

brettz9:
- BREAKING CHANGE(require-description): remove `noDefaults` option and change `contexts` to always override defaults
- feat(match-description): allow `mainDescription: string|boolean` to override or disable main description separate from default
- feat(match-description): add `contexts` options for control on which contexts the rules apply
- feat(match-description): report line number with `match-description`
- feat(match-description): allow reporting multiple errors when main description validation fails
- feat(require-jsdoc): add `contexts` option to allow working with any node type
- fix(match-description, require-description): allow `contexts` to work with any node type
- fix(match-description): tighten default regex to require punctuation at the end even if only a single character

### DIFF
--- a/.README/rules/match-description.md
+++ b/.README/rules/match-description.md
@@ -71,7 +71,7 @@ it by setting it to `false`.
 
 ##### `contexts`
 
-Set this to a string or array of strings representing the AST context
+Set this to an array of strings representing the AST context
 where you wish the rule to be applied (e.g., `ClassDeclaration` for ES6 classes).
 Overrides the defaults.
 
--- a/.README/rules/match-description.md
+++ b/.README/rules/match-description.md
@@ -6,7 +6,7 @@ The default is this basic expression to match English sentences (Support
 for Unicode upper case may be added in a future version when it can be handled
 by our supported Node versions):
 
-``^([A-Z]|[`\\d_])([\\s\\S]*[.?!`])?$``
+``^([A-Z]|[`\\d_])[\\s\\S]*[.?!`]$``
 
 #### Options
 
--- a/.README/rules/match-description.md
+++ b/.README/rules/match-description.md
@@ -51,19 +51,21 @@ tag should be linted with the `matchDescription` value (or the default).
 ```
 
 If you wish to override the main function description without changing the
-default `mainDescription`, you may use `tags` with `main description`:
+default `match-description`, you may use `mainDescription`:
 
 ```js
 {
-  'jsdoc/match-description': ['error', {tags: {
-    'main description': '[A-Z].*\\.',
-    param: true,
-    returns: true
-  }}]
+  'jsdoc/match-description': ['error', {
+    mainDescription: '[A-Z].*\\.',
+    tags: {
+      param: true,
+      returns: true
+    }
+  }]
 }
 ```
 
-There is no need to add `"main description": true`, as by default, the main
+There is no need to add `mainDescription: true`, as by default, the main
 function (and only the main function) is linted, though you may disable checking
 it by setting it to `false`.
 
--- a/.README/rules/match-description.md
+++ b/.README/rules/match-description.md
@@ -50,14 +50,39 @@ tag should be linted with the `matchDescription` value (or the default).
 }
 ```
 
+If you wish to override the main function description without changing the
+default `mainDescription`, you may use `tags` with `main description`:
 
-By default, only the main function description is linted.
+```js
+{
+  'jsdoc/match-description': ['error', {tags: {
+    'main description': '[A-Z].*\\.',
+    param: true,
+    returns: true
+  }}]
+}
+```
+
+There is no need to add `"main description": true`, as by default, the main
+function (and only the main function) is linted, though you may disable checking
+it by setting it to `false`.
+
+##### `contexts`
+
+Set this to a string or array of strings representing the AST context
+where you wish the rule to be applied (e.g., `ClassDeclaration` for ES6 classes).
+
+##### `noDefaults`
+
+By default, `contexts` will permit `ArrowFunctionExpression`,
+`FunctionDeclaration`, and `FunctionExpression`. Set this instead to `true` to
+have `contexts` override these.
 
 |||
 |---|---|
-|Context|`ArrowFunctionExpression`, `FunctionDeclaration`, `FunctionExpression`|
+|Context|`ArrowFunctionExpression`, `FunctionDeclaration`, `FunctionExpression`; others when `contexts` option enabled|
 |Tags|N/A by default but see `tags` options|
 |Settings||
-|Options|`tags` (allows for 'param', 'arg', 'argument', 'returns', 'return'), `matchDescription`|
+|Options|`contexts`, `noDefaults`, `tags` (allows for 'param', 'arg', 'argument', 'returns', 'return'), `matchDescription`|
 
 <!-- assertions matchDescription -->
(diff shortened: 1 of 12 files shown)


test/rules/assertions/matchDescription.js
```diff
@@ -41,11 +43,34 @@ export default {
...
+      `,
+      errors: [
+        {
+          line: 3,
+          message: 'JSDoc description does not satisfy the regex pattern.'
+        }
+      ],
+      options: [{
+        tags: {
+          'main description': '[\u0410-\u042F][\u0410-\u044F]+\\.',
```
golopot: 1. Shouldn't it be in camelcase? 
2. Imo it is better to place the `mainDescription` in the top level of the option. Since a) `mainDescription` is not a tag name, b) Putting mainDescription in top-level allows schema validation to prevent typo.

src/rules/matchDescription.js
```diff
@@ -47,18 +52,36 @@ export default iterateJsdoc(({
...
+            oneOf: [
+              {
+                items: {
+                  type: 'string'
+                },
+                type: 'array'
+              },
+              {
+                type: 'string'
+              }
```
golopot: I prefer allowing only `array` to keep it simple.

src/rules/matchDescription.js
```diff
@@ -47,18 +52,36 @@ export default iterateJsdoc(({
...
+            ]
+          },
           matchDescription: {
             format: 'regex',
             type: 'string'
           },
+          noDefaults: {
```
golopot: Why this option? Why not just setting `contexts`?

.README/rules/require-jsdoc.md
```diff
@@ -34,6 +34,11 @@ be checked by the rule.
   - `FunctionExpression`
   - `MethodDefinition`
 
+- `contexts` - Set this to a string or array of strings representing the additional
```
golopot: This belongs to another pr.

test/rules/assertions/matchDescription.js
brettz9: I chose "main description" as a special case as users couldn't have a tag with this name as it has a space in it, but I see your point re: allowing schema validation (even though it adds a little extra complexity to the code).

src/rules/matchDescription.js
brettz9: I prefer accepting string for simplicity to the user. :) @gajus ?

src/rules/matchDescription.js
brettz9: It is extra work for users. I believe people will almost always want to keep the defaults. Also, that is the existing behavior for `require-description`.

.README/rules/require-jsdoc.md
brettz9: I can move it to another PR, but note that it all leverages the same code.

brettz9:
`contexts` is very useful as one can, as per the request in #293 , ensure that one has valid descriptions, e.g., above properties--not only the default behavior of matching descriptions on functions only.

golopot:
I mean the rule should apply to jsdoc attached to all node types and there is no configuration to include or exclude.

brettz9:
I think it is better to be under user control, whether for performance or just because projects differ from one another in their needs; but if you want to expand the defaults, I can change to support all node types by default.

test/rules/assertions/matchDescription.js
brettz9: Done, thanks!

src/rules/matchDescription.js
gajus: Please only accept `array`. Multiple APIs create ambiguity about how to supply multiple parameters.

.README/rules/match-description.md
```diff
@@ -41,14 +50,39 @@ tag should be linted with the `matchDescription` value (or the default).
...
+}
+```
+
+There is no need to add `"main description": true`, as by default, the main
+function (and only the main function) is linted, though you may disable checking
+it by setting it to `false`.
+
+##### `contexts`
+
+Set this to a string or array of strings representing the AST context
```
gajus: Just array of strings.

src/rules/matchDescription.js
gajus: > Why this option? Isn't it more intuitive to let provided contexts to override the default context?

Agreed. Otherwise, I need to be aware of what the defaults are. `contexts` should just override the default.

brettz9: There have been a few rules already added using this approach, e.g., `require-description`, `check-types`. Revert them as well?

brettz9: (`check-types` is somewhat different in referring to avoid the defaults for your preferred types, while `require-description` is more like as in this PR.)

brettz9: I went ahead and added a commit to make a breaking change to remove `noDefaults` from `require-description` (and for `match-description`) though I haven't done so for `check-types` unless you want to confirm. Let me know if that is ok.

src/rules/matchDescription.js
brettz9: Ok, done.

.README/rules/match-description.md
brettz9: Ok, sure, done. Also for `require-jsdoc` and `require-description`.

src/rules/matchDescription.js
brettz9: @gajus : Since your approval kind of overlapped with my reversion, I just wanted to confirm--are you happy with my breaking change to undo `require-description`'s use of `noDefaults` (in addition to removing it in `match-description`)?

golopot:
In the future please keep unrelated changes in different pull requests. Stuffing all changes in a pr 1) makes it much harder to review, 2) make the reviewer unable to accept one part of change and reject another.

brettz9:
This PR has now been narrowed somewhat, and I think it is ready for merging, @gajus , if you could respond to my comments ending at https://github.com/gajus/eslint-plugin-jsdoc/pull/294#discussion_r297063836 mentioning how I added a breaking change that I understood you wanted around the time you approved this PR, so I wanted to confirm that breaking change was indeed ok.

brettz9:
If y'all are ready for rebasing, sure.

src/rules/requireJsdoc.js
```diff
@@ -7,6 +7,12 @@ import getJSDocComment from '../eslint/getJSDocComment';
 const OPTIONS_SCHEMA = {
   additionalProperties: false,
   properties: {
+    contexts: {
```
golopot: This option overlapped the option `require`?

brettz9: Good catch. Yes, I avoided a breaking change here in order to allow `require-jsdoc` to offer more arbitrary contexts while also phasing in the API used by `require-description` and `match-description`. I can make a breaking change to remove `require` in favor of `contexts` if you and @gajus wish. I think it is a good idea.

golopot: I am ok either way.

gajus: Regarding breaking-changes: I am Okay with as many breaking changes as you need. Far bigger preference for clean, unambiguous APIs without legacy remnants over how often a breaking-change is made. This seems like an unpopular opinion in JavaScript land, but in my opinion a greater number of breaking changes over time is better (as it allows more gradual adoption, one version at a time) than one major planned breaking change that batches together many unrelated API changes and requires significant time investment to upgrade.

brettz9: Yes, it is jarring at first when one sees a "major" change and expects it will require a huge number of changes, but if people get to know the repository/package (or note how large its major version already has become), and how it may do frequent though incremental breaking releases, the angst, as well as the incremental effort is less.

PR #295, opened 2019-06-25
Fix(eslint): Add object expression in getJSDocComment 

fa93hws:
ESLint team suggest to have this fix in this repo.
https://github.com/eslint/eslint/pull/11897#issuecomment-505281683


fa93hws:
Looks like the CI is using `eslint` 6 for `eslint` 5 during the test because I import something from the `eslint`.
Do you have any idea how to test it properly?

brettz9:
Btw, @gajus , as this could lead to other sinon tests being added in the future, are you ok with the choice of sinon for spy testing? It seems pretty standard fare, but thought I'd confirm with you on that kind of choice, however small.

fa93hws:
> Btw, @gajus , as this could lead to other sinon tests being added in the future, are you ok with the choice of sinon for spy testing? It seems pretty standard fare, but thought I'd confirm with you on that kind of choice, however small.

Oh sorry, `sinon` is a mistake..
I forget to remove it😂

gajus:
> Btw, @gajus , as this could lead to other sinon tests being added in the future, are you ok with the choice of sinon for spy testing? It seems pretty standard fare, but thought I'd confirm with you on that kind of choice, however small.

Yes, of course.

fa93hws:
> Does this pr change the behavior of any rule? If so please add test to the rule. If not, this pr does not provide much value.

If that's the case, is that possible to export `getJSDocComment` from the index file, which gives the consumer the option to write their own rules given the fact that it is an API in `eslint` but being marked as deprecated.

golopot:
> If that's the case, is that possible to export getJSDocComment from the index file, which gives the consumer the option to write their own rules given the fact that it is an API in eslint but being marked as deprecated.

It is out of scope of this plugin.

brettz9:
@golopot , while expressing our opinions, let's please leave policy decisions to @gajus' on whether something is in scope or not.

@fa93hws , note that it is at least already available as `import getJSDocComment from "eslint-plugin-jsdoc/dist/eslint/getJSDocComment";` (Node/CommonJS).

brettz9:
And yes it does change the behavior of rules--I can add a test.

fa93hws:
> @fa93hws , note that it is at least already available as `import getJSDocComment from "eslint-plugin-jsdoc/dist/eslint/getJSDocComment";` (Node/CommonJS).

I will be OK with importing from `dist` 👍

gajus:
@brettz9 Would it make sense to move `getJSDocComment` to the https://github.com/jsdoctypeparser/jsdoctypeparser package out of `eslint-plugin-jsdoc`?

Alternatively, could create `eslint-ast-jsdoc-utilities` package.

I agree with @golopot that exporting `getJSDocComment` from _this_ package blurs the responsibility of the package.

fa93hws:
> @brettz9 Would it make sense to move `getJSDocComment` to the https://github.com/jsdoctypeparser/jsdoctypeparser package out of `eslint-plugin-jsdoc`?
> 
> Alternatively, could create `eslint-ast-jsdoc-utilities` package.
> 
> I agree with @golopot that exporting `getJSDocComment` from _this_ package blurs the responsibility of the package.

That make much more sense, many thanks!

brettz9:
@gajus: Yup, that sounds good to me, too (either way)... Or maybe as part of `comment-parser` if its author is open to it, since `getJSDocComment` functionality is really about whole jsdoc blocks rather than jsdoc types...

brettz9:
As this is just a fix, and I have added tests now which demonstrate the utility of the fix of this PR--and a few closely related I tacked on (to fix `ClassExpression` and default to checking immediately above any node so that if the user specifies such a node, there will be a comment to check). After merging, I can reduce the size of PR #294 by removing this fix which I had already added there (though without the separate `getJSDocComment` tests that this PR adds).

@golopot : If you want to move `getJSDocComment` to its own repo, I suggest filing a separate issue.

PR #302, opened 2019-06-30
Allow rejecting normally valid tag names and/or providing custom error messages when suggesting tags to reject or replace

### LINKED ISSUE #108: Allow user to configure `check-tag-names` to report otherwise valid tag names (e.g., report `@todo`)
~If it is possible to tweak existing ESLint rules, it'd be great to have `no-warning-comments` optionally catch `@todo` style comments, and if not, it'd be nice to have a separate rule to do so.~

Rather than limiting this to todo's, I figure `check-tag-names` could--when the `tagNamePreference` setting map had a tag set to `false`--report a normally valid tag. This would be useful, e.g., for reporting `@todo`. This would mirror how we do things in `preferredTypes` (it can also accept `false`).

Ideally, like `preferredTypes`, we could also allow `message` and `replacement` properties (along with template variables within the `message`), to allow a custom message (while still optionally allowing fixing, if a different tag were preferred), e.g., so that the ESLint error for "todo" could be something like "Please resolve to-dos or add to the tracker" or for preferring `@extends` over `@augments`: "`@extends` is preferred as it is more evocative of classes than `@augments`".

brettz9:
feat(check-tag-names): allow rejecting normally valid tag names and/or providing custom error messages when suggesting tags to reject or replace (fixes #108)

Reporting normally valid tags may be of interest for tags like `@todo`. This is implemented by allowing the user to set targeted `tagNamePreference` tags to `false` or to an object with only a `message` and no `replacement`

Custom messages for `check-tag-names` are implemented by allowing the user to set  targeted `tagNamePreference` tags to an object with `message` and `replacement` (mirroring `preferredTypes` behavior). Note that for other rules leveraging `tagNamePreference` (via `utils.getPreferredTagName`) to find the user's preferred tag name, the `replacement` will be used in the likes of error messages but not the `message`.

Also, for various (param, return, and description-related) rules which have used `tagNamePreference` (via the `utils.getPreferredTagName` utility), report to user if they have blocked (probably inadvertently) and not indicated a replacement for the relevant tag needed by the rule in the `tagNamePreference` setting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,60 @@ Use `settings.jsdoc.tagNamePreference` to configure a preferred alias name for a
 }
 ```
 
+One may also use an object with a `message` and `replacement`, with `{{tagName}}` and `{{preferredType}}` (or its alias `{{replacement}}`) as template variables to be
+substituted within `message`.
+
+The following will report the message `@extends is to be used over @augments as it is more evocative of classes than @augments` upon encountering `@augments`.
+
+```json
+{
+    "rules": {},
+    "settings": {
+        "jsdoc": {
+            "tagNamePreference": {
+                "augments": {
+                  "message": "@{{replacement}} is to be used over @{{tagName}} as it is more evocative of classes than @augments",
+                  "replacement": "extends"
+                }
+            }
+        }
+    }
+}
+```
+
+If one wishes to reject a normally valid tag, e.g., `@todo`, one may set the tag to `false`:
+
+```json
+{
+    "rules": {},
+    "settings": {
+        "jsdoc": {
+            "tagNamePreference": {
+                "todo": false
+            }
+        }
+    }
+}
+```
+
+Or one may set the targeted tag to an object with a custom `message`, but without a `replacement` property:
+
+```json
+{
+    "rules": {},
+    "settings": {
+        "jsdoc": {
+            "tagNamePreference": {
+                "todo": {
+                  "message": "We expect immediate perfection, so don't leave to-dos in your code."
+                }
+            }
+        }
+    }
+}
+```
+
+
 The defaults in `eslint-plugin-jsdoc` (for tags which offer
 aliases) are as follows:
 
@@ -955,6 +1009,15 @@ export class SomeClass {
   constructor(private property: string) {}
 }
 // Message: Expected @param names to be "property". Got "prop".
+
+/**
+ * @param foo
+ */
+function quux (foo) {
+
+}
+// Settings: {"jsdoc":{"tagNamePreference":{"param":false}}}
+// Message: Unexpected tag `@param`
 ````
 
 The following patterns are not considered problems:
@@ -1276,6 +1339,42 @@ function quux (foo) {
 }
 // Settings: {"jsdoc":{"additionalTagNames":{"customTags":["bar"]}}}
 // Message: Invalid JSDoc tag name "baz".
+
+/**
+ * @todo
+ */
+function quux () {
+
+}
+// Settings: {"jsdoc":{"tagNamePreference":{"todo":false}}}
+// Message: Blacklisted tag found (`@todo`)
+
+/**
+ * @todo
+ */
+function quux () {
+
+}
+// Settings: {"jsdoc":{"tagNamePreference":{"todo":{"message":"Please resolve to-dos or add to the tracker"}}}}
+// Message: Please resolve to-dos or add to the tracker
+
+/**
+ * @todo
+ */
+function quux () {
+
+}
+// Settings: {"jsdoc":{"tagNamePreference":{"todo":{"message":"Please use {{replacement}} instead of {{tagName}}","replacement":"x-todo"}}}}
+// Message: Please use x-todo instead of todo
+
+/**
+ * @todo
+ */
+function quux () {
+
+}
+// Settings: {"jsdoc":{"tagNamePreference":{"todo":{"message":"Please use {{preferredTagName}} instead of {{tagName}}","replacement":"x-todo"}}}}
+// Message: Please use x-todo instead of todo
 ````
 
 The following patterns are not considered problems:
@@ -1390,6 +1489,13 @@ function quux (foo) {}
  *
  */
 function quux (foo) {
+
+}
+
+/**
+ * @todo
+ */
+function quux () {
 
 }
 ````
@@ -3631,6 +3737,24 @@ var quux = {
 };
 // Options: [{"contexts":["ObjectExpression"]}]
 // Message: Missing JSDoc @description declaration.
+
+/**
+ * @someDesc
+ */
+function quux () {
+
+}
+// Settings: {"jsdoc":{"tagNamePreference":{"description":{"message":"Please avoid `{{tagName}}`; use `{{replacement}}` instead","replacement":"someDesc"}}}}
+// Message: Missing JSDoc @someDesc description.
+
+/**
+ * @description
+ */
+function quux () {
+
+}
+// Settings: {"jsdoc":{"tagNamePreference":{"description":false}}}
+// Message: Unexpected tag `@description`
 ````
 
 The following patterns are not considered problems:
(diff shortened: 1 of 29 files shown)


src/rules/requireReturnsDescription.js
```diff
@@ -5,6 +5,9 @@ export default iterateJsdoc(({
   utils
 }) => {
   const targetTagName = utils.getPreferredTagName('returns');
+  if (!targetTagName) {
+    return;
```
golopot: This guard is duplicated 8 times, it can be avoided it if is placed in `utils.forEachTag()`.

test/rules/assertions/requireParamName.js
```diff
@@ -31,6 +31,28 @@ export default {
...
+          /**
+           * @param foo
+           */
+          function quux (foo) {
+
+          }
+      `,
+      errors: [
+        {
+          message: 'Unexpected tag `@param`'
```
golopot: This error is already reported by `check-tag-names`, seems like a single responsibility principle breakage for me. It also make `settings.preferredTagName` a dependency for these dozens of rules.

brettz9: I don't see that as a problem in this case. The preexisting code was checking this I think so it would avoid the need to check for all of the possible aliases, and eslint settings are intended to be shared across rules for such use cases as this. That test is to confirm that the behavior which occurs in `check-tag-names` (which reports an error message more precise to the rule's purpose as opposed to this example which is just reporting an unusual scenario where the user has blocked a parameter the rule depends on) also occurs in the rules which rely on it.

I suggest raising a separate issue if you still wish to remove the dependency as this is a preexisting behavior (though one which I think makes sense--otherwise, we'd have to check all possible aliases or force users to repeat their preferences for each rule).

src/rules/requireReturnsDescription.js
brettz9: Good point. I've changed `forEachTag` to `forEachPreferredTag`.

README.md
```diff
@@ -197,6 +197,60 @@ Use `settings.jsdoc.tagNamePreference` to configure a preferred alias name for a
 }
 ```
 
+One may also use an object with a `message` and `replacement`, with `{{tagName}}` and `{{preferredType}}` (or its alias `{{replacement}}`) as template variables to be
```
golopot: This template replacement thing looks like an excessive feature for me. In terms of UX it easier to understand and maintain a setting like this
```
"message": "@extends is preferred over @arguments."
``` 
than this
```
"message": "@{{replacement}} is preferred over @{{tagName}}."`
```

There might be rare cases where this feature helps, but the benefits cannot justify the complexity it brings.

test/rules/assertions/requireParamName.js
golopot: That make sense then.

golopot:
Just for curiosity, what is your opinion on https://gist.github.com/stephenparish/9941e89d80e2bc58a153#format-of-the-commit-message

> Any line of the commit message cannot be longer 100 characters! This allows the message to be easier to read on github as well as in various git tools.

brettz9:
If I'm being fully frank on this, I dislike the terse commit message format, I hate the overly confining restrictions of the release notes not picking up multiple items in a commit, and I don't like such a restriction on length.

I am not blind to the rationales used for these restrictions, but I personally find them very confining, as with the commit hooks which take time out to do testing and documentation when I have already tested locally (on top of the time it takes to push from my terrible internet service here), and I either forget to disable the commit hooks during testing, or if I get too used to disabling, I fear I will accidentally commit without letting them run.

I think for people who can gather their thoughts all at once, without going through a few iterations as I inevitably do, these systems may work well, and I see their rationale for enforcing things on a team, but just personally speaking, I would never want to use this on my own projects, at least not without a lot of customization. But again, I understand their utility and rationale for enforcing some order out of chaos that developers otherwise might try to commit.

README.md
brettz9: eslint already has templating in place, so there is not even a need for `.replace()` calls here--just pass the variables in. Also, I've already implemented this approach with `preferredTypes`.

If I am to undo then I think it ought to be a breaking change to remove from `preferredTypes` as well.

I agree though that in this case it doesn't add much, but I thought it should mirror behavior in `preferredTypes`. In `preferredTypes`, I'll admit that `replacement` and `badType` are not super useful there either (though they can keep consistency in messages across different replacements), but I figured it wasn't much to make those variables available since I already needed substitution variables for `tagName` and `tagValue` (both of which cannot be known ahead of time in the case of `preferredTypes` and which were therefore I think justified).

So, if I am to remove template variables in both, I'd be ok with removing the static ones (though requiring a breaking change), but not the dynamic ones.

golopot: What are static ones and dynamics ones?

brettz9: By "static", I just mean that they can be replaced by the user hard-coding the value:

- `check-types` / `settings.jsdoc.preferredTypes`: **Static**: `badType`, `preferredType`/`replacement`; **Dynamic**: `tagName`, `tagValue`
- `check-tag-names`/`settings.jsdoc.tagNamePreference`:  **All are static**:`preferredTagName`/`replacement`, `tagName`

(`tagName` is static in one and dynamic in the other since when checking types, the tag name may vary, but not when tags are what is being checked.)

golopot: That's good.

brettz9: So you mean it is good to remove the static ones and keep the dynamic ones?

golopot: Yes

brettz9: You know, on further reflection though, I'm not sure removing the static ones is such a good idea. Projects may sometimes change the type names they want to use, e.g., "Any", "*", "any", and with those "static" types, they avoid maintenance problems if there is a discrepancy between the `replacement` value and the replacement (or target) type used in the message.

For example, with this code:

```js
/**
  * @param {abc} foo
  */
 function qux(foo) {
 }
```

to get the following error (where one wanted "Abc" to be used in place of "abc"):

> Messed up JSDoc @param "foo" type "abc"; prefer: "Abc".

...with the following, if you wanted to replace `Abc` with some other preferred type (or if you were changing the type to target), you'd have to remember to make the switch in two places:

```js
          preferredTypes: {
            abc: {
              message: 'Messed up JSDoc @{{tagName}}{{tagValue}} type "abc"; prefer: "Abc".',
              replacement: 'Abc'
            }
          }
```

whereas the following doesn't suffer from that problem:

```js
          preferredTypes: {
            abc: {
              message: 'Messed up JSDoc @{{tagName}}{{tagValue}} type "{{badType}}"; prefer: "{{replacement}}".',
              replacement: 'Abc'
            }
          }
```

And passing in the "static" variables hardly takes up any coding lines at all (just passing the variable names in), and as mentioned eslint handles the replacements for us.

Maybe @gajus has an opinion?

golopot: The above comment seemed to be about `preferredTypes`, does it apply to `tagNamePreference`?

brettz9: Yes, it's the same principle... The "static" properties still allow avoiding of duplication which could be a problem if one changes the tag target (or replacement).

```js
      settings: {
        jsdoc: {
          tagNamePreference: {
            todo: {
              message: 'Please use {{preferredTagName}} instead of {{tagName}}',
              replacement: 'x-todo'
            }
          }
        }
      }
```

golopot: I still consider this a feature bloat that will be useful for very few people. Adding such a feature will negatively affect users because they have to read longer docs. And the developers have to maintain them.

brettz9: And I don't think it is much extra to "maintain" or read to provide that ability. So let's wait on @gajus . While we need to make code maintainable, I think we ought to err more on the side of being flexible for users.

gajus: > I still consider this a feature bloat that will be useful for very few people. Adding such a feature will negatively affect users because they have to read longer docs. And the developers have to maintain them.

I agree. Sounds like nuisance to everyone for the benefit of very few.

PR #303, opened 2019-06-30
Check tag names breaking

brettz9:
feat(check-tag-names): auto-allow tags in `tagNamePreference` setting to be defined; add `definedTags` option to replace `additionalTagNames.customTags` setting

BREAKING CHANGE:

Switch `additionalTagNames.customTags` setting to `definedTags` option (only used by one rule)

### DIFF
--- a/.README/rules/check-tag-names.md
+++ b/.README/rules/check-tag-names.md
@@ -75,10 +75,26 @@ version
 yields
 ```
 
+Note that the tags indicated as replacements in `settings.jsdoc.tagNamePreference` will automatically be considered as valid.
+
+#### Options
+
+##### `definedTags`
+
+Use an array of `definedTags` strings to configure additional, allowed JSDoc tags.
+The format is as follows:
+
+```json
+{
+  "definedTags": ["define", "record"]
+}
+```
+
 |||
 |---|---|
 |Context|everywhere|
 |Tags|N/A|
-|Settings|`tagNamePreference`, `additionalTagNames`|
+|Options|`definedTags`|
+|Settings|`tagNamePreference`|
 
 <!-- assertions checkTagNames -->
--- a/.README/README.md
+++ b/.README/README.md
@@ -199,6 +199,8 @@ Or one may set the targeted tag to an object with a custom `message`, but withou
 }
 ```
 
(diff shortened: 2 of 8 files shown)


.README/rules/check-tag-names.md
```diff
@@ -75,10 +75,28 @@ version
 yields
 ```
 
+#### Options
+
+##### `definedTags`
```
golopot: How about `customTags` or `additionalTags`? `definedTags` is fine though.

.README/rules/check-tag-names.md
```diff
@@ -75,10 +75,28 @@ version
...
+
+```json
+{
+  "definedTags": ["define", "record"]
+}
+```
+
+##### `preferredTagsDefined`
+
+If this boolean is set to `true`, the tags indicated as replacements in `settings.jsdoc.tagNamePreference` will automatically be considered as defined.
```
golopot: I don't think this needs to be an option. It does not make sense to set this to false.

.README/rules/check-tag-names.md
brettz9: I considered this as well, but I thought it'd be more consistent with `definedTypes` (used in the `no-undefined-types` rule).

.README/rules/check-tag-names.md
brettz9: Good point, and I think I can also remove `preferredTypesDefined` which was added earlier for the same reason. Do you want the removal in this same PR?

golopot: I like them to be in separate prs.  How about creating a new branch `11.0.0` and make it the target branch of these separate prs, and then when all things are done merge them together?

brettz9: Ok, sure.

brettz9: So I will remove `preferredTagsDefined` in this PR (i.e., avoid adding it at all) and remove `preferredTypesDefined` in the other.

brettz9: I've removed `preferredTagsDefined` from this PR and added #304 to remove `preferredTypesDefined`.

PR #304, opened 2019-07-01
Remove `preferredTypesDefined` option, applying it automatically instead

brettz9:
To add along with #302 and #303 to an `11.0.0` branch per https://github.com/gajus/eslint-plugin-jsdoc/pull/303#discussion_r298867281 before merging.

feat: make `preferredTypesDefined` option obsolete by applying the behavior automatically

BREAKING CHANGE:

One can no longer set `preferredTypesDefined` to `false` to avoid the `settings.jsdoc.preferredTypes` targets from being treated as defined.

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -315,9 +315,8 @@ but restricted to `@param`. These settings are now deprecated.
     `{{preferredType}}` (or `{{replacement}}`), noting that the latter is
     of no use when one is merely forbidding a type).
 
-If `no-undefined-types` has the option key `preferredTypesDefined` set to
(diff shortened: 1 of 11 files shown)


golopot:
It seems that this pr is accidental closed because the branch `gajus/11.0` is moved to `gajus/11.0.0`. Can you reopen it?

brettz9:
I had already added an 11.0.0 branch and rebased this onto that. It won't let me reopen as this is already a part of 11.0.0. There are no new changes, I have simply rebased, so if you want to confirm, please confirm on 11.0.0. I've also merged in #310 as you approved that as well.

golopot:
Ok, thanks.

PR #307, opened 2019-07-03
no-undefined-types: Handle GCC generic functions

l1bbcsg:
In addition to classes, GCC also has [Generic functions and methods](https://github.com/google/closure-compiler/wiki/Generic-Types#generic-functions-and-methods), which aren’t much different, they just have `@template` tag in their own comment node rather than that of class.

I couldn’t come up with a negative test case that wasn’t functionally identical to some existing one.


brettz9:
Looks good to me. But maybe for a negative test, you could one where the type in the param/return doesn't match the type in the `@template` (i.e., confirming that the presence of the template tag isn't by itself sufficient to suppress errors)?

l1bbcsg:
This makes sense, I added a test case with mismatched `@template` type.

Issue #306, opened 2019-07-03
GCC specific tags support


l1bbcsg:
Closure Compiler has a different set of tags from those in base JSDoc specification. The complete list can be found in [source code](https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/parsing/Annotation.java) and documentation for most at [their wiki](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler). Several of those tags are crucial for GCC projects and inhibit the usefulness of eslint-plugin-jsdoc for them somewhat (I’m personally stumbled by `@externs`, `@record` and `@dict`).

From my understanding, supporting Closure variation of JSDoc is one of the goals of this plugin, so I believe all of this tags should be treated as valid.

Perhaps they should not be part of default set of tags, but behind an option of sorts. Given that some other things like syntax and conventions are also different between core JSDoc and GCC maybe there should be a global option that would switch plugin rules between base jsdoc and GCC modes?

I can open up a PR adding this tags to the base list of tags, or should it be handled differently?

golopot:
Non-standard tag names can be added in settings, as per [readme](https://github.com/gajus/eslint-plugin-jsdoc#additional-tag-names):
```jsonc
// .eslintrc
{
    "rules": {},
    "settings": {
        "jsdoc": {
            "additionalTagNames": {
                "customTags": ["externs", "record", "dict"]
            }
        }
    }
}
```
I am not sure whether these GCC specific tags should be recognized out-of-box, it seemd harmless to add them though.

brettz9:
I think there enough differences in behavior between jsdoc, gcc, and typescript to justify a global setting. With a single global flag, not much complexity is added and besides the fact that it avoids being overly lenient for some modes or overly strict for others, a global setting would be nicely self-documenting and preferable to the use of comments to denote these exceptions, e.g., as I've added here when `@template had been allowed`: https://github.com/gajus/eslint-plugin-jsdoc/blob/master/src/tagNames.js#L93-L98

There are jsdoc projects where developers used to gcc may try to use these tags, and the project doesn't want to (and no doubt even more so if the mode is typescript and someone tries to use jsdoc, etc.)

We really need such a global setting, I think, if we're going to satisfactorily address these various use cases.

PR #308, opened 2019-07-04
valid-types: Add more type checks

### LINKED ISSUE #305: valid-types: Only checks the name of typedef but not the definition
The following comment has two types: typedef name (`Foo`) and its value (`String`):

```js
/**
 * @typedef {String} Foo
 */
```

`valid-types` only checks typedef name (`Foo`) for validity and not the value.

I debugged this a bit and figured `valid-types` rule when iterating over tags goes into [`isNamepathTag`](https://github.com/gajus/eslint-plugin-jsdoc/blob/4bb3af6927994c32f3f8fa6018cd324aa2199bf2/src/rules/validTypes.js#L69) execution branch while checking of they type value is done in `isTagWithType` branch. `@typedef` is essentially both and should go through both.

l1bbcsg:
Fixes #305 .

All of the problems mentioned in the issue are solved, but I'm not entirely sure the solution's perfect.

Namely, there's a lot of complexity because there are tags that *may* have a name, a type or a description and tags *must* have a name, a type or a description and each of this sets intersect with every other.

The changes introduced in this PR:

* The rule became stricter as it now checks more and reports more errors as a result. Some previously overlooked errors may now be reported as well.
* Some error messages changed, namely `Syntax error in type: ` (note the empty string) now became more specific `Tag @tag must have a type` or `Tag @tag must have a namepath`
* Parsing of `@see` tag has changed. Previously `@see {@link http://foo.bar}` was parsed as having type `@link http://foo.bar` which definitely not something that should happen. I made `{@link http://foo.bar}` a description.

The following became unintentionally valid:

```js
/**
 * @typedef {number|string}
 */
```

I think it's fine, at least as far as this rule is concerned. The type is valid, so is jsdoc itself.
The fact that it's missing typedef name via variable declaration could probably be handled by another rule. This was done to allow GCC syntax.

Also, this is valid with previous and current checks, but shouldn't be (`Array<string>` is not a valid identifier). This happens because both type and namepath are checked the same way via `jsdoctypeparser`.

```js
/**
 * @typedef {StringArray} Array<string>
 */
```


test/rules/assertions/validTypes.js
```diff
@@ -222,16 +290,6 @@ export default {
           }
       `
     },
-    {
```
l1bbcsg: I think this test case was mistakenly added. It’s identical to an [invalid test case](https://github.com/gajus/eslint-plugin-jsdoc/blob/f5c485da9f91f15da50f88c218a461252ffe5faf/test/rules/assertions/validTypes.js#L108) and they weren’t properly checked before.

test/rules/assertions/validTypes.js
```diff
@@ -245,7 +303,7 @@ export default {
     {
       code: `
           /**
-           * @callback
```
l1bbcsg: It’s also something that shouldn’t be valid as per spec: https://jsdoc.app/tags-callback.html

brettz9: Yes, I do see the spec seems to expect a name path, but I'm not so sure we want to enforce that. I would guess jsdoc will not complain if you omit it (and maybe jsdoc wouldn't even fail to build incoming links to it, if it can get the name from the function that it is describing). If you look at the `valid-types` docs, I put them into categories whereby tags (like `@callback` or `@event`) which may have potential documentation value are distinguished from those which wouldn't. `@callback`, even without a namepath, could be useful to signal that a function was intended to be used as a callback whereas a tag like `@typedef` wouldn't be useful without a namepath to reference it. I'm open to further consideration, but since jsdoc is not a strict validation spec, I'm not sure we should take its docs *too* much at their face value, unless perhaps TypeScript or GCC do such validation.

test/rules/assertions/validTypes.js
```diff
@@ -161,14 +161,82 @@ export default {
           }
       `,
       errors: [{
-        line: 3,
-        message: 'Syntax error in type: '
+        line: 2,
```
brettz9: This line number behavior is different than how others have been calculated. For consistency, it should be 3 here.

src/jsdocUtils.js
```diff
@@ -186,20 +186,26 @@ const potentiallyEmptyNamepathTags = [
...
+  'extends', 'augments',
   'constant', 'const',
   'function', 'func', 'method',
   'interface',
   'member', 'var',
   'mixin',
   'namespace',
-  'listens', 'fires', 'emits'
+  'listens', 'fires', 'emits',
+  'see',
```
brettz9: As mentioned below `@see` has special handling.

src/jsdocUtils.js
```diff
@@ -186,20 +186,26 @@ const potentiallyEmptyNamepathTags = [
   'event',
   'callback',
   'class', 'constructor',
+  'extends', 'augments',
```
brettz9: Why are these being added here? I don't think these make much sense to be empty.

src/iterateJsdoc.js
```diff
@@ -9,11 +9,21 @@ const parseComment = (commentNode, indent) => {
     // @see https://github.com/yavorskiy/comment-parser/issues/21
     parsers: [
       commentParser.PARSERS.parse_tag,
-      commentParser.PARSERS.parse_type,
+      (str, data) => {
+        if (data.tag === 'see') {
+          // @see can't contain types, only names or links which might be confused with types
+          return null;
```
brettz9: Ah, interesting. But we might want to see how jsdoc actually *behaves*. If you put in a type it may generate a link to it, despite what the docs say. The docs do refer to it referring to another "symbol" so it may be that a type is also included here.

src/jsdocUtils.js
```diff
@@ -186,20 +186,26 @@ const potentiallyEmptyNamepathTags = [
...
   'interface',
   'member', 'var',
   'mixin',
   'namespace',
-  'listens', 'fires', 'emits'
+  'listens', 'fires', 'emits',
+  'see',
+
+  // GCC syntax allows typedef to be named through variable declaration rather than jsdoc name
+  'typedef'
```
brettz9: Let's add a global `settings.jsdoc.mode` with possible values of "gcc"|"jsdoc"|"typescript" and use that to determine whether to allow here (if later versions of jsdoc change this, we can add values, e.g., "jsdoc4").

test/rules/assertions/validTypes.js
```diff
@@ -161,14 +161,82 @@ export default {
...
+    {
+      code: `
+          /**
+           * @typedef {string} UserStr%ng
+           */
+      `,
+      errors: [
+        {
+          line: 3,
+          message: 'Syntax error in type: UserStr%ng'
```
brettz9: Can we get this to refer to a "Syntax error in namepath" instead?

test/rules/assertions/validTypes.js
```diff
@@ -161,14 +161,82 @@ export default {
...
+           const FOO = 'foo';
+      `,
+      errors: [
+        {
+          line: 3,
+          message: 'Syntax error in type: str%ng'
+        }
+      ]
+    },
+    {
```
brettz9: Since `constant` allows an optional name per https://jsdoc.app/tags-constant.html, maybe you could also add a test for `@constant {type} name`?

src/jsdocUtils.js
brettz9: I see that the test below ends up insisting they are not empty but note that the variable name here is `potentiallyEmptyNamepathTags`, so I think a new variable may need to be added to avoid confusion.

l1bbcsg: The semantics here is that can omit namepath. `extends`/`augmens` can either contain a namepath or a type and must contain one of those.

I think the variables I used in the rule (`mustHaveType`, `mustHaveNamepath`, `mustHaveEither`) make more sense, but I was hesitant to change utility names.

If that is not a concer, I’d do something like `canHaveType`, `mustHaveType` and `mustHaveNamepathOrType` here. Or perhaps some configuration dictionary could be a good idea.

src/jsdocUtils.js
l1bbcsg: This is what I was hoping for in https://github.com/gajus/eslint-plugin-jsdoc/issues/306, but I think it should be done separately from this PR as it aims to fix a bug and introducing a big feature would clutter it and complicate accepting it.

test/rules/assertions/validTypes.js
l1bbcsg: Fixed this

src/jsdocUtils.js
brettz9: Ok, sure, makes sense, though let's make a mental note of that being one item relevant for targeting.

src/jsdocUtils.js
brettz9: I'd have to take a closer look, but I'm not against the PR changing the utility names in principle, as long as they are not being used (or are not expected to be used) with their old names/purpose.

src/iterateJsdoc.js
l1bbcsg: In that case perhaps it's better to be more lenient and completely ignore `@see {@link }` form at least until there's a piece of code that handles it specifically, this way types here will be treated as namepaths and validated the same.

src/jsdocUtils.js
l1bbcsg: I went ahead and renamed utils along with their usage (most was here in validTypes).
Now their naming starts with `tagMustHave`/`tagMightHave` and ends with `Type`,`Namepath`,`EitherTagOrNamepath`.
I also moved `checkSeesForNamepaths` handling out of utils since this is a specifically `valid-types` option.

src/jsdocUtils.js
l1bbcsg: Now 'see' is in `tagsWithOptionalNamepath` list as it may have a namepath or may not have it. Its handling is done within `valid-types` rule.

test/rules/assertions/validTypes.js
l1bbcsg: Isn't this the domain of `allowEptyNamepaths` flag?

I've added a valid test case with `@callback` having no namepath and `allowEmptyNamepaths` set to true.

test/rules/assertions/validTypes.js
l1bbcsg: Added. 

But I'm not entirely sure how that form is supposed to be used. What if I have different names in tag and variable declaration?

l1bbcsg:
Apologizes for dying on this PR, both work and vacation happened.

I updated it and addressed all comments I believe.

test/rules/assertions/validTypes.js
l1bbcsg: This warranted splitting `validTypeParsing` into two and a bit of duplicate code.

Also for the sake of coverage I had to remove `error.name === 'SyntaxError'` check, since `else` branch is now never executed in tests because no other kinds of errors are raised naturally. Previously on unexpected errors it would stay quiet, now it reports an error to eslint (potentially wrong if the exception was caused by mistake in code).

test/rules/assertions/validTypes.js
brettz9: I think, like `check-param-names`, name-checking could be useful within a new rule like `check-const-names` (or something more generic to work with `@member`/`@memberof!`, `@var`, and possibly `@namespace` as well as `@constant`/`@const`). However, for `valid-types`, I think just ensuring the rule doesn't break when a name is present would be fine.

test/rules/assertions/validTypes.js
brettz9: Yes, very good... You are right.

test/rules/assertions/validTypes.js
```diff
@@ -261,12 +329,25 @@ export default {
...
+          /**
+           * @callback foo
            */
           function quux() {
 
           }
       `,
+      options: [{
+        allowEmptyNamepaths: true,
+      }],
```
brettz9: It looks like you wanted to remove the namepath with @callback, no?

test/rules/assertions/validTypes.js
brettz9: So, I'll mark this as resolved, as it seems you've already added such an example.

src/jsdocUtils.js
```diff
@@ -126,17 +126,45 @@ const hasDefinedTypeReturnTag = (tag) => {
...
+  'type',
+  'typedef',
+  'yields', 'yield',
+
+  // GCC specific
+  'package',
+  'private',
+  'protected',
+  'public',
+  'static',
```
brettz9: Let's only add these if our new proposal for `settings.jsdoc.mode` is "gcc" (see #356 ).

test/rules/assertions/validTypes.js
```diff
@@ -180,12 +180,80 @@ export default {
...
+      code: `
+          /**
+           * @extends
+           */
+           class Bar {};
+      `,
+      errors: [
+        {
+          line: 3,
+          message: 'Tag @extends must have either a type or namepath',
```
brettz9: These error message changes are quite helpful, though in this particular case, I think it would be clearer to say `Tag @extends must have a type/namepath.` since it otherwise seems to suggest to me that these might be two separate components (like `@namespace {type} namepath`).

src/jsdocUtils.js
```diff
@@ -126,17 +126,45 @@ const hasDefinedTypeReturnTag = (tag) => {
   return true;
 };
 
+const tagsWithOptionalType = [
```
brettz9: I would prefer that this variable would not include those items within `tagsWithMandatoryType` and that `tagMightHaveType` would instead check within both `tagsWithMandatoryType` and `tagsWithOptionalType`.

src/jsdocUtils.js
```diff
@@ -147,95 +175,68 @@ const namepathDefiningTags = [
   'namespace',
 ];
 
-const namepathPointingTags = [
-  // NOT USEFUL WITHOUT NAMEPATH
+const tagsWithOptionalNamepath = [
+  ...namepathDefiningTags,
```
brettz9: As per my comment for `tagsWithOptionalType`, I would prefer that `tagsWithOptionalNamepath` does not include `namepathDefiningTags` but instead lists the non-mandatory items only, and that `tagMightHaveNamepath` checks both `tagMightHaveNamepath` and `tagsWithMandatoryNamepath`.

test/rules/assertions/validTypes.js
brettz9: Btw, I would think we should require `settings.jsdoc.mode` be set to `gcc` or `typescript` for  our existing test allowing `@extends {Foo<String>}` since the jsdoc spec doesn't indicate this format. In such a case, we can add one test which fails (with mode "jsdoc" or default) and one which passes (for "gcc" or "typescript").

test/rules/assertions/validTypes.js
```diff
@@ -329,5 +410,52 @@ export default {
...
+      `,
+    },
+    {
+      code: `
+          /**
+           * @extends {Foo<String>}
+           */
+           class Bar {};
+      `,
+    },
```
brettz9: I.e., for this test, this should only pass for `settings.jsdoc.mode` set to "gcc" or "typescript".

test/rules/assertions/validTypes.js
l1bbcsg: Yes, this was supposed to test the form without namepath. Fixed

src/jsdocUtils.js
l1bbcsg: Makes sense, done

src/jsdocUtils.js
l1bbcsg: A bit more complicated here because `namepathDefiningTags` is also used for other utility as well. And is happened, all of the mandatory namepath tags are namepath defining. I left those in both lists and made `tagMightHaveNamepath` check all three in case they change in the future.

test/rules/assertions/validTypes.js
l1bbcsg: > Btw, I would think we should require `settings.jsdoc.mode` be set to `gcc` 

True, this is template class syntax of GCC and not part of vanilla JSDoc. I'm not certain on typescript since I don't really know ts dialect of JSDoc.

src/jsdocUtils.js
l1bbcsg: When it hits, definitely. You think they should be removed from the list for now?
I believe it's better to leave them be for now not to break this bit of backwards compatibility.

Issue #301, opened 2019-06-30
`require-description` ought to accept jsdoc block description (or have option therefor)


brettz9:
The `require-description` rule requires an explicit `@description`.

This lack of acceptance for an implicit description (i.e., a jsdoc block description without a tag but with a description like `/** Some description */`) has two problems:

1. Projects may not care that an explicit `@description` is used but do want some kind of description enforced.
2. The approach is inconsistent with `require-description-complete-sentence` and `match-description` which check implicit jsdoc block descriptions only and the description part of returns/param tags (though I think these should also check explicit `@description`'s).

I think all three rules should support additional tags to which they apply as well, e.g., `@summary`, but that could be handled in #233.

PR #313, opened 2019-07-06
`match-description`

brettz9:
feat(match-description): allow `description` and `desc` tags to be matched for validation

Fixes part of #301. This expands the tags that can be validated through the existing `tags` option to include `@description`/`@desc` tags.

### DIFF
--- a/src/rules/matchDescription.js
+++ b/src/rules/matchDescription.js
@@ -53,9 +53,19 @@ export default iterateJsdoc(({
     return;
   }
 
+  const hasOptionTag = (tagName) => {
+    return {}.hasOwnProperty.call(options.tags, tagName) && options.tags[tagName];
+  };
+
+  utils.forEachPreferredTag('description', (matchingJsdocTag, targetTagName) => {
+    const description = (matchingJsdocTag.name + ' ' + matchingJsdocTag.description).trim();
+    if (hasOptionTag(targetTagName)) {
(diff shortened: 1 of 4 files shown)


src/rules/matchDescription.js
```diff
@@ -53,9 +53,19 @@ export default iterateJsdoc(({
     return;
   }
 
+  const hasOptionTag = (tagName) => {
+    return {}.hasOwnProperty.call(options.tags, tagName) && options.tags[tagName];
```
golopot:
```suggestion
    return !!options.tags[tagName];
```
The first part of the condition seems to be redundant.

brettz9: While it's pretty obscure and not needed here, I've gotten in the habit since Firefox (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Undefined_prop ) will, if its `about:config` setting `javascript.options.strict` is enabled (unrelated to "strict mode"), report missing property references as warnings in console, whereas the `hasOwnProperty` check will avoid this. But I can remove the check as it is indeed not really needed (unless someone is adding stuff on `Object.prototype`).

brettz9: I've adjusted per your suggestion, though using `Boolean()` to coerce instead of `!!` per the linting rules.

PR #315, opened 2019-07-06
`matchImplicit` option for `require-description`

### LINKED ISSUE #301: `require-description` ought to accept jsdoc block description (or have option therefor)

brettz9:
~feat(`require-description`): add `matchImplicit` option to accept implicit descriptions as satisfying the rule (fixes #301)~

feat(`require-description`): add `descriptionStyle` option to accept implicit descriptions as satisfying the rule by default (fixes #301)

BREAKING CHANGE:

To restore old behavior, set the option `descriptionStyle` to `"tag"`

----

Other alternatives would be to allow implicit descriptions (i.e., jsdoc blocks with any non-tag intro text) as satisfying this rule by default. An advantage of such an approach would be that it would more closely match how `match-description` and `require-description-complete-sentence` behave as they automatically check both implicit and explicit descriptions now.

But since people have been using the `require-description` rule so far to insist on an explicit `@description`, I think if such a breaking change were to be made to accept implicit descriptions, such a change should at least be accompanied by an option (maybe `matchExplicitOnly`) so people can at least opt back in to the old behavior of only accepting an explicit `@description`.

~In case you didn't want such a breaking change, I thought I'd start with a non-breaking version, though I should be able to easily adapt it to the breaking version if you prefer that instead.~ *Done.*

### DIFF
--- a/.README/rules/require-description.md
+++ b/.README/rules/require-description.md
@@ -2,25 +2,32 @@
 
 Requires that all functions have a description.
 
-* All functions must have a `@description` tag.
-* Every description tag must have a non-empty description that explains the purpose of the method.
+* All functions must have an implicit description or have the option
+  `descriptionStyle` set to `tag`.
+* Every jsdoc block description (or description tag if `descriptionStyle` is
+  `"tag"`) must have a non-empty description that explains the purpose of the
+  method.
 
 #### Options
 
 An options object may have any of the following properties:
 
 - `contexts` - Set to an array of strings representing the AST context
-  where you wish the rule to be applied (e.g., `ClassDeclaration` for ES6 classes).
-  Overrides the default contexts (see below).
-- `exemptedBy` - Array of tags (e.g., `['type']`) whose presence on the document
-    block avoids the need for a `@description`. Defaults to an empty array.
+  where you wish the rule to be applied (e.g., `ClassDeclaration` for ES6
+  classes). Overrides the default contexts (see below).
+- `exemptedBy` - Array of tags (e.g., `['type']`) whose presence on the
+    document block avoids the need for a `@description`. Defaults to an
+    empty array.
+- `descriptionStyle` - Whether to accept implicit descriptions (`"body"`) or
+    `@description` tags (`"tag"`) as satisfying the rule. Set to `"any"` to
+    accept either style. Defaults to `"body"`.
 
 |||
 |---|---|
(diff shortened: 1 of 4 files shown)


golopot:
In its current form `require-description` is enforcing an unpopular practice -- using description tag in place of intro description, the opposite of what most people is doing. So in my opinion the proposed option looks weird, and it will continue to confuse future users.

I suggest we add no option, and accept both "intro descriptions" and description tag as satisfying the rule, and release it in a minor version.

brettz9:
I sympathize with you, and that is my personal feeling as well, but I think that may be a bit unfair to users who are using the rule to just drop the functionality entirely, since any users using the rule at present (assuming they understand the rule as it currently works) would presumably not want an implicit description to satisfy the rule. @gajus ?

gajus:
If I were to implement this again:

* Rule would have a configuration `descriptionLocation=body|tag` (default=body).
* Rule by default would require implicit description and would complain about explicit description.
* If `descriptionLocation=tag`, then rule would require tag description and complain about implicit description.

I don't see a reason to allow both.

Is there a case when it can be desired to have `@description` and body description?

brettz9:
I think that'd be a good way to go now; it meets the old uses, even if as a breaking change (though I might change it from `descriptionLocation` to `descriptionStyle` unless you changed "body" to "beginning", since otherwise I'm not sure that "body" is necessarily evocative of location).

Per https://jsdoc.app/tags-description.html , a present `@description` will "override" the previous.  I guess someone could use this fact to add introductory annotations in code which don't end up in the docs but I don't personally see much of a use toward doing so.

As far as allowing either to pass the rule, the only reason I can see to allow either form is when one incorporates code from other projects which used a different approach then one doesn't need to refactor to get the benefits of ensuring there is still *some* description present, but forcing a choice of style does have the admitted benefit of always ensuring consistency.

brettz9:
@gajus: I've gone ahead and moved toward this approach, using `descriptionStyle` defaulting to `body` and allowing "tag" or "any". If you don't want "any", let me know and I'll remove.

Issue #314, opened 2019-07-06
v10.3.0-11.2.0 are broken due to missing dependency escape-regex-string


pgoldrbx:
#302 introduced dependency `escape-regex-string` for [parsing `@example` tags](https://github.com/gajus/eslint-plugin-jsdoc/pull/302/files#diff-18f32b9930d05bfb78e6dba553f97dd7R2). However this package was added to `devDependencies` instead of to the package `dependencies`. As a result, all project development and tests are passing but the package fails for downstream consumers.

This was first released in version 10.3.0 and persists still in the current release v11.2.0

```sh
Error: Failed to load plugin 'jsdoc' declared in '--config': Cannot find module 'escape-regex-string'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:582:15)
    at Function.Module._load (internal/modules/cjs/loader.js:508:25)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/Users/pgold/dev/autopilot-server/node_modules/eslint-plugin-jsdoc/dist/rules/checkExamples.js:10:49)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Module.load (internal/modules/cjs/loader.js:600:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
    at Function.Module._load (internal/modules/cjs/loader.js:531:3)
```

brettz9:
Ugh, sorry about that... Too used to Rollup routines pulling it all together. Plan to push out a fix shortly.

PR #316, opened 2019-07-06
fix: move escape-regex-string to dependencies

pgoldrbx:
Fixes https://github.com/gajus/eslint-plugin-jsdoc/issues/314

Moves `escape-regex-string` to dependencies


pgoldrbx:
fixed by https://github.com/gajus/eslint-plugin-jsdoc/commit/ed5693581ed66cc571d016b53f92f8079f035f35

Issue #327, opened 2019-07-11
Tag name autofix confused by adjacent tags


jaydenseric:
As you can see below, if a fixable tag name error is preceded by tags of the same type, the autofix incorrectly targets the first tag, causing multiple autofixes to further and further scramble the first tag name:

<img src="https://user-images.githubusercontent.com/1754873/61031537-107d1200-a403-11e9-88ca-91d0a5936aea.gif" alt="eslint-plugin-jsdoc-fix-bug" width="336" />

brettz9:
@jaydenseric , thanks for the report, and @a-nozeret , thanks for the fix... I've liberalized the regex a little further to allow for a word boundary (`\b` within a `RegExp`) since a tag might not be followed with a space if at the end of a line.

FWIW, I also found and fixed an issue in this rule whereby we needed to treat bad tags as defined so that, if they were not defined elsewhere, they could be found for purposes of reporting and replacement.

Also, @gajus / @golopot , FWIW, I'm working on a stringifier for parsed comment JSON which, if `comment-parser` doesn't accept, we might put in its own repo and add as a dependency, as it should simplify working with fixers. We should be able to use it to make a fix on the parsed comment JSON and then serialize that JSON to a string to get the fixed output. Besides making for easier-to-read fixer code, it should avoid some bugs like in this issue which get introduced because we are trying to reimplement our own parsing.

PR #329, opened 2019-07-11
fix(checkTagNames): #327 - Tag name autofix confused by adjacent tags

a-nozeret:
This fix assumes that all tagNames are followed with a space, would that be correct?

### DIFF
--- a/src/rules/checkTagNames.js
+++ b/src/rules/checkTagNames.js
@@ -57,7 +57,7 @@ export default iterateJsdoc(({
 
       if (preferredTagName !== tagName) {
         report(message, (fixer) => {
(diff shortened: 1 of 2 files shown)


gajus:
Please add a text case showing where it was breaking before.

a-nozeret:
@gajus this fix relates to issue #327 where the bug is detailed. Or do you mean something else missing in the PR?

gajus:
> @gajus this fix relates to issue #327 where the bug is detailed. Or do you mean something else missing in the PR?

If you are fixing bug, then previous test cases were not covering some case. To avoid the bug recurring in the future we should add a test case that shows that the bug is no longer present after this fix is added.

a-nozeret:
Sure! done that @gajus

PR #340, opened 2019-07-15
docs: remove wrong examples in README.md(check-alignment)

yeonjuan:
The two examples you explained in README.md(check-alignment) are wrong I think, you mentioned 'thees are wrong', but in my opinion it is right.
When I run eslint with plugin these two didn't warn.

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,23 +389,7 @@ function quux (foo) {
 // Message: Expected JSDoc block to be aligned.
 
 /**
- * @param {Number} foo
- */
-function quux (foo) {
-
-}
-// Message: Expected JSDoc block to be aligned.
-
-/**
   * @param {Number} foo
- */
-function quux (foo) {
-
-}
-// Message: Expected JSDoc block to be aligned.
-
-/**
- * @param {Number} foo
  */
 function quux (foo) {
 


brettz9:
The examples are built from the test files, and the test files do not have this issue.

The issue is within `src/bin/assertions.js`, specifically the function `trimCode` which trims initial whitespace and uses the last line of code to determine a fixed amount of whitespace to remove from each line besides the first (so as to remove the extra whitespace typically used to offset the code in the test samples).

So if you could file an issue rather than this specific PR (or alter the PR to fix the function), we can track the need for fixing the `trimCode` function.

brettz9:
It seems your second example wasn't fixed by the #342 fix though. Could you take a look at how we might fix that, @yeonjuan ?

yeonjuan:
@brettz9 
The Solution is just change 
```js
  lines = lines.map((line, index) => {
   // delete (index === 0 ) condition and statement
    // if (index === 0) {
      // return line.slice(Math.min(firstIndentSize, lastIndentSize));
    // }
    return line.slice(Math.min(firstIndentSize, lastIndentSize));
  });
```

But In `./test/rules/assertions/` there are some incorrect exemple rules with 'block alignment rule' .
ex) './test/rules/assertions/requireReturns.js' line 174 ~ 176 ( need one space more)

So If you agree, I'll fix it.

PR #342, opened 2019-07-15
fix: Fix trim bug in trimCode

yeonjuan:
Because of this issue(trim bug in trimCode) - #340, I fixed it.


brettz9:
Excellent, thanks!

Issue #323, opened 2019-07-09
Allow empty line after JS Doc in require-jsdoc rule


curiousity4ever:
Currently, the following block of code returns the following exception:

```
/**
 * JS Doc rules!
 */

function ignoreEmptyLineAbove() {
    rule();
}
```

```
 5:1  error  Missing JSDoc comment                                          jsdoc/require-jsdoc
```

Is there a way to make the plugin ignore the empty line after JS Doc / before the statement?

golopot:
It make sense to allow empty line between jsdoc and functions. Jsdoc doc generator allows it, and [projects in the wilds](https://github.com/expressjs/express/blob/4.17.1/lib/express.js) uses it.

brettz9:
Closed by #353.

PR #345, opened 2019-07-18
feat: Allow empty line between jsdoc and functions

yeonjuan:
PR for this issue #323. (allow empty line)
In my humble opinion.. This change seems to be fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5134,6 +5134,12 @@ array.forEach(function() {});
 function MyClass() {}
 
 /**
+  Empty line between jsdoc and functions is allowed
+ */
+ 
+function myFunction() {}
+
+/**
  Function doing something
  */
 function myFunction() {}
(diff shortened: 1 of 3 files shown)


brettz9:
Great to see the responsible source identified. I'd personally like to provide jsdoc settings which can be used to allow a different minimum and/or a maximum line distance, whether for enforcing preferred styling or to avoid introductory comments preceding a function block getting falsely treated as belonging to that function.

To avoid needing to parse ranges with hyphens, etc., I think the simplest implementation would be to have one setting for the max. number of lines and another for the minimum, with a minimum defaulting to 0 and a maximum defaulting to 1 or possibly 2.

In any case, I think we need to support the difference of 0, as there are jsdoc comments like this:

```js
/** @lends MyClass */ { 
    // ...methods
}
```

Any opinions, @gajus or @golopot ?

PR #346, opened 2019-07-20
docs(readme): remove section "Reference to jscs-jsdoc"

golopot:


brettz9:
Perhaps this should go on the wiki instead of being removed entirely? Anyways, I'm fine with it either way, but as this was here before I joined, I think @gajus ought to review this one.

gajus:
I am fine either way. I think that by now most people who would be migrating from ESLint rules would have already made that transition. Happy with this just being removed.

Issue #336, opened 2019-07-13
Missing fixers


brettz9:
The following can have fixers added, with some fixers being by option only:

- `check-examples` (for rules that have fixers)
- `check-param-names` (#47)
- `check-indentation`
- `check-syntax` (add optional param brackets instead)
- `implements-on-classes` - Option to strip tag or change to `@type`
- `match-description` fixers? allow arbitrary 
- `no-types` (removes types)
- `no-undefined-types` (optionally auto-create typedef skeleton);
- `require-description` (fix the missing declaration if opted for "tag" option, i.e., insisting on `@description`)
- `require-description-complete-sentence` (optionally add period at end of previous line).
- `require-example` (fix the missing tag declaration)
- `require-jsdoc` (adds an empty docblock)
- `require-param-name` (add the name/default and whether optional)
- `require-param-type` (optionally add curly brackets so one only need to add the type)
- `require-param` (add the tag only)

yeonjuan:
@brettz9 #396 - Adding fixer for require-param is done! plz check it. :)

brettz9:
@yeonjuan : Would you mind submitting a PR with:

1. An additional test for fixing `bar` in the following:

```js
         /**
           * @param foo
           * @param baz
           */
          function quux (foo, bar, baz) {
          }
```

2. Can we not get all params to be fixed in https://github.com/yeonjuan/eslint-plugin-jsdoc/blob/893916ce6e42c24309d27ac6b5eb71deb6014d5a/test/rules/assertions/requireParam.js#L32-L56 , https://github.com/yeonjuan/eslint-plugin-jsdoc/blob/893916ce6e42c24309d27ac6b5eb71deb6014d5a/test/rules/assertions/requireParam.js#L83-L106 and https://github.com/yeonjuan/eslint-plugin-jsdoc/blob/893916ce6e42c24309d27ac6b5eb71deb6014d5a/test/rules/assertions/requireParam.js#L136-L160

3. I'd also like to know what happens if there is a destructured param like:

```js
          function quux (foo, {bar, baz}) {
          }
```

Lastly, while I wouldn't expect this now, particularly if we add destructuring support (#11), I just thought I'd mention here that it would be nice to have an option to auto-insert some configurable value like `cfg` or `options`:

```js
          /**
           * @param foo
           * @param cfg
           * @param cfg.bar
           * @param cfg.baz
          */
          function quux (foo, {bar, baz}) {
          }
```

[...]

PR #349, opened 2019-07-20
feat: add fixer for no-types

yeonjuan:
Add fixer for no-types(#336) : remove types

### DIFF
--- a/README.md
+++ b/README.md
@@ -3311,6 +3311,14 @@ function quux () {
 
 }
 // Message: Types are not permitted on @returns.
+
+/**
+ * @param {number} foo {number}
+ */
+function quux (foo) {
+
+}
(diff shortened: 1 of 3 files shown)


brettz9:
With `comment-parser` now including a `stringify` method, we should be able to make for a cleaner and less error prone fixing process by modifying `jsdoc.tags` and stringifying the result rather than attempting to parse and replace with strings and regular expressions. The latest in `master` (from commit 0503b4ac9c14b578fe27465b622ad61689a06bf6 ) adds a new method `utils.reportJSDoc` which can be used to report with a fixer that stringifies `jsdocNode`.

It would be great if you could rebase on `master` and see about reimplementing with that approach.

If there are any limitations in the comment-parser `stringify` or or in our API, let us know, so we can try to meet the need.

Issue #354, opened 2019-07-23
Check  `yield` and `@yields`/`@yield`


brettz9:
While #271 aims to support `yield` with `require-returns-check` (in environments like TypeScript where there is no `@yields` tag support and thus it makes some sense to use `@returns` to contain this information), I think a new rule ought to be added for regular jsdoc mode (i.e., not TypeScript or GCC which don't support `@yields`/`@yield`) to ensure a `yield` is present when the `@yields`/`@yield` tag is used. 

Similarly, I think we need another rule equivalent to `require-returns` as well, which checks that if a `yield` is present, that there is a `@yields`/`@yield` tag.

~And this rule should allow `IterableIterator<void>` and `IterableIterator<SomeType>` within `@returns` as well (for TypeScript/GCC which don't support `@yields` currently).~ *`@returns` might perhaps be reserved for the generator return values or if using the actual Generator/iterator such as `IterableIterator` with a generic/template, then use any `return` value as the variable portion rather than the `yield` value.*

golopot:
This will be a seldom used rule. I will accept it if enough people shows interest in it. Otherwise it is better to keep the project small and simple.

brettz9:
People have already expressed interest by filing an issue, and I would like the same. This is `eslint-plugin-jsdoc` not `eslint-plugin-jsdoc-small-subset-of-tags`. Yes, we don't want bloat, but there is no reason we can't meet multiple needs and users can ignore rules they don't want. What is so oppressive about having an extra rule that you don't personally use? We are presumably not just here to meet your or my needs.

PR #355, opened 2019-07-23
Allow `yieldAsReturn` option in conjunction with new `settings.jsdoc.mode`

### LINKED ISSUE #271: require-returns-check doesn't work with generator functions
When trying to write a generator function (`*myfunction`) `require-returns-check` complains that no return statement is found even though the function contains the `yield` keyword (which essentially works as a return.

Example code (using Typescript): 

```ts
    /**
     * Generator for prizes:
     * @returns {IterableIterator<number>}
     * @memberof GameController
     */
    public *prizes(): IterableIterator<number> {
        for (const prizeValue of this.prizes) {
            yield prizeValue;
        }
    }
```

I searched through existing issues and found that there were some issues with the checker and implicit returns before. I believe this would be a similar kind of issue, if someone gives me some tips I could try and PR a fix for it.

brettz9:
feat(require-returns-check): with `settings.jsdoc.mode` (`typescript` or `gcc`), allow `yieldAsReturn` option as "always" or "argument" to treat `yield` as `return` (fixes #271)

The `mode` setting--as discussed in another issues--would allow precise tweaking of behavior across other rules as well (e.g., for `check-tag-names` to disallow `@template` by default in regular jsdoc mode, but allow it with `gcc`). But this PR only ensures that regular jsdoc mode cannot treat yield as a return since regular jsdoc has the `@yields` tag which may be used for that purpose.

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -83,6 +83,12 @@ You can then selectively add to or override the recommended rules.
 
 ## Settings
 
+### Mode
+
+While TypeScript and the Google Closure Compiler (GCC) support some jsdoc tags,
+there are some differences. The `settings.jsdoc.mode` lets you choose the precise
+mode (`"typescript"`, `"gcc"`, or `"jsdoc"` (the default)).
+
 ### Allow `@private` to disable rules for that comment block
 
 - `settings.jsdoc.ignorePrivate` - Disables all rules for the comment block
--- a/.README/rules/require-returns-check.md
+++ b/.README/rules/require-returns-check.md
@@ -4,6 +4,19 @@ Requires a return statement in function body if a `@returns` tag is specified in
 
 Will also report if multiple `@returns` tags are present.
 
+#### Options
+
+##### `yieldAsReturn`
+
+If the setting `settings.jsdoc.mode` is set to `typescript` or `gcc`, this option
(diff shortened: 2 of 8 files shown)


golopot:
I think a better approach is to skip `require-returns-check` rule for all generator functions. A generator function always returns an iterator, regardless of the presence of yield statement. This approach is analogous to how we treat async functions in the rule.

brettz9:
For jsdoc mode, I was in agreement--and I would hope TypeScript/GCC will come to support `@yields` to differentiate. My thinking was to allow for TypeScript/GCC just for the time-being, overloading `@returns`.

However, in looking at https://github.com/microsoft/TypeScript/issues/23857 , I see that they already have a syntax for overloading `@returns`, indicating (similar to how Promises can be documented) the type yielded by the iterator, `IterableIterator<Something>` in the issue example. The issue mentions eslint `valid-jsdoc` as complaining, but that, as we know, is deprecated, and we can of course, get it to work ourselves.

So rather than skipping checking for generators, I think we should make sure that unless it is `IterableIterator<void>` or such (in which case just a `yield` might be expected), we should require `yield` with a value. It shouldn't be hard to adapt this PR to do this, and it'd avoid the need for the option.

golopot:
That feature should be addressed in #354.

PR #387, opened 2019-09-27
`require-jsdoc`: Highlight first line only of missing jsdoc

### LINKED ISSUE #386: Atom require-jsdoc
I'm not sure if this should be handled by this plugin, but when running Atom (with linter-eslint), the jsdoc/require-jsdoc rule causes the entire function to be highlighted.
When running intellij, it just highlights the first line of the function.
I would prefer that it not highlight the entire function.

brettz9:
fix: avoid highlighting whole function block upon missing jsdoc block in favor of highlighting first line only; fixes #386

I'm calling it a "fix" as it is not really a new feature, and more of fixing an annoyance, though it's kind of change in behavior a little, so requesting your review, @golopot .

### DIFF
--- a/src/rules/requireJsdoc.js
+++ b/src/rules/requireJsdoc.js
@@ -172,6 +172,19 @@ export default {
         return fixer.insertTextBefore(baseNode, insertion);
       };
 
+      const report = () => {
+        const loc = {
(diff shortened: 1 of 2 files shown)


golopot:
Thanks!

Issue #211, opened 2019-05-08
`check-examples` and whitespace-based rules (like `indent`)


brettz9:
As currently implemented, `check-examples` has issues when the rule is `indent` as initial spacing is stripped out during parsing. May need to tweak parsing or use `jsdocNode`.

slieker-d:
Can confirm. Having the same issue.

brettz9:
Be aware that in the meantime you can use `overrides` to disable only for examples--using the rule in conjunction with `settings.jsdoc.matchingFileName`, e.g., `{settings: {jsdoc: {matchingFileName: "dummy.md"}}}` (here to get the same rules one has for Markdown files--and possible use with `eslint-plugin-markdown`, but you could match against any file or file type) and then setting your `.eslintrc.*`'s `overrrides` to match that file or extension, e.g.,

```js
overrides: [
  {
    files: ["**/*.md"],
    rules: {
      indent: 'off'
    }
  }
]
```

slieker-d:
I added the overrides section to our `.eslintrc.json` file, but we're still getting this:

<img width="765" alt="image" src="https://user-images.githubusercontent.com/30084394/58302387-131a9e00-7da8-11e9-9318-b8c36b1279ca.png">

slieker-d:
This is our entire eslintrc file:

```json
{
    "env": {
        "es6": true,
        "node": true
    },
    "extends": ["airbnb-base", "plugin:jest/recommended"],
    "globals": {
        "Atomics": "readonly",
        "SharedArrayBuffer": "readonly",
        "jest": "readonly"
    },
    "parserOptions": {
        "ecmaVersion": 2018,
        "sourceType": "module"
    },
    "plugins": [
        "jest",
        "jsdoc"
    ],
    "rules": {
        "jest/no-disabled-tests": "warn",
        "jest/no-focused-tests": "error",
        "jest/no-identical-title": "error",
        "jest/prefer-to-have-length": "warn",
        "jest/valid-expect": "error",

        "jsdoc/check-alignment": 1,
        "jsdoc/check-examples": 1,
        "jsdoc/check-indentation": 0,
        "jsdoc/check-param-names": 1,
        "jsdoc/check-syntax": 1,
        "jsdoc/check-tag-names": 1,
        "jsdoc/check-types": 1,
        "jsdoc/implements-on-classes": 1,
        "jsdoc/match-description": 1,
        "jsdoc/newline-after-description": 1,
[...]

PR #388, opened 2019-09-30
Fix: ignore examples in `check-indentation`

### LINKED ISSUE #334: Don't check indentation within code blocks
Currently, if `jsdoc/check-indentation` is set and a markdown code block is included that contains indentation, the comment is considered to be an error because there is indentation in the comment. However, code blocks are expected to be properly indented for display, so this rule should either not check within blocks, or it should have an option to ignore code blocks.

Example:
```javascript
/**
 * This is a class that injects some HTML elements.
 * The injected elements can be styled as follows:
 * ```xml
 * <container>
 *  <element class="readout__gauge gauge">
 *    <element class="gauge__name">{Caption}</element>
 *    <element class="gauge__value">{Value}</element>
 *  </element>
 *  <element class="readout__gauge gauge">
 *    <element class="gauge__name">{Caption}</element>
 *    <element class="gauge__value">{Value}</element>
 *  </element>
 * </container>
 * ```
 */
```

ahwayakchih:
This should fix ~#334~ by making `check-indentation` rule ignore whole `@example` blocks.

UPDATE: actually it does not fix #334, but it fixes similar issue with example code blocks.

It could be improved to allow to pass which tags should be ignored as an option, instead of rather specific `excludeExamples` i added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -835,10 +835,36 @@ function quux () {}
 
 Reports invalid padding inside JSDoc block.
 
+<a name="eslint-plugin-jsdoc-rules-check-indentation-options-1"></a>
+#### Options
+
+This rule has an object option.
+
+<a name="eslint-plugin-jsdoc-rules-check-indentation-options-1-excludeexamples"></a>
+##### <code>excludeExamples</code>
+
+This boolean property allows to "hide" example code from reports.
+
+By default, whole JSDoc block is checked for invalid padding.
+That includes example blocks too, which may get in the way of adding full,
+readable examples of code without ending up with multiple linting issues.
+
+When enabled, following code will lint without any padding issue:
+
+```js
+/**
+ * @example
+ * anArray.filter((a) => {
+ *   return a.b;
+ * });
+ */
+```
+
 |||
 |---|---|
 |Context|everywhere|
 |Tags|N/A|
+|Options| `excludeExamples` |
 
 The following patterns are considered problems:
 
@@ -871,14 +897,13 @@ class Moo {}
  * foo
  *
  * @example
- * function xFoo () {
- *   return 'foo';
- * }
+ * anArray.filter((a) => {
+ *   return a.b;
+ * });
  */
 function quux () {
 
 }
-// Options: ["excludeExamples":false}]
 // Message: There must be no indentation.
 ````
 
@@ -904,14 +929,14 @@ function quux () {
  * foo
  *
  * @example
- * function xFoo () {
- *   return 'foo';
- * }
+ * anArray.filter((a) => {
+ *   return a.b;
+ * });
  */
 function quux () {
 
 }
-// Options: ["excludeExamples":true}]
+// Options: [{"excludeExamples":true}]
 ````
 
 
@@ -1265,10 +1290,10 @@ yields
 
 Note that the tags indicated as replacements in `settings.jsdoc.tagNamePreference` will automatically be considered as valid.
 
-<a name="eslint-plugin-jsdoc-rules-check-tag-names-options-1"></a>
+<a name="eslint-plugin-jsdoc-rules-check-tag-names-options-2"></a>
 #### Options
 
-<a name="eslint-plugin-jsdoc-rules-check-tag-names-options-1-definedtags"></a>
+<a name="eslint-plugin-jsdoc-rules-check-tag-names-options-2-definedtags"></a>
 ##### <code>definedTags</code>
 
 Use an array of `definedTags` strings to configure additional, allowed JSDoc tags.
@@ -1610,7 +1635,7 @@ Date
 RegExp
 ```
 
-<a name="eslint-plugin-jsdoc-rules-check-types-options-2"></a>
+<a name="eslint-plugin-jsdoc-rules-check-types-options-3"></a>
 #### Options
 
 `check-types` allows one option:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -835,10 +835,36 @@ function quux () {}
 
 Reports invalid padding inside JSDoc block.
 
+<a name="eslint-plugin-jsdoc-rules-check-indentation-options-1"></a>
+#### Options
+
+This rule has an object option.
+
+<a name="eslint-plugin-jsdoc-rules-check-indentation-options-1-excludeexamples"></a>
+##### <code>excludeExamples</code>
```
golopot: I think this fix does not need to be under an option. As there is no case where `check-indent` in `@example` block is welcomed.

test/rules/assertions/checkIndentation.js
```diff
@@ -48,6 +48,27 @@ export default {
...
+
+          }
+      `,
+      errors: [
+        {
+          line: 7,
+          message: 'There must be no indentation.',
+        },
+      ],
+    },
```
golopot: Can you add a test to make sure blocks after example block are still checked?
```js
/**
 * @example 
 * @returns
 *     eeee
 */
```

README.md
brettz9: There could be some rare exceptions when there is explanatory text or if one wanted the formally specified `@example` `<caption>` inside tag to be indented, but I'd tend to agree it'd be unusual and even then, it wouldn't work with the sample code following (I guess one could check the `<caption>` is indented though, as we know that to be regular text).

Also, FWIW, I've reopened #211 as the intent there is not merely to disable the ESLint `indent` rule for JavaScript within `@example` but to parse indenting correctly in our rule so that `indent` can be applied.

test/rules/assertions/checkIndentation.js
ahwayakchih: I tested it locally, but yeah, it's a good idea to keep it.

ahwayakchih:
@brettz9, @golopot i could also add an option to ignore "code blocks" in description, which would fix #334, but i'm not sure if you'd be willing to accept such change, and if it's ok for this PR or better to add it separately.

golopot:
Another pr to handling code blocks (three ticks) would be great and it does not need to be under an option. It looks like that three ticks is recognized by jsdoc doc generator through the [markdown plugin](https://jsdoc.app/plugins-markdown.html#overview). And VS code recognized it.

ahwayakchih:
> Another pr to handling code blocks (three ticks) would be great and it does not need to be under an option.

OK, i prepared #391, but it has to wait because it builds on top of this PR (so i don't have to solve merge conflicts later).

README.md
golopot: I still do not find a use case for the option `excludeTags`. Adding an option is an expensive thing so unless there is a solid use case it shouldn't be added.

The `@example <caption>...` use case confuses me. Isn't it that you need `ignoreExample: true` to properly indent that case?

ahwayakchih: @golopot There's also `@description` that allows "freestyle" text. There are also `@classdesc` and `@summary`, but i don't know if they allow multiple lines of text. Who know what else will be added in future?
I can see now that one could want to enforce simplest, almost one-liner-like examples in JSDoc, with additional `@see` links to full-blown example code in separate, "executable" files. In which case running `check-indentation` on them could help.

As for `@example <caption>...</caption>` case: i don't know, i never used captions before. I find its syntax... weird/out-of-place compared to the rest of the JSDoc syntax ;).

golopot: Okay, that's fair.

ahwayakchih: Thanks :).

src/rules/checkIndentation.js
```diff
@@ -17,6 +31,18 @@ export default iterateJsdoc(({
 }, {
   iterateAllJsdocs: true,
   meta: {
+    schema: [{
+      additionalProperties: false,
+      properties: {
+        excludeTags: {
+          items: {
+            type: 'string',
```
golopot: @ahwayakchih 
I forgot this thing, can you add this in a follow up pr? Validating string passed to regexp looks like a good idea.   
```suggestion
            type: 'string',
            pattern: '^[a-zA-Z]+$',
```

ahwayakchih: @golopot no problem. I added it to #391. To keep things super clean and tidy, it should be a separate PR, but the change is so tiny, i assumed it will be ok. Let me know if you want it separated.

README.md
brettz9: Sorry for the delayed reply... The idea with `@example <caption>` is merely that you could conceivably want indenting for this:

```js
/**
* @example <caption>
* Here is a long
* summary of this
* example
* </caption>
* ```js
* function hi () {
*   alert('Hello');
* }
* ```
*/
```

You might want `check-indentation` to apply to the caption contents (though you'd want the js block to instead be linted by normal `indent` rule as per #211).

But I don't personally care enough about this use case to think it warrants a fix/issue unless someone else did.

ahwayakchih: @brettz9 thanks for the explanation. The way you explained it, it should work with the changes i made :). Just not by default, because example tag is excluded then. But if you disable excluding it, it will be checked and code will not because it's wrapped into fenced code block.
I can add that to the test assertions:
```js
    {
      code: `
          /**
          * @example <caption>
          * Here is a long
          * summary of this
          * example
          * </caption>
          * \`\`\`js
          * function hi () {
          *   alert('Hello');
          * }
          * \`\`\`
          */
      `,
      options: [{
        excludeTags: [],
      }],
    },
```
Now, if you would add some indentation inside the caption lines, it would error. Same if JS code would not be wrapped into fenced code block.

brettz9: Such an assertion sounds good, thanks!

ahwayakchih: @brettz9 done: #392

PR #391, opened 2019-10-02
Fix #334: ignore code blocks in check-indentation

### LINKED ISSUE #334: Don't check indentation within code blocks

ahwayakchih:
This should fix #334.

~It builds on top of #388, to prevent merge conflicts later, if both are to be merged.~ #388 was merged, this one was rebased on latest master.

### DIFF
--- a/.README/rules/check-indentation.md
+++ b/.README/rules/check-indentation.md
@@ -2,6 +2,20 @@
 
 Reports invalid padding inside JSDoc block.
 
+Ignores parts enclosed in Markdown's "code block". For example,
+following description is valid:
+
+```js
+/**
+ * Some description:
+ * ```html
+ * <section>
+ *   <title>test</title>
+ * </section>
+ * ```
+ */
+```
+
 #### Options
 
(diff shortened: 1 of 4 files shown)


ahwayakchih:
Thanks :).

Issue #356, opened 2019-07-23
JSDoc/GCC/TypeScript mode


brettz9:
#354 proposes adding a `mode` setting for allowing different behavior between TypeScript, Closure, and regular jsdoc. This setting could be used for other tweaks:

- [x] Avoid `parseClosureTemplateTag` in `no-undefined-types` except for `closure`
- [x] Toggle `check-syntax` behavior (report non-GCC when in closure mode)
- [x] Start checking `module` for namepaths (see "todo" in `jsdocUtils.js` source), but only for "jsdoc" (and "closure"?) mode (TS doesn't support)
- [x] `valid-types`
    - [x] `valid-types` - Ensure tags with optional type (`tagsWithOptionalTypePosition`) only allow for `package`, `private`, `protected`, `public`, and `static` if in closure mode.
    - [x] `valid-types` - Only allow `@extends {SomeType}` with brackets if in closure mode.
    - [x] `valid-types` - Only allow `@typedef {SomeType}` (without a name) in closure mode.
    - [x] Report name on [`@interface`](https://jsdoc.app/tags-interface.html) in Closure
- [x] Don't check optional name on [`@interface`](https://jsdoc.app/tags-interface.html) in Closure
- [x] For `check-tag-names`, avoid allowing `@template` except in `closure` and `typescript`
- [x] For `check-tag-names`, add other GCC tags for `closure` mode (#306)
- [x] `check-tag-names` to prefer other defaults with `tagNamePreference` (`return` over `returns`)
- [x] See about getting jsdoctypeparser (e.g., see https://github.com/jsdoctypeparser/jsdoctypeparser/issues/50 on TypeScript support) to support different modes so long as jsdoc/TypeScript/Closure differ in the formats supported (might have comments which are used to build alternate versions of the peg file, e.g., stripping out those for TypeScript or Closure only). One explicitly desired rule to be avoided for closure per #495 is to disallow tuples. Might also have a tolerant version (the default?) which accepted all forms though (?).

Might accept a version in an array along with the mode string in case jsdoc, Closure, or TypeScript make breaking changes or add more features.

l1bbcsg:
`check-syntax` is another important rule to tweak. Currently it tweaks agains GCC style, but in gcc mode it could have the opposite behavior.

`check-types ` rule, `preferredTypes` and `tagNamePreference` settings could use different defaults in GCC mode, i.e. prefer `return` over `returns`.

brettz9:
[...]

PR #400, opened 2019-10-14
Begin `settings.jsdoc.mode` (`check-tag-names` and others indirectly)

brettz9:
feat(`check-tag-names` and other rules): begin `settings.jsdoc.mode`; utilize in `getPreferredTagName`, `isValidTag`, `getTagsByType`; fixes part of #356

BREAKING CHANGE: Allowable values: jsdoc|typescript|closure

I don't like the relatively large list of tags added to the README, but I also don't like not testing them so not sure if there is a better approach.


gajus:
LGTM

PR #404, opened 2019-10-17
feat(`require-param`): change fixer to fix all missing tags at once

yeonjuan:
* @brettz9 - [ This is PR for 1, 2](https://github.com/gajus/eslint-plugin-jsdoc/issues/336#issuecomment-541258157)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6377,6 +6377,27 @@ function quux (foo) {
 }
 // Settings: {"jsdoc":{"tagNamePreference":{"param":false}}}
 // Message: Unexpected tag `@param`
+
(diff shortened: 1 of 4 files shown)


brettz9:
Although ESLint runs fixes in multiple passes and so this may technically seem unnecessary, I think there is still an advantage to such changes as these since it makes the documentation more clear when we indicate the expected output as having fixed all items.

Re: my destructuring concern, I pushed an additional commit to get satisfaction on that (it is safely ignored for now no matter the position).

Thanks very much! Great work as usual!

Issue #368, opened 2019-08-22
 Expected JSDoc block to be aligned.  jsdoc/check-alignment


OmgImAlexis:
The first returns a warning. The second is what was returned from `--fix`. The example I'm using is from the examples on [jsdoc/check-alignment](https://github.com/gajus/eslint-plugin-jsdoc#eslint-plugin-jsdoc-rules-check-alignment
).

```
  ⚠  8:0  Expected JSDoc block to be aligned.  jsdoc/check-alignment

  1 warning
```

```js
module.exports = function (si) {
	/**
      * Desc
      *
      * @param {number} foo
      */
     function quux(foo) {
     }
};

```
```js
module.exports = function(si) {
	/**
  * Desc
  *
  * @param {number} foo
  */
	function quux(foo) {
	}
};
```

golopot:
Looks like the rule does not handle tab indents properly.

brettz9:
Fixed by #412

Note, however, that the fix is based on the assumption that the initial line is the one with the correct indentation. So in your case, the fixed code will look like:

```js
module.exports = function (si) {
	/**
	 * Desc
	 *
	 * @param {number} foo
	 */
     function quux(foo) {
     }
};
```

If you wanted to see the indent instead made relative to the function or other context it is preceding, a new issue could be filed.

OmgImAlexis:
@brettz9 a new issue would be a great idea. Would you mind opening it?

brettz9:
I have filed #413 . In the future, please file yourself so you can be automatically subscribed to the issues and can state the concern as you see it.

To subscribe to that issue without needing to add any comments, you can click Github's "Subscribe" button on the page for #413 on the right-hand side of the page.

Thanks for the report!

OmgImAlexis:
You don’t need to treat me like an idiot. I know how to use Github. I was on my phone when I saw the notification that’s all. 💁‍♀️

brettz9:
I didn't know about the Subscribe button last year or so despite being on Github for years! But then again, I may be idiot, so I wouldn't know how to treat others otherwise. :)

brettz9:
And FWIW, as a former teacher, I tend to overexplain things to *everyone* in the event not only that the immediate listener doesn't happen to know but also in case others coming across the issue don't either, so even if I do state something seemingly obvious, please don't take it as being directed at you.

646767173:
> 

you don't need to apologize, you are doing a right thing, my friend.

Issue #369, opened 2019-08-22
With TypeScript CTRL+S adds bad comments to a function inside a class.


jasson2788:
Hey guys!

Inside a class and pressing CTRL+S will add to this code.
```
class Dog {
	eat() {
		
	}
}
```

A comment like that: 
```
class Dog {
    eat/**
 *
 */
() {
		
    }
}
```
Expected result:
```
class Dog {
    /**
     * 
     */
    eat() {
		
    }
}
```

P.S. I am using TypeScript.

Thank you guys.

jasson2788:
Hello, @golopot.

I hope you are doing well. Any updates on that issue? The issue still happens when autofix is on.

golopot:
Pull requests are welcomed.

Issue #199, opened 2019-04-12
Feature request: rule to enforce asterisks


rasenplanscher:
It would be nice to have a rule that enforces every line in a JSdoc comment to start with an asterisk, as in `https://palantir.github.io/tslint/rules/jsdoc-format/`

golopot:
Comparison between `tslint/jsdoc-format` and `eslint-plugin-jsdoc`: 
- [ ] each line contains an asterisk
- [x] asterisks must be aligned - jsdoc/check-alignment
- [x] disallow extra space after asterisks - jsdoc/check-indentation
- [ ] require one space after asterisks
- [ ] multiline comments don’t allow text after /** in the first line (under option)

Both unsupported:
- [ ] bare block comment: disallow asterisks

skeggse:
I submitted a PR for this at #446. Not amazingly happy about the name, though - thoughts?

brettz9:
Other checks I can think of on this general topic might be:

1. Avoiding multiple asterisks (beyond the obligatory first two), whether at the beginning, end, or middle
2. Avoiding single line jsdoc blocks such as `/** @param abc */`, perhaps excepting some tags like [`@lends`](https://jsdoc.app/tags-lends.html) which, um, "lend" themselves to single line expressions because of their terseness.

brettz9:
Regarding reporting jsdoc blocks with multiple asterisks at the beginning, I've just pushed a fix at 28397c7ed54b34a320a9d3961131da2dbf87c534 so that `no-bad-blocks` can also check 3+ asterisks when what appears to be a tag is found therein. That case is being handled separately because multiple asterisks at the beginning are not actually jsdoc (so we use that rule to catch comments which *look like* jsdoc).

I've also filed #728 so as to allow preventing `/*** ... */` in that rule even where the contents appear to have no tag (normally not prevented since it could just be a multiline comment with a lot of asterisks).

brettz9:
After the current PRs, I think the one remaining item here would belong in a rule like `multiline-comments` with options;

1. No line 0 text if multiline
2. Avoid single line entirely, unless optionally possessing specific tags like `@lends`
3. Require single lines only, optionally if under a certain length and/or if tags and/or particular tags are present

brettz9:
> After the current PRs, I think the one remaining item here would belong in a rule like `multiline-comments` with options;
> 
>     1. No line 0 text if multiline
> 
[...]

PR #412, opened 2019-10-28
fix: Alignment fixer should use indentation character from sourceCode.

kaidjohnson:
addresses #368 

There were a few other areas where `' '.repeat` is used that probably need to be looked into further, as I believe they apply to rules other than "check-alignment".

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,7 +338,31 @@ The following patterns are considered problems:
   * @param {Number} foo
  */
 function quux (foo) {
+  // with spaces
+}
+// Message: Expected JSDoc block to be aligned.
 
+/**
+  * @param {Number} foo
+ */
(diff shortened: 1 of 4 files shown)


gajus:
Please add a test case.

kaidjohnson:
Uncovered some inconsistency with how each line trimmed before evaluation while working through the test. Let me know if there's anything else with this update that should be addressed.

brettz9:
Nice work! I've added a couple test cases to ensure we're ok flush at the beginning of a line, and those tests pass as well.

If you may be so inspired to help with any other indentation-impinging issues (e.g., #369, #199 , #211, or the `check-indentation` fixer (to strip indents) in #336), it would be most appreciated as well!

PR #417, opened 2019-11-03
feat(`getJSDocComment`): ignore line comment

whitneyit:
Often it is required to disable linting in-between the JSDoc and the
code. This PR allows that to happen.

Without the following change the `jsdoc/require-jsdoc` rule needs to be
constantly disabled.

Consider the following example:

```ts
/**
 * A user.
 * @interface User
 */
interface User {
  /**
   * The users age.
   * @type {number}
   * @memberof User
   */
  age: number;

  /**
   * The users name.
   * @type {string}
   * @memberof User
   */
  name: string;
}

/**
 * Determine if the provided value is a `User`;
 * @param {*} user The value to check.
 * @returns {boolean} Whether or not the provided value is a `User`.
 */
// Allow the Function to check any input.
// tslint:disable-next-line: no-any
function isUser(user: any): user is User {
  if (user == null) {
      return false;
  }
  const { age, name } = user;
  if (typeof age !== `number`) {
      return false;
  }
  if (typeof name !== `string`) {
      return false;
  }
  return true;
}
```

In the above example, using `unknown` wouldn't cut it. This is because
`unknown` cannot be destructured. Eg: `const { age, name } = user;`.

This PR also updates the checkExamples.js assertion so as not to require
an absolute path to @typescript-eslint/parser.

This absolute path would mean constant calls to npm run create-readme
because of the different paths that developers would have the repo
checked out at.

### DIFF
--- a/README.md
+++ b/README.md
@@ -790,7 +790,7 @@ function f () {
 function quux () {
 
 }
-// Options: [{"baseConfig":{"parser":"/Users/brett/eslint-plugin-jsdoc/node_modules/@typescript-eslint/parser/dist/parser.js","parserOptions":{"ecmaVersion":6},"rules":{"semi":["error","always"]}},"eslintrcForExamples":false}]
+// Options: [{"baseConfig":{"parser":"@typescript-eslint/parser","parserOptions":{"ecmaVersion":6},"rules":{"semi":["error","always"]}},"eslintrcForExamples":false}]
 // Message: @example error (semi): Missing semicolon.
 ````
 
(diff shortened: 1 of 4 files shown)


whitneyit:
This PR also updates the `checkExamples.js` assertion so as not to require an absolute path to `@typescript-eslint/parser`.

This absolute path would mean constant calls to `npm run create-readme` because of the different paths that developers would have the repo checked out at.

brettz9:
This looks like a very useful addition. I could have sworn that absolute paths became required with eslint@6, but it seems CI is passing.

whitneyit:
@brettz9 this would be very helpful in reduces the number of times that the `jsdoc/require-jsdoc` has to be disabled.

Funnily enough, I could have sworn that this worked before. I checked some older code and it ran fine. I'm not sure what has changed.

brettz9:
Well it is passing checks now, so maybe eslint stopped requiring absolute paths...

brettz9:
I figured out that it is when passing strings to the `parser` property in the RuleTester that one gets the message "Parsers provided as strings to RuleTester must be absolute paths". Within the options, however, it is ok.

PR #419, opened 2019-11-05
fix(`no-undefined-types`): do not crash on variadic arguments

TimvdLippe:
When running `no-undefined-types` on our codebase, I discovered that it
crashes on `variadic` arguments. It turns out that jsdoctypeparser does
not cleanly traverse over variadic arguments. Thus, used the
`parsedType` to traverse over instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4090,6 +4090,13 @@ class Foo {
   }
 }
 // Message: The type 'TEMPLATE_TYPE_A' is undefined.
+
+/**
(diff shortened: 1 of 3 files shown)


brettz9:
I've submitted https://github.com/jsdoctypeparser/jsdoctypeparser/pull/100 as I think it makes sense to fix over there. Thank you for identifying the issue and providing tests... I'll see about incorporating the tests after jsdoctypeparser may accept the changes.

PR #426, opened 2019-11-09
feat(`check-examples`): support global regexes and other flags besides now default "u" (i.e., any of `gimys`)

### LINKED ISSUE #331: check-examples exampleCodeRegex option with global and multiline expression flags
It would be great if the `check-examples` rule `exampleCodeRegex` option could work with global and multiline expression flags, so that multiple markdown code blocks could be linted per example. Here is a contrived example, [supported by `jsdoc-md`](https://github.com/jaydenseric/jsdoc-md#example-content):

```js
/**
 * @example <caption>Say `Hello!` to the user.</caption>
 * First, import the function:
 * 
 * ```js
 * import popup from './popup'
 * ```
 *
 * Then use it like this:
 *
 * ```js
 * popup('Hello!')
 * ```
 *
 * Here is the result on macOS:
 *
 * ![Screenshot](path/to/screenshot.jpg)
 */
```

Here is an example ` /^```(?:js|javascript)\n([\s\S]*?)```$/gm ` pattern for the `exampleCodeRegex` option:

https://regexr.com/4h9sh

Each regex result should be linted in isolation; I tried quickly adding `gm` flags to the source code here:

https://github.com/gajus/eslint-plugin-jsdoc/blob/f0c246cc583ed9d79869d2556023453d47807f67/src/rules/checkExamples.js#L70

But that caused weirdness where ESLint would not allow the same named `const` in two code blocks, etc. and the errors for following code blocks were reporting in the wrong places.

brettz9:
Fixes #331.

### DIFF
--- a/.README/rules/check-examples.md
+++ b/.README/rules/check-examples.md
@@ -27,11 +27,16 @@ syntax highlighting). The following options determine whether a given
   first such group treated as one to include. If no parenthetical group
   exists or matches, the whole matching expression will be used.
   An example might be ````"^```(?:js|javascript)([\\s\\S]*)```\s*$"````
-  to only match explicitly fenced JavaScript blocks.
+  to only match explicitly fenced JavaScript blocks. Defaults to only
+  using the `u` flag, so to add your own flags, encapsulate your
+  expression as a string, but like a literal, e.g., ````/```js.*```/gi````.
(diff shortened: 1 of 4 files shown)


brettz9:
Just want to review a little bit further and see about possibly adding another test or so.

PR #429, opened 2019-11-13
`check-values`: add `licensePattern` option

brettz9:
feat(`check-values`): add `licensePattern` option to allow delimiting portion of license description to extract

Thereby can allow extra descriptive text before or after a license pattern.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3156,11 +3156,20 @@ be checked for.
 An array of allowable license values or `true` to allow any license text.
 If present as an array, will be used in place of SPDX identifiers.
 
+<a name="eslint-plugin-jsdoc-rules-check-values-options-5-licensepattern"></a>
+##### <code>licensePattern</code>
+
+A string to be converted into a `RegExp` (with `u` flag) and whose first
+parenthetical grouping, if present, will match the portion of the license
+description to check (if no grouping is present, then the whole portion
+matched will be used). Defaults to `([^\n]*)`, i.e., the SPDX expression
+is expected before any line breaks.
+
 |||
 |---|---|
 |Context|everywhere|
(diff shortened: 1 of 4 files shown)


test/rules/assertions/checkValues.js
```diff
@@ -201,6 +217,17 @@ export default {
...
+      code: `
+      /**
+       * @license MIT
+       * Some extra text...
+       */
+      function quux (foo) {
+
+      }
+      `,
+    },
```
golopot: Tests for the added option is missing

README.md
```diff
@@ -3156,11 +3156,20 @@ be checked for.
 An array of allowable license values or `true` to allow any license text.
 If present as an array, will be used in place of SPDX identifiers.
 
+<a name="eslint-plugin-jsdoc-rules-check-values-options-5-licensepattern"></a>
+##### <code>licensePattern</code>
```
golopot: Can the name for this option be more clear that it deals with the license description instead of license name?

test/rules/assertions/checkValues.js
brettz9: Ah, good find, ok, plan to add shortly.

README.md
brettz9: It actually does deal with the license name though. It finds the license name out of the whole tag description.

brettz9:
I've added some tests. Let me know about the option name, and if you think it is not clear, if you could offer an alternative. But its function is to extract the license name, even though its effect is also to indirectly allow license text before or after that name (since it only checks the name for validity).

brettz9:
@golopot : I'll go ahead and merge then if I don't hear anything back from you in the next day or so.

PR #438, opened 2019-11-22
feat(`require-description-complete-sentence`): tighten constraint

whitneyit:
Currently it is possible to experience a false positive with the
`jsdoc/require-description-complete-sentence` rule.

The false positive is caused by starting a new line with a capital
letter when the previous line does not end with a period. Normally, this
is the correct behaviour, however, English sentences are more nuanced
than that.

The problem arises when using a word that starts with a capital letter
that just so happens to be at the start of a line. For instance, `I`, or
a programmatic word like `XMLHttpRequest`.

This PR tightens the constraint of the `isNewLinePrecededByAPeriod`
Function by checking for a lowercase letter immediately after an
uppercase letter.

This will still produce false positives if the new line starts with
`UserInterface` for example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5376,6 +5376,22 @@ function quux (foo) {
  * |   1  | do it       |
  * |   2  | do it again |
  */
+
+/**
+ * This is something that
+ * I want to test.
+ */
+function quux () {
+
+}
+
+/**
+ * When making HTTP requests, the
+ * URL is super important.
+ */
+function quux () {
+
+}
 ````
 
 
--- a/src/rules/requireDescriptionCompleteSentence.js
+++ b/src/rules/requireDescriptionCompleteSentence.js
@@ -45,7 +45,7 @@ const isNewLinePrecededByAPeriod = (text) => {
   const lines = text.split('\n');
 
   return !lines.some((line) => {
-    if (typeof lastLineEndsSentence === 'boolean' && !lastLineEndsSentence && /^[A-Z]/u.test(line)) {
+    if (typeof lastLineEndsSentence === 'boolean' && !lastLineEndsSentence && /^[A-Z][a-z]/u.test(line)) {
       return true;
     }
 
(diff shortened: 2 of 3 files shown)


brettz9:
Thank you for this PR. I'm wondering though whether, given the interest in this rule, we should, rather than reinventing the wheel re: parsing English grammar, instead use something like https://www.npmjs.com/package/parse-english . (I see there is also https://www.npmjs.com/package/sentence-extractor but this is AGPL.)

(Btw, I know it'd be ideal to be generic in parsing sentences across languages for this, but 1) I don't think there is npm software for doing this, and 2) We already have requests for more English-specific handling such as for "etc." abbreviations.)

whitneyit:
I'm not sure if using `parse-english` (or similar) would be in the scope of this PR. I would assume that parsing English grammar would affect many rules.

Though not perfect, this PR fixes numerous false positives that I have encountered.

brettz9:
I'm not asking you to implement it in this PR. I'm just wondering whether we should replace your PR (and thus also fix the false positives). It shouldn't be hard to add it, and we only need to add the parser for this rule.

brettz9:
I see that the `parse-english` parser does not report bad grammar.<sup>1</sup> We'll therefore still need code to check the same issues if using this parser, I think using such a distinct tokenizer still makes sense as I think it makes for cleaner code than trying to handle all the regexes and string parsing ourselves.

<sup>1</sup> For example, I see from `parse-english` that:

1. As long as it is preceded by sentence punctuation, even single newlines will be treated as a new "paragraph".
2. If there is no preceding sentence punctuation (like a period), newlines will be parsed as just whitespace nodes (within whatever paragraph/sentence started earlier). To use this parser then, one would need to go through the word tokens to look for capitalized words and look for any preceding whitespace.
3. Uncapitalized sentences like "Hello. how are you?" will be parsed as though "Hello" is a word token followed by a punctuation token, but with all of this text as children of just one sentence token rather than two.  To use this parser then, one would need to go through the word tokens to look for punctuation tokens and see if the preceding "word" is an abbreviation like "etc" or if it is otherwise like in this case.

PR #447, opened 2019-12-14
test(no-types): check that the fix for no-types preserves asterisks

skeggse:
Currently, the fix for `no-types` results in the removal of some asterisks from multiline descriptions. This makes `eslint-plugin-jsdoc` somewhat jarring to adopt.

The proper fix here probably requires https://github.com/syavorsky/comment-parser/pull/74.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5295,6 +5295,17 @@ function quux () {
 
 }
 // Message: Types are not permitted on @returns.
+
+/**
+ * Beep
+ * Boop
+ *
+ * @returns {number}
+ */
+function quux () {
+
+}
+// Message: Types are not permitted on @returns.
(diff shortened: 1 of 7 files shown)


brettz9:
I see this is not working even after the `comment-parser` update.

brettz9:
Btw, I've filed https://github.com/syavorsky/comment-parser/pull/78 which should fix the underlying issue, and then once it may be merged, I've added a commit to your PR which should get things fixed on our end as well.

brettz9:
If you already updated, please update further to 24.0.6 as it includes some fixes for a regression introduced in 24.0.3 (on which this merged PR built).

PR #456, opened 2019-12-26
Refactor jsdoc iteration to walk through nodes comments are associated with

### LINKED ISSUE #363: no-undefined-types: False positives with class templates
Test case 
```js
    {
      code: `
      /**
       * @template TEMPLATE_TYPE
       */
      class Foo {
        /**
         * @return {TEMPLATE_TYPE}
         */
        passes () {}
      
        /**
         * @return {TEMPLATE_TYPE}
         */
        fails () {}
      }
      `,
    },
```

The first method (`passes()`) does not raise any erroros while the second one (`fails()`) does.
The reason for this is that `getClassJsdoc()` returns null for the second method.

This was introduced in a1c9824be2bbbf1766e04145a4e4c7a52bd9935d. Namely, `getAncestor` implementation walks through tokens rather than AST nodes and happens to work for the first node because the third token is `class` keyword while for the second method it's a `{` punctuator.

Another way to break this would be declaring the class as `class Foo extends Bar`, this way the third token from the node would be `Foo` identifier. There should be plenty of other ways as well.

This potentially affects all calls to `getClassJsdoc` and `getClassNode` so it might have impact on some other rules as well.

l1bbcsg:
In attempt to fix #363 I changed the way iterating over jsdoc comments works.
Previously, `iterateJsdoc` walked over comment nodes and processed them without regards of their context and most importantly, the AST node they're describing.
I changed this to iterate overs all js AST nodes, then selecting ones that have JSDoc comments and processing those along with js their node. This allowed to implement `getClassNode` properly and should probably allow more context-aware checks.

Some JSDoc comments are not associated with any nodes however and so they warranted a second pass on `Program:exit`.
Also `getJSDocComment` attributes JSDoc to several nodes, for example here:

```js
/**
 * Comment.
 */
const foo = {
   bar: 'baz'
}
```
`getJSDocComment` will return the JSDoc node for both `Identifier` and `ObjectExpression`, unlike base eslint `SourceCode.getJSDocComment`. I'm not sure if this is the intended behaviour or nor.
Because of this two things I had to keep track of JSDoc nodes processed while iterating js nodes.

This ended up being a major change in internal behaviour, but I believe it should be backwards compatible. 
Another negative consequence is that it increased processing time. Linting one of my JSDoc heavy GCC project now takes 2 seconds longer (14 from 12). But tests run as fast as previously, so exact impact depends on complexity of the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5044,6 +5044,22 @@ class Foo {
 // Settings: {"jsdoc":{"mode":"closure"}}
 
 /**
+ * @template TEMPLATE_TYPE
+ */
+class Foo {
+  /**
+   * @return {TEMPLATE_TYPE}
+   */
+  bar () {}
+
+  /**
+   * @return {TEMPLATE_TYPE}
+   */
+  baz () {}
+}
+// Settings: {"jsdoc":{"mode":"closure"}}
+
+/**
  * @template TEMPLATE_TYPE_A, TEMPLATE_TYPE_B
  */
 class Foo {
--- a/src/eslint/getJSDocComment.js
+++ b/src/eslint/getJSDocComment.js
@@ -127,6 +127,7 @@ const getJSDocComment = function (sourceCode, node, settings) {
   };
 
   const reducedNode = getReducedASTNode(node, sourceCode);
+  /* istanbul ignore next */
   if (!reducedNode) {
     return null;
   }
@@ -134,4 +135,5 @@ const getJSDocComment = function (sourceCode, node, settings) {
   return findJSDocComment(reducedNode);
 };
 
+export {getReducedASTNode};
 export default getJSDocComment;
--- a/src/iterateJsdoc.js
+++ b/src/iterateJsdoc.js
@@ -419,59 +419,61 @@ const makeReport = (context, commentNode) => {
 const iterateAllJsdocs = (iterator, ruleConfig) => {
   const trackedJsdocs = [];
 
-  const callIterator = (context, node, jsdocNode) => {
+  const callIterator = (context, node, jsdocNodes) => {
     const sourceCode = context.getSourceCode();
-    if (!(/^\/\*\*\s/).test(sourceCode.getText(jsdocNode))) {
-      return;
-    }
-    const sourceLine = sourceCode.lines[jsdocNode.loc.start.line - 1];
-    const indent = sourceLine.charAt(0).repeat(jsdocNode.loc.start.column);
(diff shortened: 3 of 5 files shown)


brettz9:
Looks to be very good, and after some changes (see https://github.com/interfaced/eslint-plugin-jsdoc/pull/1 ), I think the performance concerns are effectively mitigated (perhaps even better than before).

I'm not sure what you mean by:

> getJSDocComment will return the JSDoc node for both Identifier and ObjectExpression

Did you mean `VariableDeclaration` and `ObjectExpression`, because those are the only node types I see which give non-null for a JSDoc node with that code (changing our `getJSDocComment` test code to report on `'Program:exit'` and`'*:not(Program)'`)?

l1bbcsg:
Thanks for performance improvements, this really helped.
Some sample timings from my tests can found in [this comment](https://github.com/interfaced/eslint-plugin-jsdoc/pull/1#issuecomment-569056700).

l1bbcsg:
> Did you mean `VariableDeclaration` and `ObjectExpression`, because those are the only node types I see which give non-null for a JSDoc node

Yes, I might have confused the exact node, sorry.

I think in general it's pretty important which JS elements JSDoc gets attributed to and perhaps there should be always one and only one node that JSDoc is attached to, but it's not necessary for this MR or `no-undefined-types` fixes it brings. I attempted to change this behaviour, but then a lot of other rules became affected and some tests failed, so instead I tried to reduce amount of nodes with `getReducedASTNode` and then also track comments to make sure each is only checked once.

brettz9:
Yes, unfortunately, I'm not sure there can be a 1-to-1 mapping because besides the fact that we allow our users currently to express any context they want, sometimes they need to do this.

For example some projects may have some comments like this:

```js
const foo = /**
     * Comment.
     */ {
       bar: 'baz'
    }
```

...this will report at `ObjectExpression` but not `VariableDeclaration`.

And of course here only `VariableDeclaration` will be available:

```js
    /**
     * Comment.
     */
    const foo = 5
```

If we only check the comment immediately above `ObjectExpression`, then the user will be forced to check all `VariableDeclaration`s so they don't miss:

```js
/**
 * Comment.
 */
const foo = {
   bar: 'baz'
}
```

...and maybe they are not interested in also documenting `const foo = 5`.

The only alternative I can think of is to support specifying nested contexts like `VariableDeclaration > ObjectExpression`, but I'm afraid this would become more complicated both for users and ourselves.

I will look to merge now as this seems ready though feel free to comment further.

PR #458, opened 2019-12-27
`require-property*`: rules for ensuring `@property` has name, type, description

### LINKED ISSUE #409: Rules to check `@property` type, name, or description
Similar to #406, we should be able to check `@property` tags (e.g., on `@typedef` or `@namespace` and as apparently suggested by the [docs](https://jsdoc.app/tags-property.html), with `@class` as well). Though we can't readily check the source for matching names, we can at least check any present property's type, name, and description, as with the `require-param-*` and `require-returns-*` rules.

At least for `@typedef`, the context can be anywhere, so I suggest we either accept a `contexts` option or make it apply everywhere, possibly subject to the presence of particular tags (e.g., when on `typedef` only).

brettz9:
feat(require-property*): add rules for ensuring property tags  have a name, type, and/or description (in any context); fixes #409

BREAKING CHANGE:

Adds rules to `recommended`


brettz9:
Bundled with #459.

PR #460, opened 2019-12-28
`require-property`: New rule to check for `@property` on `@typedef` or `@namespace` for plain object type

### LINKED ISSUE #410: Rule to require that `@typedef` (and maybe `@namespace`) with `object` type have at least one `@property`
When one defines a `@typedef` (and maybe for `@namespace` as well), when defining it as a `object`, `Object`, or `PlainObject`, this will typically only be meaningful if there are specific properties documented thereon.

The proposed rule would not report errors for signatures like `@typedef {Object<string, string>}` as these at least provide some idea of the value type.

brettz9:
feat(`require-property`); add rule to check for `property` on `typedef` or `namespace` where the type is `object`; fixes #410

BREAKING CHANGE:

Adds new rule to `recommended` config.


brettz9:
Bundled with #459.

PR #461, opened 2019-12-28
`check-types`: add `exemptTagContexts` option

### LINKED ISSUE #255: `check-types` option to allow `@typedef` to be exempt from `preferredTypes` setting (at least rejections)
While one may wish to enforce, at least for rejections, `preferredTypes` on `@typedef` in the `check-types` rule as is now the behavior, sometimes the reason for rejecting a type is because one wishes to use a more specific tag, but `@typedef` may need to use a generic type to indicate its base type.

brettz9:
feat(`check-types`): add option `exemptTagContexts` to exempt type-checking (certain types or any types) on specific tags; fixes #255

### DIFF
--- a/.README/rules/check-types.md
+++ b/.README/rules/check-types.md
@@ -27,6 +27,18 @@ RegExp
   - with the key `noDefaults` to insist that only the supplied option type
     map is to be used, and that the default preferences (such as "string"
     over "String") will not be enforced. The option's default is `false`.
+  - with the key `exemptTagContexts` which will avoid reporting when a
+    bad type is found on a specified tag. Set to an array of objects with
+    a key `tag` set to the tag to exempt, and a `types` key which can
+    either be `true` to indicate that any types on that tag will be allowed,
+    or to an array of strings which will only allow specific bad types.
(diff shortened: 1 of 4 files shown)


binury:
@brettz9  Was there some/any issue with using expressions? Having to specify the exact match means that custom types can't be used without an inline disable or editing the eslintrc like:

```json
{
      "jsdoc/check-types": [
      2,
      {
        "exemptTagContexts": [
          {
            "tag":"type",
            "types":[
              "Object.<string, string>",
              "Object.<string, object>",
              "Object.<string, number>",
              "Object.<string, boolean>",
              "Object.<string, SomeCustomType>",
              "Object.<string, Flower>"
            ]
          }
        ]
      }
    ],
}
```
vs, e.g.,
```json
{
      "jsdoc/check-types": [
      2,
      {
        "exemptTagContexts": [
          {
            "tag":"type",
            "types":[ "Object.<*, *>" ]
          }
        ]
      }
    ],
}
```

brettz9:
@binury : No guarantees I'll get to looking into it, but should be best to file as a separate issue so this doesn't get lost...

PR #465, opened 2019-12-29
Global function checking

### LINKED ISSUE #406: Allow `require-param-*` and `require-returns-description`/`require-returns-type`/`require-returns` rules to check `@callback` and `@function` tags
The `@callback` tag has no associated context, so it would need to be checked throughout the file, but although we can't check `require-param` if we don't know the JavaScript function to which it applies, we can at least check `require-param-name`, `require-param-type`, and `require-param-description` if enabled for this tag.

Similarly, we can apply `require-returns-description`/`require-returns-type` and a modified form of `require-returns` (i.e., like `forceRequireReturn` which, if enabled, always checks its presence); `require-returns-check` would not be applicable.

The `@function` tag is mentioned here since in the case of `@interface` functions, these may not appear above an actual JavaScript function. See https://jsdoc.app/tags-interface.html .

brettz9:
Builds on #459, #462, and #464 (depends on some refactoring), so probably easier to review those first.

feat(`require-param-*`, `require-returns-*`, `require-returns`): allows `contexts` option to be set to "any" to check virtual function docs like `@callback` or `@function` with `@interface`; fixes #406

(Since it allows `contexts`, whose behavior in other functions is to override, if a project does not wish docs on all function types, they can now do so as well.)

### DIFF
--- a/.README/rules/require-param-description.md
+++ b/.README/rules/require-param-description.md
@@ -2,6 +2,18 @@
 
 Requires that each `@param` tag has a `description` value.
 
+#### Options
+
+##### `contexts`
+
+Set this to an array of strings representing the AST context
+where you wish the rule to be applied.
+Overrides the default contexts (see below). Set to `"any"` if you want
+the rule to apply to any jsdoc block throughout your files (as is necessary
+for finding function blocks not attached to a function declaration or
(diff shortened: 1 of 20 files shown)


brettz9:
Included with #468

PR #471, opened 2020-01-03
WIP: docs: break documentation into multiple files

### LINKED ISSUE #365: Add summary of rules in readme
I think the readme can be improved with the addition of a summary of rules (a list of rules with short description). Users can use it to 1) find a rule 2) scan all available rules.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/78805867-add-summary-of-rules-in-readme?utm_campaign=plugin&utm_content=tracker%2F23037809&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23037809&utm_medium=issues&utm_source=github).
</bountysource-plugin>

zypA13510:
I don't want to be harsh, but having all the information in one file, especially with all the long snippets, makes the documentation very painful to read.

I believe we can break it down to individual rules + an overview + guides instead, just like what ESLint did, which should make it more readable (and easier to maintain?). Also, you can use GitHub Pages to render all the docs into live HTML.

And I also intend to change some formatting to be as consistent with ESLint docs as possible, e.g. layout of sections.

This would be a lot of work. For starters, I have created a page for `check-access` and I'd like to get your feedback, so I don't waste time working on something you would eventually reject for some reason.

This will close #365 when finished.


brettz9:
While I personally think it would be a very good idea to move documentation to single pages at this point, this is currently done through convenient auto-generation based on whatever is within `test/rules`. So we can't just manually break up the README like this. `.README/README.md` is the main template that controls the final layout of the main README, and the other files like `.README/rules/check-access.md` are templates.

We are currently using the tool [gitdown](https://github.com/gajus/gitdown). I wonder whether it can be made to support writing sections to separate files rather than to separate headings, as that should indeed be helpful, as we could still get the benefits of auto-integration of tests and auto-generation of a Table of Contents (though one Table of Contents across files, and one within each page).

brettz9:
Actually I guess it is already using manual include statements, so if @gajus is ok with the change, we can see about changing these to links and have gitdown instead process Table of Contents for each individual file instead.

zypA13510:
Yes, I noticed that gitdown is used, and that examples are being generated from tests.

However, in my opinion, maybe it's better if we separate tests and documentation. If you check the examples in README, many of them are needlessly long. Maybe it's good to be verbose and thorough for tests, but do we really need to point out the obvious in documentation? Take `check-access` for an example, there is a case about `@access foo` and `ignorePrivate`, which if you ask me, should be left out (because `ignorePrivate` is documented elsewhere and `@access foo` probably had nothing to do with it).

To summarize, my idea is to completely abandon the gitdown approach. A new docs folder with rules/guides will be created, and it can be manually modified. The upside to this approach is, imo, improved maintainability - you can reword the doc, choose concise and easy-to-understand examples, all without relying on much else. This will be similar to the official eslint repo.

The downside of course is, the doc needs to be updated if expected behavior changes. Maybe we can include a check in PR template like many other projects do:
```
- [ ] new behavior is properly documented
```

brettz9:
Though I think separate files is a very good idea, as one of the maintainers, I don't want to have to work with a manual solution. I think an auto-generated Table of Contents is useful and desirable.

As far as the tests, they should all be relevant, even while I can agree that they are not all of great interest to average users. Some tests ensure that coverage is fully complete, and some tests necessarily check `settings` behavior which is documented elsewhere because settings act globally across multiple rules, but specific rules still need to test this behavior, and it should help users to see how those settings are employed with the rule and what their consequences are.

In the example you cite, this is relevant because `@access private` is supposed to be equivalent to `@private`, and we now allow `@access private` to cause the `ignorePrivate` setting to work (except on the `check-access` rule). That test is confirming that `ignorePrivate` does not cause the rule to be ignored as is the case for other rules.

That all being said, I'd be ok with a boolean being added to the tests to indicate whether it should be included in the docs or not (with the default being "true") and a link being generated to the tests (i.e., pointing to users where they can find more examples in the `tests` folder).

zypA13510:
Well, if you say so. Then I guess I can close this PR? Because if you would like to avoid a manual solution, much of the changes here won't help, I guess>

brettz9:
I suggest opening an issue for putting the docs into separate files. I also like the table format you started, but I presume it'd be easier to add that when automation is in place, so yeah I think this PR can be closed. Feel free to comment further though as needed, or, again, open an issue.

PR #475, opened 2020-01-08
Create separate pages for docs

### LINKED ISSUE #365: Add summary of rules in readme

### LINKED ISSUE #472: Document one rule per file
As it currently stands, the README contains all the documentation of every rule. This makes it very long and hard to read. We could break them into separate files, and add an overview that links to each file.

Related: #365, #471

brettz9:
docs: create separate pages for docs (fixes #472), add/adjust table headers and contents, add fixer section placeholders

Add summary of rules (fixes #365  ).

Includes some other misc. document fixes/clarifications.

(See what it looks like (and its linked rules) from https://github.com/brettz9/eslint-plugin-jsdoc/blob/doc-multi-pages/README.md .)


gajus:
What is the benefit of multiple pages over what we have at the moment?

In both cases user needs to use index to access individual rules.

The benefit of having everything on one page is that user can use browser's "search in page" functionality. This is particularly useful when user can only recall the rule/ configuration he is looking by some keywords or when trying to identify rules that share the same configuration.

gajus:
I do like the index you've added though and it addresses the issue described in https://github.com/gajus/eslint-plugin-jsdoc/issues/365.

Perhaps we just need to add that to the top of the current documentation.

brettz9:
I'm personally ok without changing to separate files (and I also like to take advantage of browser search as you suggest), but besides load time issues (here in China, for example, it can be kind of slow to load foreign sites), I think when users see one big page at once, they may not all be in the habit of looking in sequence from the top. They may instead scroll around a little to get a bird's eye view, and/or get intimidated seeing the amount of contents, with them coming to the false conclusion that there is a lot of config needed to get started. And if one loses one's place by an errant search, it can be harder to find one's way back.

Maybe @zypA13510 or others can provide any other rationales. But again, I'm personally ok with just an index.

zypA13510:
I think the primary reason is the length of examples. I want to review each rule before I enable them, so, given the current structure, it makes sense to just scroll down continuously. But soon I discovered that I'm spending most of the time staring at examples I have no interest in. I want to quickly scroll past them, but at the same time, I'm afraid to inadvertently skip a rule.
- This is why I suggest picking examples manually, to reduce the length of examples, and this way we can also separate basic behavior and additional behaviors controlled by options, like ESLint does, which should make examples more organized thus more comprehensible.
- Separate them into pages will also help because usually, the most important bits of information is at the top. I can open multiple rules in tabs, then after I finish reading information that concerns me, close the tab and start reading the next one.

Another reason would be the consistency with official ESLint docs. This is a plugin of ESLint, most readers would probably also be readers of ESLint docs, I know I am, and I've been reading ESLint docs from time to time for a few years now. It wouldn't hurt to be more consistent in style, I guess?

gajus:
> This is why I suggest picking examples manually, to reduce the length of examples, and this way we can also separate basic behavior and additional behaviors controlled by options, like ESLint does, which should make examples more organized thus more comprehensible.

This sounds reasonable and fairly easy to implement. Just an additional property in the test setup. However, it is quite a bit of work to go through all the tests and select those that should be included.

If scrolling through an examples is the problem, though then why not simply collapse examples such as this:

<details>
  <summary>Click to expand</summary>
  whatever
</details>

brettz9:
Closing in favor of #499. If I can get that working with gitdown, I can hopefully add expanded-by-default section containers so people can narrow their view down to some degree.

PR #480, opened 2020-01-15
fix:`check-values` for `@version` and `@since` reported invalid version with leading/trailing whitespace

sbusch:
`semver.valid()` requires a trimmed string: with leading or trailing whitespace, `semver.valid()`always returns `false`

### DIFF
--- a/README.md
+++ b/README.md
@@ -3891,7 +3891,21 @@ function quux (foo) {
 }
 
 /**
+ * @version      3.4.1
+ */
+function quux (foo) {
+
+}
+
+/**
  * @since 3.4.1
+ */
+function quux (foo) {
+
+}
+
+/**
+ * @since      3.4.1
  */
 function quux (foo) {
 
--- a/src/rules/checkValues.js
+++ b/src/rules/checkValues.js
@@ -15,8 +15,8 @@ export default iterateJsdoc(({
   } = options;
 
   utils.forEachPreferredTag('version', (jsdocParameter, targetTagName) => {
-    const version = jsdocParameter.description;
-    if (!version.trim()) {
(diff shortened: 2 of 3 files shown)


brettz9:
I think this may be working as intended. The reason for the `trim` on the check was to avoid it saying it was "Missing" when it really did have something (even if just some extra whitespace).

But trimming the version value itself suggests allowing extra whitespace with a valid version to be treated as valid, and I'd personally like to avoid this being treated as valid (I like the enforcement of searching through files in a predictable manner, e.g., one could rely on searching for `@version \d`).

If a change were to be made, I would think we could just remove the `trim()` entirely (since it's not actually needed here at all in this case).

Did you come across this in your own code or just in browsing our library code?

sbusch:
I stumbled about this here: <https://github.com/autoNumeric/autoNumeric/blob/next/src/AutoNumeric.js> where the JSDoc tag values are aligned with whitespaces.

brettz9:
Makes sense. I've added. We want tests for any change, so just letting you know if you submit other PRs in the future (I added a couple this time).

Btw, FWIW, I noticed your branch on `@throws`. That sounds like a good idea, but I recommend adding instead as an option to `valid-types` similar to `checkSeesForNamepaths` (but instead of checking for namepaths, it would be checking types; and actually, come to think of it, I think it should accept an array instead so users could require types on other tags where it is normally optional, e.g., `typedef`). (It should be optional though as it technically can be a free-form description.)

sbusch:
Thanks for your quick responses & the merge!

Tests: thanks for the hint, I added some for my new PR #481 (the one about @throws). I'll add your other comments to this PR (which I sent before seeing your comment).

PR #487, opened 2020-01-24
swap object.entries-ponyfill for object.entries

edwardgalligan:
# Motivation

Getting the following error on an automated licence check:
```
object.entries-ponyfill@1.0.1 -- license null not allowed
   - "eslint-plugin-jsdoc" depends on it
   - Hoisted from "eslint-plugin-jsdoc#object.entries-ponyfill"
```

Could whitelist it in the licence check, but looking into it, only one issue exists on the `object.entries-ponyfill repo, and it's about this issue https://github.com/m59peacemaker/js-object.entries/issues/2 While they have a LICENSE file in their repo, from reading that issue, it seems they don't have rights to publish under that license.

## Further rationale

According to their README, the raison d'être for creating `object.entries-ponyfill` was to reduce filesize compared to `object.entries`; this shouldn't be too relevant for linting.


brettz9:
@gajus (and any others): What do you think if we just make a change to require Node 8+ ? With Node 8, we could simply use the native `Object.entries`.

Even Node 8 recently reached end of life, so some projects are requiring Node 10+ even.

gajus:
I am Okay with that.

PR #490, opened 2020-02-07
New rule require-deprecated-description

l1bbcsg:
`@deprecated` annotations without description about the reasoning behind deprecation and instructions on what to do to avoid deprecated apis can be very frustrating. This rule aims to help to document more helpful deprecation messages.

Functionally it's pretty similar to `require-param-description`, except without context parameter.

### DIFF
--- a/.README/rules/require-deprecated-description.md
+++ b/.README/rules/require-deprecated-description.md
@@ -1,0 +1,5 @@
+### `require-deprecated-description`
+
+Requires that each `@deprecated` tag has a `description` value.
+
+<!-- assertions requireDeprecatedDescription -->
--- a/src/index.js
+++ b/src/index.js
@@ -17,6 +17,7 @@ import noBadBlocks from './rules/noBadBlocks';
 import noDefaults from './rules/noDefaults';
 import noTypes from './rules/noTypes';
 import noUndefinedTypes from './rules/noUndefinedTypes';
+import requireDeprecatedDescription from './rules/requireDeprecatedDescription';
 import requireDescriptionCompleteSentence from './rules/requireDescriptionCompleteSentence';
 import requireDescription from './rules/requireDescription';
 import requireExample from './rules/requireExample';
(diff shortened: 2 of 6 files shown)


brettz9:
Do you think this rule can be avoided in favor of just the following:

```js
{
  'jsdoc/match-description': ['error', {tags: {
    deprecated: '\\S+'
  }}]
}
```

l1bbcsg:
Damn, sorry, it didn't even occur to me to try `match-description`.

Yes, the same can be achieved with a regex in `match-description`, the rule settings just require some more configuration and more complex regexps if it's also used for other purposes (for example we use it to forbid non-english comments).

So yeah, `require-deprecated-description` is somewhat redundant but also allows a bit more atomic configuration. Same could be said about `require-description-complete-sentense` I guess.

I'm not sure if it should be accepted or not, guess it's up for discussion.

brettz9:
> the rule settings just require ... more complex regexps if it's also used for other purposes (for example we use it to forbid non-english comments).

Maybe allow an array of AND-ed regexes to handle such cases (OR can be handled by `|`) could mostly address this use case?

> Same could be said about `require-description-complete-sentense` I guess.

That rule does more than a check of `[A-Z].*[.?!]`, actually parsing paragraphs and what not in a manner that I don't think regexes could even manage, IIRC.

brettz9:
And no worries re: reporting an overlapping rule--we have an inevitable mix of rules named after tags and others named after features, or even some not really suggested well in rule names at all (e.g., some determined by `settings`).

brettz9:
I think I will close, as there was a desire here to keep the project as small as possible (though without preventing genuine new utility).

However, feel free to open an issue to allow an array of AND-ed regexes to handle cases, e.g., where you wanted a description to match say the pattern of initial caps and of preventing non-English? And of course, if there is some functionality that specifically applies to `@deprecated`, feel free to file that. Best wishes...

PR #498, opened 2020-03-01
Add support for checking destructured properties and TypeScript literal object types

dstaley:
This PR modifies the `check-param-names` rule to verify that destructured object properties are documented with a `@param` directive. Additionally, it verifies that the properties on TypeScript object literal types also have accompanying documentation, and checks that there aren't any `@param` directives for non-existent properties.

With this PR, the following are now errors:

```js
/**
 * @param cfg
 * @param cfg.foo
 */
function quux ({foo, bar}) {
}
// Message: Missing @param "cfg.bar"
```

```ts
export class SomeClass {
  /**
   * @param prop
   * @param prop.foo
   */
  constructor(prop: { foo: string, bar: string }) {}
}
// Message: Missing @param "prop.bar"
```

```ts
export class SomeClass {
  /**
   * @param options
   * @param options.foo
   * @param options.bar
   */
  constructor(options: { foo: string }) {}
}
// Message: @param "options.bar" does not exist on options
```

`getFunctionParameterNames` will now report destructured parameters and object literal types as a tuple `[?string, Array<string>]`, where the first element represents the parameter name if provided, otherwise `undefined` in the case of destructiring. The second element represents the properties of the parameter.

```ts
function quux({ foo, bar }) {} // [undefined, ["foo", "bar"]]

function quux(options: { foo: string, bar: string }) {} // ["options", ["foo", "bar"]]
```

I tried to be as exhaustive as possible, but please let me know if there's a particular case I didn't write a test-case for. I also didn't touch and code related to providing fixes, so please let me know if I ended up breaking something there that wasn't accounted for in the unit tests.


brettz9:
Looks great, thanks! I need to take a closer look, but for now, I've rebased on some just-updated `master` changes, fixing the Travis errors.

While there may be some duplication involved, I think it would be better if `require-param` could be made to report the destructuring-ish errors rather than ignore them. It already is reporting some param mismatches, so I see no reason we should not do so there also if we are doing validation in both. I would probably want to add that, so if you could add it, I expect it would expedite my review.

Thanks!

dstaley:
Sure, I'm not opposed to that! I'm a bit confused though: would that mean both rules would report the same errors? Or would one handle the TypeScript object literal property checks and the other handle destructuring?

If both rules report the same errors, would the tests for each rule confirm that errors from the other rule are also reported?

brettz9:
I think both should report the same kind of errors (as they both report errors now--just not with destructuring).

However, `require-param` only iterates through the actual function parameter names, not concerning itself if the user has added too many `@param`'s (as does `check-param-names`). I think we can keep this behavior.

So in summary, I think the same tests can be used for both with the exception of tests checking extra `@param` tags (`require-param` should be checked to ensure it gives no errors whatever these extra params say and `check-param-names` should report problems with extra `@param` tags (as it already does with non-destructuring-type params)).

dstaley:
Just so that I fully understand before I dive in, you're asking for the following to be errors, correct?

```js
[
    {
      code: `
          /**
           *
           */
          function quux ({foo}) {

          }
      `,
      errors: [
        {
          message: 'Missing JSDoc @param "foo" declaration.',
        },
      ]
    },
    {
      code: `
          /**
           * @param
           */
          function quux ({foo}) {

          }
      `,
      errors: [
        {
          message: 'Missing JSDoc @param "foo" declaration.',
        },
      ]
    },
    {
      code: `
          /**
           * @param options
           */
          function quux ({foo}) {

          }
      `,
      errors: [
        {
          message: 'Missing JSDoc @param "options.foo" declaration.',
        },
      ]
    }
]
```

If so, what should the error for the second test case be since we don't have a name for the argument?

brettz9:
Right, those should be errors.

I think for case no. 2, it might be nice to report `Missing JSDoc @param declaration for root of "foo".` (I think it would be sufficient to report the first destructured name rather than needing to say `Missing JSDoc @param declaration for root of "foo", "bar", and "baz".` had the argument looked like `function quux ({foo, bar, baz: boo}) {`)

Btw, speaking of that example, I think we ought to be showing the key instead of the value since the key is what consumers of the function need to be concerned about, and knowing the value is just about tracing it back to where it is defined (even if quite early, it is still after `baz` is destructured).

dstaley:
Gotcha!

As far as displaying "the key instead of the value": could you explain a different way or provide an example? I'm not sure what you mean.

brettz9:
To change your last example, if it were:

```js
{
      code: `
          /**
           * @param options
           */
          function quux ({foo: bar}) {

          }
      `,
      errors: [
        {
          message: 'Missing JSDoc @param "options.foo" declaration.',
        },
      ]
    }
```

...the error still references `foo` even though it destructures (immediately) to `bar`. We want the key `foo` in the message (unless you think listing both will not be confusing, e.g., `Missing JSDoc @param "options.foo (bar)" declaration.`).

dstaley:
Ah, okay yeah that makes sense now! I think I have all the answers I need to go through this over the next few days. Thanks again for your help!

brettz9:
Also, not just in messages:

```js
          /**
           * @param options
           * @param options.bar
           */
          function quux ({foo: bar}) {

          }
```

...should be tested and give an error (since the `@param` should have been named `options.foo`).

brettz9:
And, btw, your examples made me realize I think we want a test (unrelated to destructuring) like this for `require-param` too to check for the case when there are the same number of arguments, just named differently:

```js
{
      code: `
          /**
           * @param bar
           */
          function quux (foo) {

          }
      `,
      errors: [
        {
          message: 'Missing JSDoc @param "foo" declaration.',
        },
      ],
      output: `
          /**
           * @param foo
           */
          function quux (foo) {

          }
      `,
    },
```

dstaley:
Good news! I finally managed to get some time to implement checking properties for arbitrarily nested destructuring assignments! The tests are now passing, including all the cases you've mentioned here.

There's probably a few places where performance can be improved, so I'm gonna take a few more days and see if anything stands out to me.

dstaley:
Alrighty, I think it's ready for a review! If there's anything that's unclear, please let me know. I also added a new TypeScript test case using an actual function definition from one of my codebases, and it's really exciting to see it accurately reporting missing parameters!

test/rules/assertions/checkParamNames.js
```diff
@@ -231,7 +231,7 @@ export default {
            * @param cfg.foo
            * @param cfg.foo
            */
-          function quux ({foo, bar}) {
+          function quux ({foo}) {
```
brettz9: Could you indicate what the reasons were for each of the cases where the tests were modified? That should make reviewing faster. While we normally hopefully won't need to refactor old tests and should just add tests, I understand that in these tests we were previously ignoring destructured properties rather than reporting them. But if you could indicate for each change why the change was needed (whether a removal or even an addition to a preexisting test), that would help (e.g., why were underscores or numbers added to some variable names, etc.). I'll mark the specific lines besides this one.

test/rules/assertions/checkParamNames.js
```diff
@@ -250,7 +267,7 @@ export default {
            * @param [cfg.foo]
            * @param baz
            */
-          function quux ({foo, bar}, baz) {
+          function quux ({foo}, baz) {
```
brettz9: Also here...

test/rules/assertions/checkParamNames.js
```diff
@@ -266,20 +283,76 @@ export default {
           /**
            * @param cfg
            * @param cfg.foo
-           * @param [cfg.foo="with a default"]
```
brettz9: I assume this was due to our previously silently ignoring destructured. Is there a rule elsewhere checking duplicate destructured param names? I think that should be reported if not.

test/rules/assertions/requireParam.js
```diff
@@ -414,19 +480,27 @@ export default {
       /**
        *
        */
-      function quux ({bar, baz}, foo) {
+      function quux ({bar, baz___}, foo) {
```
brettz9: Also here...

test/rules/assertions/requireParam.js
```diff
@@ -414,19 +480,27 @@ export default {
...
         },
       ],
       output: `
       /**
+       * @param root.bar
+       * @param root.baz___
        * @param foo
        */
-      function quux ({bar, baz}, foo) {
+      function quux ({bar, baz___}, foo) {
```
brettz9: Also here...

test/rules/assertions/requireParam.js
```diff
@@ -1008,10 +1171,10 @@ export default {
        * Assign the project to an employee.
        *
        * @param {object} employee - The employee who is responsible for the project.
-       * @param {string} employee.name - The name of the employee.
-       * @param {string} employee.department - The employee's department.
+       * @param {string} employee.name1 - The name of the employee.
+       * @param {string} employee.department1 - The employee's department.
        */
-      function assign({name, department}) {
+      function assign({name1, department1}) {
```
brettz9: And these above...

test/rules/assertions/requireParam.js
```diff
@@ -1035,6 +1198,7 @@ export default {
          *
          * @param args Arguments compiled and provided by StephanClient.
          * @param args.options The options as provided by the user, or an empty object if not provided.
+         * @param args.client The options as provided by the user, or an empty object if not provided.
          * @param defaultOptions The default options as provided by the plugin, or an empty object.
```
brettz9: I guess this was just to ensure multiple properties were being checked.

brettz9:
Btw, I've pushed a couple small commits to this branch which though seemingly unrelated, were mostly meant to reduce our disabling of eslint or istanbul directives since these should really only be done with good reason (and with these minimized in prior code, we can more easily search the code to find added ignores to understand whether they were justified or not).

And where they actually are needed, I feel the istanbul ones should have an explanation as to why they are acceptable (which I've done here). Nothing to do, as it was all good, just explaining the reason for the extra commits.

test/rules/assertions/requireParam.js
dstaley: This was accidentally leftover from debugging, so I've removed it.

test/rules/assertions/requireParam.js
dstaley: This was accidentally leftover from debugging, so I've removed it.

test/rules/assertions/requireParam.js
dstaley: This was accidentally leftover from debugging, so I've removed it.

test/rules/assertions/checkParamNames.js
dstaley: This change removed the `bar` property since the test was for duplicate properties, not non-existent properties. Leaving it as `{foo, bar}` results in the test failing and only reporting the missing property.

test/rules/assertions/checkParamNames.js
dstaley: This change removed the `bar` property since the test was for duplicate properties, not non-existent properties. Leaving it as `{foo, bar}` results in the test failing and only reporting the missing property.

test/rules/assertions/checkParamNames.js
dstaley: This one was a preemptive removal based on the previous examples, but after testing I see that `cfg.foo` and `[cfg.foo="with a default"]` are treated as a single property.

test/rules/assertions/requireParam.js
dstaley: Correct, this was to ensure that the test continues to pass now that we're checking each destructured property. If you'd like I can update the property description since it looks like I just copied the description from `args.options`.

dstaley:
I've fixed a few of the issues that were ultimately unintentional; they were either leftover debugging things or instances where I thought a test would fail without the modification, but it turns out to be fine. The remaining issues are valid modifications used to keep a test either failing on the expected error or passing.

test/rules/assertions/requireParam.js
```diff
@@ -442,10 +516,18 @@ export default {
...
+        },
       ],
       output: `
       /**
        * @param foo
+       * @param root.bar
+       * @param root.baz
        */
       function quux (foo, {bar, baz}) {
       }
```
brettz9: I think we need three things here:

1. Tests for `function quux ({foo}, {bar, baz}) {` to ensure we are not adding `root` for both parameters.
1. Option to avoid a fixer for missing roots (some might not want variable names injected)
1. Configuration to change the name from "root", e.g., an array of preferred (`["root", "cfg", "config"]`) and when these are exhausted, auto-appending numbers (e.g., `root2`)

brettz9: I think we need three things here:

1. Tests for `function quux ({foo}, {bar, baz}) {` (but with no existing `@param foo` tag) to ensure we are not adding `root` for both parameters.
1. Option to avoid a fixer for missing roots (some might not want variable names injected)
1. Configuration to change the name from "root", e.g., an array of preferred (`["root", "cfg", "config"]`) and when these are exhausted, auto-appending numbers (e.g., `root2`)

brettz9: The last is less critical (esp. with no. 2) implemented.

dstaley: > 1. Tests for `function quux ({foo}, {bar, baz}) {` (but with no existing `@param foo` tag) to ensure we are not adding `root` for both parameters.

Oh, great catch! In this instance, would the names be `root.foo, root1.bar, root1.baz`?

> 2. Option to avoid a fixer for missing roots (some might not want variable names injected)

Would this option disable the fixer for all missing parameters, or just those that are destructured?

> 3. Configuration to change the name from "root", e.g., an array of preferred names (`["root", "cfg", "config"]`) and when these are exhausted, auto-appending numbers (e.g., `root2`)

So, in the example from 1, the names would be `root.foo, cfg.bar, cfg.baz`? I'm not sure I understand how that would be useful. I can see customizing what `root` is (for example, Kotlin uses `p`), so I definitely support customizing, but I'm not sure what providing more than one option accomplishes.

Just a bit more info on Kotlin: it uses `p{int}` as parameter names. So, the first example would be `p0.foo, p1.bar, p1.baz`. I'll defer to your judgement as to whether we should append a `0` for the first parameter.

brettz9: > In this instance, would the names be `root.foo, root1.bar, root1.baz`?

Yeah, good question, I guess so. One could argue for immediately jumping to `root2` or changing `root` to `root1` or as you mention for Kotlin, using "0". As long as the fixer is disable-able, I'd personally be fine with whatever choice (though maybe having all as numbers is good when they are being used to avoid the impression that the numbered ones are to be understood as a subset of the unnumbered one somehow).

> Would this option disable the fixer for all missing parameters, or just those that are destructured?

If we're auto-adding `"root"` as a name for non-destructured parameters, then yes, it should disable that behavior too. But I thought that was a new introduction only, and in such a case, it is fine to apply just for destructured.

> So, in the example from 1, the names would be `root.foo, cfg.bar, cfg.baz`?

Right, that was the idea. Maybe a better example would be `["config", "cfg", "options"]` where the names are generic enough but related (e.g., for my codebases which tend to use object options). The advantage would be just that it can avoid adding numbers.

dstaley:
Should the following be valid?

```js
/**
* @param root0.foo
* @param root1.bar
*/
function quux ({foo}, {bar}) {

}
```

brettz9:
Shouldn't it be:

```js
/**
* @param root0
* @param root0.foo
* @param root1
* @param root1.bar
*/
function quux ({foo}, {bar}) {

}
```

If you are asking whether the roots can go missing, I wouldn't think so as the only documented usage on the `@param` page includes them: https://jsdoc.app/tags-param.html#parameters-with-properties . Likewise for `@property`: https://jsdoc.app/tags-property.html#examples (btw, at some point, it'd be great for us to port your work to our property rules, but not needed for this PR).

If you're asking about something else besides the missing root, I don't see how else it could be expected.

dstaley:
Yup, I just wanted to confirm that the root is required!

dstaley:
When you say "the fixer is disable-able", how do you see that happening? Should I add it as a config option, and then only emit the `fixer` argument to `reportJSDoc` when it's set to true?

brettz9:
Right, that's the general idea, though you can make the default to be `true`.

dstaley:
Alrighty, I've added the ability to configure the unnamed root base and the ability to disable the fixer! Feel free to modify the names of variables and the options since I'm not super sold on any of them.

brettz9:
Hi thanks for the additions/changes.

But I'm sorry there seems to be a little bit lost in translation...Appreciate your patience and steadfastness here...

1. While I'm fine with any rule having an option to also being disabled in whole to fixing, the disabling of the fixer here was meant to only avoid fixing the special case where the project might not want any (root) variables auto-added, e.g., `root0.foo`. A project might want to be forced to make a choice whenever there is a missing root, about what the root should be called and not get any names added automatically in such cases. (But where there was no missing root, then the auto-fixing should normally be fine.)
2. For projects that *are* ok with auto-adding roots, the idea with the `unnamedRootBase` being an array is to prevent numbers from applying (at least until the array is exhausted). So, 

Instead of:

```js
    {
      code: `
          /**
           * 
           */
          function quux ({foo}, {bar}) {

          }
      `,
      errors: [
        {
          message: 'Missing JSDoc @param "arg0.foo" declaration.',
        },
        {
          message: 'Missing JSDoc @param "config0.bar" declaration.',
        },
      ],
      options: [
        {
          unnamedRootBase: ['arg', 'config'],
        },
      ],
```

I would expect these errors instead:

```js
{
      errors: [
        {
          message: 'Missing JSDoc @param "arg.foo" declaration.',
        },
        {
          message: 'Missing JSDoc @param "config.bar" declaration.',
        },
      ],
}
```

The idea with this option (which is only a nice-to-have) was that it would avoid numbers getting added as part of params. I would only expect numbers to be auto-added with this config if the last item were exhausted. So if the above function were:

```js
function quux ({foo}, {bar}, {baz}) {
}
```

...then it might give:

```js
{
      errors: [
        {
          message: 'Missing JSDoc @param "arg.foo" declaration.',
        },
        {
          message: 'Missing JSDoc @param "config0.bar" declaration.',
        },
        {
          message: 'Missing JSDoc @param "config1.bar" declaration.',
        },
      ],
}
```

...or if you prefer:

```js
{
      errors: [
        {
          message: 'Missing JSDoc @param "arg.foo" declaration.',
        },
        {
          message: 'Missing JSDoc @param "config.bar" declaration.',
        },
        {
          message: 'Missing JSDoc @param "config0.bar" declaration.',
        },
      ],
}
```

...because there were no more items in the array which could avoid numeric counting:

dstaley:
Heh, sorry! I really should have shared my thoughts in a bit more detail before implementing them. Thanks for the clarifications!

For disabling the unnamed roots fixer, would you mind maybe sharing some more details on how that would work on an implementation level? I'm still not totally sure how eslint applies fixers. Would I need to check settings inside `fixAll`?

For `unnamedRootBase`, I'd love to find a solution that can accommodate _always_ using numbers (like how Kotlin handles it).

For example, with your solution, having a single element in `unnamedRootBase` would produce the following:

```
@param root.foo
@param root0.foo
@param root1.foo
```

Ideally we'd avoid having something like `root` along with `root0`.

What if we always used the last element with a numeric suffix? That way, if you only specified one element, it'd always have a number, but if you had something like `["arg", "config"]`, you'd get the following:

```
@param arg.foo
@param config0.foo
@param config1.foo
```

brettz9:
> Heh, sorry! I really should have shared my thoughts in a bit more detail before implementing them.

No worries, all good!
 
> For disabling the unnamed roots fixer, would you mind maybe sharing some more details on how that would work on an implementation level? I'm still not totally sure how eslint applies fixers. Would I need to check settings inside `fixAll`?

I'd probably add a boolean within each of the items in `missingTags`, and then when iterating in the `forEach`, destructure to use that boolean to conditionally set the `fixer` to `null` (if they have the option to disable fixers in this manner). With a `null` fixer, ESLint will then avoid doing anything besides reporting it as an issue.
 
> For `unnamedRootBase`, I'd love to find a solution that can accommodate _always_ using numbers (like how Kotlin handles it).
> 
> For example, with your solution, having a single element in `unnamedRootBase` would produce the following:
> 
> ```
> @param root.foo
> @param root0.foo
> @param root1.foo
> ```
> 
> Ideally we'd avoid having something like `root` along with `root0`.
> 
> What if we always used the last element with a numeric suffix? That way, if you only specified one element, it'd always have a number, but if you had something like `["arg", "config"]`, you'd get the following:
> 
> ```
> @param arg.foo
> @param config0.foo
> @param config1.foo
> ```

Sounds like a good solution!

brettz9:
Oh, I see that the `forEach` iteration might not like my suggestion there since it is already doing some conditional applying. I guess I'd add a boolean but filter down the `missingTags` passed to `fixAll`. I'd see if something like that might work.

dstaley:
@brettz9 I just wanted to let you know that I'm still working on this! With all the stuff that's going on right now, it's been a bit difficult to find the time and energy to work on it, but I'm hoping to knock out the last little bit this weekend.

I hope you're doing well!

brettz9:
Great to hear, and no worries! Have been a bit tied up myself anyways. Is getting exciting though to come closer to a release for this, as I think this PR should get you a seat at our Round Table! :-)

Doing well here, thank you... Likewise hope all is well for you...

brettz9:
Noticed that this PR hasn't yet been tied to #11 , the issue that it (mostly) fixes. (I can't mark it as closing it unless this were also adding rest element support.)

brettz9:
@dstaley : I've moved some of the little doc additions I ended up adding to this branch while reviewing your code to `master` and then merged master back in here, so the diffs shouldn't be as convoluted--i.e., it should only show your changes now. Do you think you will be able to revisit this soon? It'd be awesome to have this merged.

If you are able to take a look, btw, it'd be nice, though not strictly necessary for this PR, to see if errors are detected with mismatches between docs and:

1. `RestElement` (e.g., `function quux (a, ...b)`) including within arrays (e.g., `function quux ([a, ...b])`)
2. `ExperimentalRestProperty` (e.g., `function quux ({num, ...extra})`)

brettz9:
I've added a commit (doing a little "Documentation-driven development)) where I added docs on how the proposed changes should work (as well as new proposed option `autoIncrementBase` and some info on how I think destructured arrays need to work and be tested). If I have a chance, I may make the fixes myself (I'll see if I have time today), but letting you know in case you get to the work first. I may also add docs later on `RestElement` and `ExperimentalRestProperty`.

brettz9:
I've added a commit implementing `autoIncrementBase`. Besides the items mentioned and documented-in-advance, we need to get the `output` added for all `require-param` (and `check-param-names`) tests. In attempting to do so, I noticed that we are not adding the root in the fixer when missing.

So I'd expect:

```js
          /**
           *
           */
          function quux ({foo}) {

          }
```

...should have `output` as:

```js
          /**
           * @param root0
           * @param root0.foo
           */
          function quux ({foo}) {

          }
```

This is how jsdoc [documents it](https://jsdoc.app/tags-param.html#parameters-with-properties).

brettz9:
I've added most of the proposed changes now. The behavior of `unnamedRootBase` still needs adjusting and `enableRootFixer` should be added.

I'd also ideally like to see `checkRestProperty` and `enableRestElementFixer` added as well.

Will see if I can get to it, but I hope you can review my changes to provide an extra pair of eyes before we merge.

brettz9:
Ok, all proposed fixes/docs/additions/tests have been added. I think this should mostly be ready now.

The one remaining issue... While not a really big problem, the one real world issue I came across when applying this to a pretty large code base (svgedit) is that sometimes one may wish to just use the parent object and let the type speak to the properties in use.

For example, we had destructuring in this function's parameters:

```js
/**
* Converts a `SVGRect` into an object.
* @function module:utilities.bboxToObj
* @param {SVGRect} bbox - a SVGRect
* @returns {module:utilities.BBoxObject} An object with properties names x, y, width, height.
*/
export const bboxToObj = function ({x, y, width, height}) {
  return {x, y, width, height};
};
```

But doc errors were given for the fact that `x`, `y`, etc. were not documented. But the `SVGRect` type ought to indicate this information (or in other cases where the type was a `module:` namepath to one of our own types).

I could refactor the above as follows to avoid needing to repeat the docs, but I think we should probably provide an option so the rule only applies if the type is missing or is "object" or "array" (or some whitelist) so that refactoring is not required:

```js
/**
* Converts a `SVGRect` into an object.
* @function module:utilities.bboxToObj
* @param {SVGRect} bbox - a SVGRect
* @returns {module:utilities.BBoxObject} An object with properties names x, y, width, height.
*/
export const bboxToObj = function (bbox) {
  const {x, y, width, height} = bbox;
  return {x, y, width, height};
};
```

brettz9:
Ok, I've implemented for `check-param-names` and `require-param` with a `checkTypesPattern` regex string option.

So this issue is now fully ready for review. Since there seems to be trouble in Github handling comparisons with this multi-merged branch, I've squashed the commits in https://github.com/brettz9/eslint-plugin-jsdoc/tree/destructured-squashed , with a complete diff of changes introduced by the PR available at https://github.com/gajus/eslint-plugin-jsdoc/compare/master...brettz9:destructured-squashed?expand=1

brettz9:
This has now been incorporated in 30835cd1760928d50e52da43285511dd60251039 . Thanks so much for your work on this!

PR #499, opened 2020-03-10
Doc revisions

brettz9:
This is meant as a replacement for #475 (fixing #365 ), bringing the improvements of a Table view of the rules, as well as expanding headings and doc text (without breaking up into separate files).

The build is not currently working though as I'd need figure out the Gitdown usage and why it is not recognizing the `rootId`'s I have used (assuming you like each rule having its own mini Table of Contents as I thought could be convenient). But I thought I'd put it out here before spending more time, to see if this is the kind of approach you'd like, @gajus .

Thanks!

Btw, I guess #472 can be closed then if we are going with this approach. We could point out that the source is still available in individual files, so users could at least look at source separately if they felt overwhelmed.

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -25,7 +25,11 @@ npm install --save-dev eslint-plugin-jsdoc
 
 ## Configuration
 
-Add `plugins` section and specify `eslint-plugin-jsdoc` as a plugin.
+To enable rules, you may use either of the following approaches in setting
+up your ESLint configuration file.
+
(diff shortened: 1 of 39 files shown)


edg2s:
I think #475 is preferable. Even on my fast device the large readme takes >2000ms to load, and on my mobile device it often hangs.

Having an index + individual readme's is the style used my most other plugins (and eslint itself).

brettz9:
Closing in favor of #940

Issue #503, opened 2020-03-18
Getters and require-returns


JacobFischer:
I'm migrating from eslint's old `valid-jsdocs` rule, to this excellent plugin. However I noticed a non-configurable difference. That is, `valid-jsdocs` does not require `@returns` documented on Getters, where-as this plugin does.

For example:

```js
// jsdoc/require-returns: "error"

class Foo {
    /** This is the description of what this getter returns. */
    get bar() {
        return "foobar";
    }
}
```

This will produce an error `Missing JSDoc @returns declaration.eslint(jsdoc/require-returns)`.

There does not seem to be much discussion around this; probably because proper Getter/Setters are somewhat niche in JS. I did find this comment however: https://github.com/gajus/eslint-plugin-jsdoc/blob/55c958dfaa54f41b225bf9ddd5afa5f2d33652f1/src/rules/requireReturns.js#L38-L39

Is there any reason for this difference? I can certainly understand circumstances in a code base where you would and would not want to require the `@returns` on a Getter. To that end I think an option on this rule would be ideal.

Unless there's some special reason why not to, I'll throw together a PR to add this feature as an option for `require-returns`.

brettz9:
We could probably get this behavior automatically if we implemented https://github.com/gajus/eslint-plugin-jsdoc/issues/272#issuecomment-600078533 (the user would need to add `exemptedContexts: ['Property[kind='get']']` to the rule), but I'd still be fine with a PR to add a specific option in `require-returns` (keeping the default as requiring its presence to avoid breaking changes and because if you want `@returns` enforced, a getter makes as much sense to have one as others (if not more)).

JacobFischer:
Awesome thanks for merging #504. I agree that node kind selectors would work too, but that would be far more complex than the option I added.

I'll close this issue as the problem I outlines is now solved. Thanks again!

PR #504, opened 2020-03-18
feat(require-returns): add checkGetter option

JacobFischer:
Implimentation for https://github.com/gajus/eslint-plugin-jsdoc/issues/503

This PR adds for the rule `jsdoc/require-returns` a new boolean option `checkGetters`, which defaults to `true`, as is the rule current behavior of the rule without the option.

When `checkGetters = false`, it allows developers to omit the `@returns` tag for `get` functions, e.g.:

```js
// jsdoc/require-returns: ["error", { checkGetters: false }]

class Foo {
    /** gets bar value of zero */
    get bar() {
        return 0;
    }
}
```

The motivation behind this option is because in regular `eslint`, the `valid-jsdoc` rule automatically ignores `get` functions, so direct porting to this plugin is now possible. This also allows for a greater flexibility as developers can choose if they do or do not want to require the `@returns` tag for these `get` functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10512,6 +10512,15 @@ async function foo(a) {
 }
 // Options: [{"forceReturnsWithAsync":true}]
 // Message: Missing JSDoc @returns declaration.
+
+class foo {
+  /** gets bar */
+  get bar() {
+    return 0;
+  }
+}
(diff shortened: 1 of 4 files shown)


brettz9:
Thanks!

PR #516, opened 2020-04-13
Add options for excluding constructors, getters and setters

### LINKED ISSUE #513: [require-param] Add a `checkSetters` option
As you mostly have a `value`-parameter in `setters` the jsdoc might be irrelevant there.
Therefore I'd recommend to add a `checkSetters` option to the `require-param` rule which acts just like the `checkGetters` option in the `require-returns` rule.

manuth:
This PR closes #513.
In some cases you might not want to have `constructor`s, `get`ters and `set`ters as the description might be redundant.
A `get`ter for example mostly has a description like `Gets or sets the display-name of the object.` which, at the same time, which makes the `@returns`-tag useless in this case.

This PR will add a `checkConstructor`, `checkGetter` and `checkSetter`-option to the `require-example` and `require-param`-rules.
It also will ad a `checkConstructor` and `checkGetter`-option to the `require-returns`-rule.

Please notice, that commit f2889fd contains breaking changes as a rule-option is removed.


manuth:
@brettz9 it took quite a long time as I'm not that experienced at procedural programming. Could you please have a look at f2889fd and 6fd9a89?
I'm not quite sure whether this is the way to go for loading the default settings as this would require that the user-configuration always is located at index `0` of `context.options`.

brettz9:
Have been a bit preoccupied. Still hope to return to this PRs as I have the chance.

src/iterateJsdoc.js
```diff
@@ -464,6 +481,16 @@ const iterate = (
...
+  const userConfig = context.options.length > 0 ? _.cloneDeep(context.options) : [{}];
+
+  if (ruleConfig.meta.schema) {
+    new Ajv({useDefaults: true}).compile(
+      {
+        items: ruleConfig.meta.schema,
+        type: 'array',
+      },
+    )(userConfig);
+  }
```
brettz9, line 493: Is this block for the intent of validating the schema? If so, that shouldn't be necessary as eslint handles that for us.

manuth: I have noticed, that `eslint` doesn't apply defaults to the `options`-array unless...
- Either a `default` is declared at the root of the meta-data:
```js
meta: {
  fixable: 'code',
  schema: [
    {
      default: {
        allowTrailingParamDocs: false // Duplicate declaration of the default  value for `allowTrailingParamDocs`
      },
      additionalProperties: false,
      properties: {
        allowExtraTrailingParamDocs: {
          type: 'boolean',
          default: false // Duplicate declaration of the default value for `allowTrailingParamDocs`
        },
      },
      type: 'object',
    },
  ],
  type: 'suggestion',
}
```

Or by using the bunch of code that I added here (which is providing an array with an empty object `[{}]` to ajv.)

brettz9: While that is a sound approach, in order to avoid extra dependencies (and in case eslint ever fixes this and then we can drop the handling entirely), I've added 49f106c1b43331be63e1886f8a50e0b5079a0247 which changes to just manually check the schema defaults of interest here.

brettz9:
Closing as this has now been included in `master` (via 49f106c1b43331be63e1886f8a50e0b5079a0247 ). It seems `semantic-release` is not working to put out a new release at the moment due to Travis being down at the moment for maintenance. If you can test against `master` though, that would be helpful.

src/iterateJsdoc.js
manuth: Awesome thank you very much 😄

PR #523, opened 2020-05-04
readme: Update shields.io links to HTTPS and fix mixed line endings

sylveon:
This fixes a mixed content issue on the npm page for this package:
image

It also normalizes line endings to LF, GitHub's online editor would previously say they are mixed, and then normalize them to CRLF, causing 11k lines diffs when you want to change a small thing.

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -1,8 +1,8 @@
 # eslint-plugin-jsdoc
 
 [![GitSpo Mentions](https://gitspo.com/badges/mentions/gajus/eslint-plugin-jsdoc?style=flat-square)](https://gitspo.com/mentions/gajus/eslint-plugin-jsdoc)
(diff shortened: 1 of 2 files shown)


brettz9:
If you revert the README carriage return changes, I can see about merging.

The issue with the carriage returns is that we have test cases that need to use carriage returns, and when Gitdown assembles them into the generated README, they are no longer shown as escapes. But this is not really a problem as the README shouldn't be edited directly anyways--only `.README/README.md`. Updates to `README.md` can and should be made by running `npm run create-readme`.

sylveon:
Ah my bad. I did this pull request after somebody online noticed the issue, so I'm not used to this project and assumed it was a mistake. I rewrote the commit history to remove those changes.

brettz9:
Thanks for the updates!

PR #531, opened 2020-05-11
feat(check-param-names, check-property-names): add `enableFixer` option

brettz9:
Allows disabling duplicate fixing.

### DIFF
--- a/.README/rules/check-param-names.md
+++ b/.README/rules/check-param-names.md
@@ -39,10 +39,10 @@ See `require-param` under the option of the same name.
 
 ##### `enableFixer`
 
-Set to `false` to avoid auto-removing `@param`'s duplicates (based on
-identical names).
+Set to `true` to auto-remove `@param` duplicates (based on identical
+names).
 
-Note that, by default, duplicates of the same name are removed even if
+Note that this option will remove duplicates of the same name even if
 the definitions do not match in other ways (e.g., the second param will
 be removed even if it has a different type or description).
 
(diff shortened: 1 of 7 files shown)


brettz9:
> I am more inclined to change to rule to never fix this case, without an option. There are other rules in the ecosystem similar to this case: `no-dupe-keys`, `no-duplicate-case`, and `imoprt/no-duplicate`. They do not provides fixes. The general idea is that "if the fix could break peoples' code, don't fix".

People can review fixes before committing them (and really should), so it can still save them time even if the user wishes to opt in to enable a potentially over-aggressive fixer. And in our case, in the vast majority of cases, at worst it should only strip out a useful comment. Also, these fixers have been in place for some time without complaint--I'm only now giving the opportunity to disable them. I'm ok with a breaking change to change the default though if you want to err on the side of caution.

golopot:
That would be great, thanks. And it should be a minor release. https://github.com/eslint/eslint#semantic-versioning-policy

brettz9:
Sure, though it breaks in the sense of no longer working quite as before, it at least won't lead to more broken builds.

brettz9:
I've switched to the new default (of `enableFixer: false`).

brettz9:
@golopot : It'd be great if you could find time to take a look to see that my latest changes are ok and if this is ready for merging (I've rebased, but the changes were made in the final commit of this PR). I intend to look at fixing #532 tomorrow.)

PR #532, opened 2020-05-11
`checkDestructured` property to disable destructured checks

### LINKED ISSUE #530: Option to configure `require-param` to work identical to pre - v25
To start off, I've really loved this plugin and especially the `require-param` rule. This has been great for enforcing consistency of docs on our team!

Our codebase is entirely typescript so we set `'jsdoc/require-param-type': 0,` but leave on `jsdoc/require-param` since it will enforce that each function param has an associated comment which is HUGE.

The `v25` major upgrade however caused some problems since every object param we had was automatically expanded to have the `opts.arg1`, `opts.arg2` docs.

We practically never want this expansion to happen because all of these object params have an associated typescript type, and we use tslint `completed-docs` to enforce each param in our typescript types has a comment.

My Ask:

Is there a way that we can configured `require-param` to never do the automatic expansion of object params? I read the changelog and read the updated docs but I could not for the life of me figure out how to configure the eslint rules so that it never expanded, but still auto fixed to add a jsdoc for the arugment if it was missing. The old setup was so perfect and I :pray: that there's a way to configure this to keep fix on but to not warn or fix this object expansion

Bonus Points:

Also trying to move off of tslint and finding a full replacement for the `completed-docs` rule is really hard. We like to write out types like the following:

```ts
/**
 * A container definition
 */
export type Container = {
  /** Constants */
  constants?: ObjByString;
  /** Enumerated values, where the values are CONSTANT_CASE by default */
  enums?: { [key in string]: TypescriptEnum };
  /** Helper functions for that container */
  helpers?: { [key in string]: AnyFunction };
};

```

And the `completed-docs` rule can enforce that each of the type params (i.e. `constants`, `enums`, `helpers` has an inline jsdoc). I could see this being outside the scope of this pacakge, but would be incredible if that capability was done here in eslint and not tslint!

brettz9:
Currently builds on #531

Allows disabling of destructured checks (by setting `checkDestructured` to `false`).

Fixes #530.

### DIFF
--- a/test/rules/assertions/requireParam.js
+++ b/test/rules/assertions/requireParam.js
@@ -87,6 +87,39 @@ export default {
     {
       code: `
           /**
+           * @param foo
+           */
+          function quux (foo, bar, {baz}) {
+
+          }
+      `,
+      errors: [
+        {
+          message: 'Missing JSDoc @param "bar" declaration.',
+        },
+        {
+          message: 'Missing JSDoc @param "root0" declaration.',
+        },
+      ],
+      options: [
+        {
+          checkDestructured: false,
+        },
+      ],
+      output: `
+          /**
+           * @param foo
+           * @param bar
+           * @param root0
+           */
+          function quux (foo, bar, {baz}) {
+
+          }
+      `,
+    },
+    {
+      code: `
+          /**
            *
            */
           function quux ({foo}) {
@@ -2695,6 +2728,23 @@ export default {
       }
       `,
       parser: require.resolve('@typescript-eslint/parser'),
+    },
+    {
+      code: `
+          /**
+           * @param foo
+           * @param bar
+           * @param cfg
+           */
+          function quux (foo, bar, {baz}) {
+
+          }
+      `,
+      options: [
+        {
+          checkDestructured: false,
+        },
+      ],
     },
   ],
 };
--- a/src/rules/requireParam.js
+++ b/src/rules/requireParam.js
@@ -1,3 +1,4 @@
+import _ from 'lodash';
 import iterateJsdoc from '../iterateJsdoc';
 
 type T = [string, () => T];
@@ -47,11 +48,14 @@ export default iterateJsdoc(({
   }
 
   const {
+    autoIncrementBase = 0,
+    checkRestProperty = false,
+    checkDestructured = true,
+    checkTypesPattern = '/^(?:[oO]bject|[aA]rray|PlainObject|Generic(?:Object|Array))$/',
     enableFixer = true,
     enableRootFixer = true,
-    checkRestProperty = false,
     enableRestElementFixer = true,
-    checkTypesPattern = '/^(?:[oO]bject|[aA]rray|PlainObject|Generic(?:Object|Array))$/',
+    unnamedRootBase = ['root'],
   } = context.options[0] || {};
 
   const lastSlashPos = checkTypesPattern.lastIndexOf('/');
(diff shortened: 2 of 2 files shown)


test/rules/assertions/requireParam.js
```diff
@@ -84,6 +84,39 @@ export default {
...
+          function quux (foo, bar, {baz}) {
+
+          }
+      `,
+      errors: [
+        {
+          message: 'Missing JSDoc @param "bar" declaration.',
+        },
+        {
+          message: 'Missing JSDoc @param "root0" declaration.',
```
golopot, line 101: This message is confusing, what is `root0`?

brettz9: This is part of the feature added in #498 --that when a destructured parameter is also missing its root declaration, that by default, the word "root" will be used followed by a 0-based auto-incrementing number for each subsequent root (one can avoid the numeric additions with specific config, or add additional items to an options array so that until all the items in the array are exhausted, it will cycle through the names, e.g., `["root", "cfg", "config"]` will add `@param root` then `@param cfg`, then `@config0`, with the last item auto-incrementing). You could see our discussion about it in the PR for more background.

brettz9: Though I think it makes sense to use `root0` in the fixes, subject to user configuration (disablable by setting `enableRootFixer` to `false`), we could give a message like `"Missing JSDoc @param destructuring root object declaration"`. Does that sound better? And I'd be personally ok having `enableRootFixer` default to `false`, but sometimes I think it is good when the defaults are--within reason--a little over-aggressive since it informs users of the features and if they *don't* want it, they can come looking.

golopot: Missing JSDoc @param declaration for destructured parameter `{bar}`.

brettz9: That could potentially become quite long though, no? e.g., `{bar, baz, foo: {def: {ghi}}, many, others}`

brettz9: I guess we could just slice it to a certain maximum length, adding the curly bracket always at the end.

brettz9: How should we differentiate in this test case though?
```js
      code: `
          /**
           *
           */
          function quux ({foo}) {

          }
      `,
      errors: [
        {
          message: 'Missing JSDoc @param "root0" declaration.',
        },
        {
          message: 'Missing JSDoc @param "root0.foo" declaration.',
        },
      ],
```

Report `Missing JSDoc @param "{foo}" declaration` twice?

brettz9: How should we differentiate in this test case though?
```js
      code: `
          /**
           *
           */
          function quux ({foo}) {

          }
      `,
      errors: [
        {
          message: 'Missing JSDoc @param "root0" declaration.',
        },
        {
          message: 'Missing JSDoc @param "root0.foo" declaration.',
        },
      ],
      output: `
          /**
           * @param root0
           * @param root0.foo
           */
          function quux ({foo}) {

          }
      `,
```

Report `Missing JSDoc @param "{foo}" declaration` twice?

brettz9: We don't want to report just once because the root object may or may not already be defined, and if not defined, there should be a separate message and fix.

brettz9: Btw, if it's ok, can we discuss this as a separate issue for fixing reporting messages? This discussion, though valid, is concerning a preexisting behavior (we refer to `root0` in other messages, not just in this PR). It would be nice to get this merged so the disabling functionality is available sooner.

brettz9: As far as the solution, I guess I can just change it to report:

> Missing JSDoc @param "{foo}" declaration
> Missing JSDoc @param "foo" in "{foo}" declaration

PR #539, opened 2020-05-13
fix(check-tag-names): Workaround for 'constructor' key in tagNamePreference

### LINKED ISSUE #537: Can't set preference for 'constructor' in tagNamePreference due to upstream issue
In the Wikimedia config, we want to allow `@class` and `@constructor`.

This should be possible with the following config:
```json
{
  "tagNamePreference": {
    "class": "class",
    "constructor": "constructor"
  }
}
```

However due to an upstream issue (https://github.com/eslint/eslint/issues/13289) it is impossible to use settings keys that exist on Object.prototype (so `constructor`, `toString` etc.).

edg2s:
Fixes #537

### DIFF
--- a/src/jsdocUtils.js
+++ b/src/jsdocUtils.js
@@ -223,8 +223,15 @@ const getPreferredTagName = (
     return name;
   }
 
-  if (_.has(tagPreference, name)) {
-    return tagPreference[name];
+  // Allow keys to have a 'tag ' prefix to avoid upstream bug in ESLint
+  // that disallows keys that conflict with Object.prototype,
+  // e.g. 'tag constructor' for 'constructor' (#537)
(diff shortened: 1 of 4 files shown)


src/jsdocUtils.js
```diff
@@ -216,20 +216,27 @@ const getPreferredTagName = (
...
+  });
+
+  if (tagPreferenceFixed.includes(name)) {
+    return tagPreferenceFixed[name];
   }
 
   const tagNames = getTagNamesForMode(mode, context);
 
-  const preferredTagName = _.findKey(tagNames, (aliases) => {
+  const preferredTagName = Object.keys(tagNames).find((aliases) => {
```
brettz9, line 239: Sorry, made some premature changes re: further reducing lodash, but working on fixing locally.

brettz9:
Thanks for the issue and PR!

PR #542, opened 2020-05-16
chore: update dep. spdx-expression-parse and devDeps.; apply new linting

brettz9:
@gajus, here's how the linting changes may look...  (Related to https://github.com/gajus/eslint-config-canonical/issues/23 )

Though I re-enabled `reduce` in a couple cases, I think it is mostly more readable. (I conceptually like `reduce` better as it is just one action (e.g., not a `map` and `join`, or a variable declaration and `forEach`), but I'll admit it can *look* less intuitive, so except in the two cases, I think it has indeed become more readable after dropping `reduce`).

### DIFF
--- a/src/rules/requireJsdoc.js
+++ b/src/rules/requireJsdoc.js
@@ -119,20 +119,22 @@ const getOptions = (context) => {
         return false;
       }
 
-      return Object.keys(baseObj.properties).reduce((obj, prop) => {
+      const properties = {};
+      Object.keys(baseObj.properties).forEach((prop) => {
         const opt = getOption(context, baseObj, 'publicOnly', prop);
-        obj[prop] = opt;
+        properties[prop] = opt;
+      });
 
-        return obj;
-      }, {});
+      return properties;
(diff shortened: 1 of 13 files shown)


src/rules/requireJsdoc.js
```diff
@@ -119,20 +119,22 @@ const getOptions = (context) => {
         return false;
       }
 
-      return Object.keys(baseObj.properties).reduce((obj, prop) => {
+      const properties = {};
+      Object.keys(baseObj.properties).forEach((prop) => {
```
gajus, line 123: Any reason for using `forEach` over `for ... of`?

gajus:
LGTM

src/rules/requireJsdoc.js
brettz9: I am just in the habit of using `forEach`, as I like that I can just move the function out of the `forEach` if needed, changing only the reference within `forEach`. It feels more easily composable than `for...of`.

PR #544, opened 2020-05-18
More accurate Configuration section

Twoody:
Users and newcomers should not need to know what file to modify to properly configure the plugin.

This revision allows for no asking if this is package.json or some other file, but is in fact .eslintrc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ npm install --save-dev eslint-plugin-jsdoc
 <a name="eslint-plugin-jsdoc-configuration"></a>
 ## Configuration
 
-Add `plugins` section and specify `eslint-plugin-jsdoc` as a plugin.
+Add `plugins` section to `.eslintrc` and specify `eslint-plugin-jsdoc` as a plugin.
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


brettz9:
Thank you for the PR.

Using an extension is recommended these days, so I'd recommend changing the text to . `.eslintrc.*` with the asterisk. I'd also like to see us refer them by link to https://eslint.org/docs/user-guide/configuring#configuration-file-formats to explain that format (and, btw, one can in fact use `package.json` via `eslintConfig`, but I think providing the link will be sufficient to address all that).

Twoody:
@brettz9 Thanks for the quick reply! Anymore, it feels like I wait weeks to hear back from pr's :D Would you like me to make those modifications, or is this something that maintainers will do? (I.e. what next steps do I need to take?)

brettz9:
My pleasure. Can't always get to them quickly, but hope to. I hear ya on the waiting...

Yes, generally it is preferable if you can save us the time by making the requested modifications. If you add commits and push to this same branch (in your case, the default `master`), they will be added to this PR.

FWIW, I'm personally ok with amended commits (except when there are a lot of large changes) if you prefer to keep the commit history clean (though we may auto-squash the commits together into a single commit ourselves later anyways (though not always)). But the simple way of just adding more commits is fine and often preferable to allow us to only need to review new changes.

brettz9:
Looks good... Thanks for the PR!

PR #548, opened 2020-05-22
`require-description-complete-sentence`: `newlineBeforeCapsAssumesBadSentenceEnd` option

### LINKED ISSUE #427: require-description-complete-sentence heuristic for detecting incomplete sentences via capitalization is broken.
The `require-description-complete-sentence` rule assumes that capital letters always indicate the start of a sentence. This heuristic fails whenever a description contains a capitalized word that appears at the beginning of a line, for example due to line wrapping:

```js
/**
 * Implements support for the
 * Swahili voice synthesizer.
 */
function speak() { ... }
```

This causes a conflict between the linter rule and natural language grammar rules which require certain words to always be capitalized - e.g. above: proper nouns (names of people, countries, languages, nationalities) must always be capitalized in English.

I would suggest to add an option to the `require-description-complete-sentence` rule to selective enable or disable the capital letter heuristic.

brettz9:
Fixes #427.

- feat(`require-description-complete-sentence`): add `newlineBeforeCapsAssumesBadSentenceEnd` option to disable heuristic for caps after newlines being assumed to be a bad sentence end.

I might suggest further changing this to be `false` by default but it would be breaking in the sense of changing behavior for those expecting it to continue reporting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6348,6 +6348,7 @@ function quux () {
 function quux () {
 
 }
+// Options: [{"newlineBeforeCapsAssumesBadSentenceEnd":true}]
 // Message: A line of text is started with an uppercase character, but preceding line does not end the sentence.
 
 /**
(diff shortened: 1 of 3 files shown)


brettz9:
I'm aware of ESLint's own versioning policy, but that doesn't necessarily mean it needs to be ours. And I think it only requires time to learn the option if people are unhappy with the defaults.

I'm personally ok with dropping the behavior, but I tend to like to err on the side of providing too many options rather than dropping access for people who liked the behavior no less since it's already been active and implemented.

For people who are aware enough to avoid capital letters on new lines, the old way may be useful to them. But again, I'm ok dropping it if there are no dissenting voices wishing to keep it.

brettz9:
Per discussion on #427, it seems the consensus is to disable the old behavior by default but make the old behavior available when opting in.

PR #554, opened 2020-05-28
Add documentation for `require-description-complete-sentence`

hbergren:
We were missing explicitly stating `newlineBeforeCapsAssumesBadSentenceEnd` was an option for this rule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6266,12 +6266,18 @@ You can provide an `abbreviations` options array to avoid such strings of text
 being treated as sentence endings when followed by dots. The `.` is not
 necessary at the end of the array items.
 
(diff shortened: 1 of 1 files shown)


brettz9:
Thank you for the PR. I prepared a separate commit c4903c07923a73369c4d1fb975a474688024e2cf (now in `master`) based on your work since we need to modify the file within `.README/rules` which is then used, when `npm run create-readme` is executed, to build the main `README.md`. But this was good to have documented, thank you!

hbergren:
Good to know! I'll keep in mind that the source of truth is `.README/rules` the next time I do something documentation-related.

PR #555, opened 2020-05-29
`require-hyphen-before-param-description`: add `tags` option

### LINKED ISSUE #553: Option to support or ban hyphens before other tag descriptions
## Motivation

We use the [require-hyphen-before-param-description](https://github.com/gajus/eslint-plugin-jsdoc#eslint-plugin-jsdoc-rules-require-hyphen-before-param-description) rule, and in some cases, people get confused and add hyphens for other JSDoc tags. The most common one I see is adding a hyphen to the `@returns` tag.

## Current behavior

No way to discourage writing something like:
```
/**
 * ...
 * @returns - description
 */
```

## Desired behavior

Have some way to discourage using hyphens for JSDoc tags that are not `@param`.

## Alternatives considered

I tried using the `match-description` and setting a regex, but it seemed to ignore the presence/absence of a hyphen.

brettz9:
feat(require-hyphen-before-param-description): add `tags` option for setting specific tags (or any tags) to follow rules for or against hyphen descriptions; now allows property to be set differently than param; fixes #553

BREAKING CHANGE:

`{checkProperties: true}` should be replaced by: `{tags: {'property': 'always|never'}}`

### DIFF
--- a/src/rules/requireHyphenBeforeParamDescription.js
+++ b/src/rules/requireHyphenBeforeParamDescription.js
@@ -5,12 +5,13 @@ export default iterateJsdoc(({
   utils,
   report,
   context,
+  jsdoc,
   jsdocNode,
 }) => {
-  const [circumstance, {checkProperties} = {}] = context.options;
-  const always = !circumstance || circumstance === 'always';
+  const [mainCircumstance, {tags} = {}] = context.options;
 
-  const checkHyphens = (jsdocTag, targetTagName) => {
+  const checkHyphens = (jsdocTag, targetTagName, circumstance = mainCircumstance) => {
+    const always = !circumstance || circumstance === 'always';
     if (!jsdocTag.description) {
       return;
     }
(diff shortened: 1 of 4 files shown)


src/rules/requireHyphenBeforeParamDescription.js
```diff
@@ -47,8 +48,28 @@ export default iterateJsdoc(({
...
-  if (checkProperties) {
-    utils.forEachPreferredTag('property', checkHyphens);
+  if (tags) {
+    const tagEntries = Object.entries(tags);
+    tagEntries.forEach(([tagName, circumstance]) => {
+      if (tagName === '*') {
+        const preferredParamTag = utils.getPreferredTagName({tagName: 'param'});
+        (jsdoc.tags || []).forEach(({tag}) => {
+          if (tag === preferredParamTag || tagEntries.some(([tagNme]) => {
+            return tagNme !== '*' && tagNme === tag;
```
hbergren, line 58: Should this be `tagName` as opposed to `tagNme`?

brettz9: The `tagNme` is just used because earlier in this block, there is a different `tagName` (and there is a linting rule discouraging shadowing, i.e., reusing the same name for a different variable in a more nested scope).

brettz9:
Yeah, I'm ok with a breaking change to change the whole rule name (and then the implementation to fit), but this had been about params, then we tacked on property support, and now want to add it for others.

If we make the rule independent of params, then both points should be clearer. Just the concern that renaming a rule is more consequential in that it causes a single point of failure upon an upgrade (unless one was using ("recommended") config).

brettz9:
So, @golopot (and any others), do you think we should introduce a breaking change and rename the rule or just use the added option?

Note that instead of being `["always", {"*": "never"}]`, it would actually be: `["always", {tags: {"*": "never"}}]` which I think is not quite as contradictory, though I know it would be clearer if `@param` were treated in the same way as other tags, e.g., if one had to do `{tags: {'*', 'never', 'param: 'always'}}`.

hbergren:
> Though I know it would be clearer if @param were treated in the same way as other tags

If we made a breaking change and renamed the rule, we could do something like this:

```js
'jsdoc/require-hyphen-before-tag-description': ['error', {
  '*': 'never', // or maybe 'all'?
  param: 'always',
}]

brettz9:
@hbergren : True, though such approaches can tend to lock one in with an inability to add further options in the future if needed. But I think we can go forward for now without breakage.

PR #575, opened 2020-06-10
Fix bug with destructured parameters in check-param-names

bennieswart:
This PR fixes the false negatives in the check-param-names rule for the following cases:

```js
/**
 * @param cfg
 * @param cfg.foo
 * @param cfg2
 */
function quux ({foo}, cfg2) {

}
// Results in: @param "cfg2" does not exist on cfg

/**
 * @param cfg
 * @param cfg.foo
 * @param baz
 * @param baz.cfg
 */
function quux ({foo}, {cfg}) {

}
// Results in: @param "baz.cfg" does not exist on cfg
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -2132,6 +2132,25 @@ function quux ({foo}, baz) {
 }
 
 /**
+ * @param cfg
+ * @param cfg.foo
+ * @param cfg2
+ */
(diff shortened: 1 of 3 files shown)


brettz9:
Great, thanks a lot!

PR #580, opened 2020-06-16
Add meta.docs.description to every rule

edg2s:


brettz9:
Do you know of a tool that uses this info? It's good to have the meta-data in any case, but just curious if anyone knows... Thanks!

edg2s:
@brettz9 The built in json-with-metadata formatter includes the descriptions (I don't know who uses that), as do readme builders like https://github.com/j-f1/eslint-docs and https://github.com/wikimedia/eslint-docgen.

PR #584, opened 2020-06-18
annotation mismatches

### LINKED ISSUE #491: Support for detecting missing/extraneous object properties in TypeScript
I'm working through updating the JSDoc comments in one of my libraries, and the doc comments are significantly diverged from the code. I was hoping this plugin would help me catch those instances, and while it's incredibly useful, it's missing one use case that'd be a huge help in my codebase.

Given the following TypeScript, I was hoping eslint would report two errors:

```typescript
/**
 * Testing
 *
 * @param options
 * @param options.one One
 * @param options.two Two
 * @param options.four Four
 */
function testingEslint(options: {
  one: string;
  two: string;
  three: string;
}): string {
  return one + two + three;
}
```

Expected errors:

1. Missing `@param` for `options.three`
2. `@param` "options.four" does not match an existing function parameter

I tested by enabling all the available rules, but please let me know if this is doable through some extra configuration.

brettz9:
feat(`check-param-names`): report missing/extraneous TypeScript annotation properties in both directions and report mismatches of parameter names which are parents to annotations; fixes #491

Also provides more precise line number reporting.

In some ways this is more of a fix, but I wanted a regular TS user to confirm that a case like:

```js
        export class SomeClass {
          /**
           * @param prop
           * @param prop.foo
           * @param prop.bar
           */
          constructor(options: { foo: string, bar: string }) {}
        }
```

...should report a mismatch between `prop` and `options` and not just the nested items--as seems to make sense to me.

### DIFF
--- a/test/rules/assertions/checkParamNames.js
+++ b/test/rules/assertions/checkParamNames.js
@@ -580,11 +580,23 @@ export default {
       errors: [
         {
           line: 4,
+          message: '@param "prop" does not match annotation name "options"',
+        },
+        {
+          line: 4,
           message: 'Missing @param "options.foo"',
         },
         {
           line: 4,
           message: 'Missing @param "options.bar"',
+        },
+        {
+          line: 5,
+          message: '@param "prop.foo" does not exist on prop',
+        },
+        {
+          line: 6,
+          message: '@param "prop.bar" does not exist on prop',
         },
       ],
       parser: require.resolve('@typescript-eslint/parser'),
@@ -605,7 +617,7 @@ export default {
       `,
       errors: [
         {
-          line: 4,
+          line: 6,
           message: '@param "options.bar" does not exist on options',
         },
       ],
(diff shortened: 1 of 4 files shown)


dstaley:
@brettz9 yeah, in this particular instance I'd expect an error since accessing `prop` in the body of the function would return an error. If the argument is named (`options` in this case), I'd want the linter to enforce that the same name is used in the JSDoc.

test/rules/assertions/checkParamNames.js
```diff
@@ -578,6 +578,10 @@ export default {
         }
       `,
       errors: [
+        {
+          line: 4,
+          message: '@param "prop" does not match annotation name "options"',
```
golopot, line 583:
```suggestion
          message: '@param "prop" does not match parameter name "options"',
```
"options" is not an annotation name.

brettz9: Thanks--good catch. I've pushed a fix.

PR #595, opened 2020-06-28
`require-jsdoc`: add `enableFixer` option

### LINKED ISSUE #372: Empty docblocks are added above functions on lint
When I run the linter at adds empty docblocks above every function that doesn't already have a docblock. I assume this is part of the auto fixing but this messes doc coverage reports by incorrectly reporting that a function is documented when it is not. 

Is there a way to disable the behavior of auto adding empty doc blocks?

brettz9:
feat(`require-jsdoc`): add `enableFixer` option; if set to `false`, will avoid fixer; fixes #372

As per discussion in #372, let me know if you want me to change this to make a break from previous behavior by instead making `enableFixer` off by default. (If so, I think `enableFixer` should be added as `true` on most existing tests so that the docs continue to make clear how/where the fix will appear.)

Also, @gajus, @golopot seems to favor [ESLint's approach](https://github.com/eslint/eslint#semantic-versioning-policy) of only treating something as breaking if it is likely to increase the number of errors. I personally favor an approach which I think is more in line with the spirit of semver by requiring a breaking change if there will be fewer errors (and there was legitimate reason to see them previously as errors and those errors were not the result of a mere bug). If you want to weigh in on this, I think it will save us from some future back-and-forth.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8367,6 +8367,12 @@ function myFunction() {
 // Settings: {"jsdoc":{"minLines":1}}
 // Message: Missing JSDoc comment.
 
+function myFunction() {
+
+}
+// Options: [{"enableFixer":false}]
+// Message: Missing JSDoc comment.
+
 export var test = function () {
 
 };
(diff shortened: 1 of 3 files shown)


brettz9:
I guess I can merge as it is now and make it a breaking change later if you decide you want it as a breaking change.

Issue #598, opened 2020-07-01
Does not support destructuring in function arguments if the key in a destructured object is a string


michalczaplinski:
## Expected behavior
The parser should be able to handle destructuring with renaming in function parameters when the key is a string.

## Actual behavior
Eslint throws the `Unsupported function signature format` error and stops linting: 
```
[Error - 11:50:20 AM] ESLint stack trace:
[Error - 11:50:20 AM] Error: Unsupported function signature format.
Occurred while linting /Users/czapla/Testing/frontity/packages/components/link.tsx:71
    at getParamName (/Users/czapla/Testing/frontity/node_modules/eslint-plugin-jsdoc/dist/jsdocUtils.js:178:11)
```

## The offending code

```js
const someFunction = ({ 
  "value": renamedValue,
  otherArg,
}) => {
  console.log("not important");
}
```

If I comment out the line with `"value": renamedValue`, the parser continues to work. This is a rare use case, because normally you can omit the quotes, but becomes necessary if the key includes a hyphen like `{ "renamed-value": renamedValue }`

## Environment

* Node version: `v14.2.0`
* ESLint version `eslint@6.8.0`
* `eslint-plugin-jsdoc` version: `27.1.1`

PR #599, opened 2020-07-01
fix: Object destructuring in function parameters if key is string

michalczaplinski:
Fixes: #598 

A key in a destructured object that is passed as a function parameter could be a string or a number. This case was previously not handled by `getParamName()`.

Example that failed before and succeeds now:
```js
const someFunction = ({ 
  "value": renamedValue,  // eslint throws Unsupported function signature format.
  otherArg,
}) => {
  console.log("not important");
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -2103,7 +2103,16 @@ function quux ({"a": A, b}) {
 
 /**
  * @param foo
- * @param foo.a-b
+ * @param foo."a"
+ * @param foo.b
+ */
+function quux ({a: A, b}) {
+
+}
+
+/**
+ * @param foo
+ * @param foo."a-b"
  * @param foo.b
  */
 function quux ({"a-b": A, b}) {
--- a/README.md
+++ b/README.md
@@ -2094,6 +2094,24 @@ function quux ({a, b}) {
 
 /**
  * @param foo
+ * @param foo.a
+ * @param foo.b
+ */
+function quux ({"a": A, b}) {
+
+}
+
+/**
+ * @param foo
+ * @param foo.a-b
+ * @param foo.b
+ */
+function quux ({"a-b": A, b}) {
+
+}
+
+/**
+ * @param foo
  * @param foo.bar
  * @param foo.baz
  * @param bar
--- a/test/rules/assertions/checkParamNames.js
+++ b/test/rules/assertions/checkParamNames.js
@@ -40,7 +40,7 @@ export default {
       settings: {
         jsdoc: {
           tagNamePreference: {
-            param: 'arg',
+            param: "arg",
           },
         },
       },
@@ -73,7 +73,8 @@ export default {
       errors: [
         {
           line: 3,
-          message: '@param path declaration ("Foo.Bar") appears before any real parameter.',
+          message:
+            '@param path declaration ("Foo.Bar") appears before any real parameter.',
         },
       ],
     },
@@ -90,7 +91,8 @@ export default {
       errors: [
         {
           line: 4,
-          message: '@param path declaration ("Foo.Bar") root node name ("Foo") does not match previous real parameter name ("foo").',
+          message:
+            '@param path declaration ("Foo.Bar") root node name ("Foo") does not match previous real parameter name ("foo").',
         },
       ],
     },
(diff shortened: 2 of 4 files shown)


test/rules/assertions/checkParamNames.js
```diff
@@ -925,6 +925,18 @@ export default {
...
+          /**
+           * @param foo
+           * @param foo.a
+           * @param foo.b
+           */
+          function quux ({"a": A, b}) {
+
+          }
+      `,
+    },
```
brettz9, line 939: This looks good.  But could you add an example with a hyphen? I think that will make it more clear when the tests get added to the README (and you can do that in this PR by running `npm run create-readme` after you finish updating tests, btw) why one might wish to use a literal. Also, if numbers would work here, it'd be nice I think to have such an example as well.

README.md
```diff
@@ -2086,6 +2086,24 @@ function quux ({a, b}) {
...
+ * @param foo.a
+ * @param foo.b
+ */
+function quux ({"a": A, b}) {
+
+}
+
+/**
+ * @param foo
+ * @param foo.a-b
```
michalczaplinski, line 2100: I think we have a slight problem here. I wanted to add this as `@param foo["a-b"]` because this is the valid JS syntax, but then the tests would not pass. `foo.a-b` is syntactically valid but not semantically (cause it's a subtraction operation 😅 )

michalczaplinski:
Unrelated to the current PR: I've had to push with `git push --no-verify` because otherwise eslint would yell at me about errors in unrelated files like:

```

/Users/czapla/Testing/eslint-plugin-jsdoc/src/iterateJsdoc.js
  44:23  error  There should be a linebreak after this element  array-element-newline
  44:33  error  There should be a linebreak after this element  array-element-newline
  44:44  error  There should be a linebreak after this element  array-element-newline
  44:54  error  There should be a linebreak after this element  array-element-newline
  45:22  error  There should be a linebreak after this element  array-element-newline
  45:33  error  There should be a linebreak after this element  array-element-newline
  45:42  error  There should be a linebreak after this element  array-element-newline
  45:53  error  There should be a linebreak after this element  array-element-newline
```

Are you aware of that?

brettz9:
To just respond on the linting issue for now...

Ah yes, wasn't seeing this until `npm update --depth 20`, updating my `package-lock.json` copy.

@gajus:

1. Can we add `package-lock.json` to the repo (ignoring for npm releases) so devs can be on the same page? It only causes lock-in for local `npm install` use.
2. IMV, the `array-element-newline` rule can be pretty oppressive, e.g., I hardly find any problem with the following, yet would find adding line breaks to each line to be very bloated (and stuffing it all on one line to be annoying for those of us avoiding wrapping as a default on our IDEs):

```js
          if ([
            'example', 'return', 'returns', 'throws', 'exception',
            'access', 'version', 'since', 'license', 'author',
            'default', 'defaultvalue',
          ].includes(data.tag)) {
```

README.md
brettz9: How about `foo."a-b"`? I think that is the jsdoc namepath way and seems to work in https://jsdoctypeparser.github.io/ ...

test/rules/assertions/checkParamNames.js
```diff
@@ -540,9 +544,9 @@ export default {
           message: 'Expected @param names to be "property". Got "prop".',
         },
       ],
-      parser: require.resolve('@typescript-eslint/parser'),
+      parser: require.resolve("@typescript-eslint/parser"),
       parserOptions: {
-        sourceType: 'module',
+        sourceType: "module",
       },
```
brettz9, line 550: What's with all the double-quotes? Is that a new ESLint rule? I hope not...

michalczaplinski:
@brettz9 Hey, sorry about the delay - the quotes I think were a quirk of default formatting with `prettier` in vscode, my bad!

Thanks for merging in any case! 🙂

brettz9:
@michalczaplinski : Re: quotes, no worries! I apologize I only later edited in a smile to suggest I wasn't complaining at you. But I'll admit I was nervous there was some new config that was imposing the abomination that is double quotes. :)  And no worries re: the delay either. Thanks for the helpful fix!

PR #609, opened 2020-07-11
Report more precisely when `descriptionStyle: 'body'` and only `@desc`/`@description` present

### LINKED ISSUE #608: require-description with descriptionStyle: body produces "Missing JSDoc block description" when @description is used
## Expected behavior

When I set the `descriptionStyle: 'body'` option on `jsdoc/require-description`, and my code uses `@description` instead of omitting it, I expect to get a message that indicates I should not use that tag.

## Actual behavior

Instead the error says "Missing JSDoc block description", which I find misleading and unhelpful.

## ESLint Config

```js
module.exports = {
  extends: ['plugin:jsdoc/recommended'],
  rules: {
    'jsdoc/require-description': ['error', {descriptionStyle: 'body'}],
  },
};
```

## ESLint sample

```js
/** @description Foo! */
function foo() {}
```

## Environment
* Node version:  v10.21.0
* ESLint version  v7.4.0
* `eslint-plugin-jsdoc` version: v29.1.3

brettz9:
feat(`require-description`): report more precisely the action to take when "body" `descriptionStyle` is set and where user (only) has a desc/description tag; fixes #608

### DIFF
--- a/README.md
+++ b/README.md
@@ -7276,6 +7276,24 @@ function quux () {
 // Message: Missing JSDoc block description.
 
 /**
+ * @desc Not a blank description
+ */
+function quux () {
+
+}
+// Options: [{"descriptionStyle":"body"}]
+// Message: Remove the @desc tag to leave a plain block description or add additional description text above the @desc line.
+
+/**
+ * @description Not a blank description
+ */
+function quux () {
+
+}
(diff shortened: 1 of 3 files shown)


joeyparrish:
Thank you so much!

PR #610, opened 2020-07-15
feat: `structuredTags` setting to control type/name portions of existing or custom tags

brettz9:
feat(`check-types`, `no-undefined-types`, `valid-types`): Add `structuredTags` setting to control whether the type and namepath portions should be checked for validity and whether such portions are required, and to let user-defined "namepath-defining" tags be added to defined types. Closes #481

BREAKING CHANGE:

Drops `checkSeesForNamepaths` setting. Use `{settings: {jsdoc: {structuredTags: {name: 'namepath', type: false, required: ['name'],}}}}` instead.

Also:
1. Clarifies in more cases where a problem is specific to the mode or not
2. Reports simultaneous invalid name *and* type errors
3. `typdef` now requires `allowEmptyNamepaths: false,` to report empty names (as with other tags)
4. Requires a name for `event` and `external` (and `extends` in jsdoc mode); some tweaking of other tags per docs

(Also does some jsdocUtil refactoring along the way to pull tag meta-data all into one map for easier management and to minimize tag-specific handling blocks within `valid-types`.)

(Since settings can now throw, needed to add a test to a few other rules to cover such cases.)

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -380,6 +380,49 @@ key nor the value will be reported. Thus in `check-types`, this fact can
 be used to allow both `object` and `Object` if one has a `preferredTypes`
 key `object: 'Object'` and `Object: 'object'`.
 
+### `structuredTags`
+
+An object indicating tags whose types and names/namepaths (whether defining or
+referencing namepaths) will be checked, subject to configuration. If the tags
+have predefined behavior or `allowEmptyNamepaths` behavior, this option will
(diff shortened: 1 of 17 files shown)


brettz9:
Unless there are objections, I'm thinking of also changing this PR to make `allowEmptyNamepaths` default to `false`. I think it is better for the default to be stricter for tags which are documented as expecting them.

PR #620, opened 2020-08-04
fix(`check-examples`): disable jsdoc/require-jsdoc

remcohaszing:
Examples inside JSDoc comments cant comply with the `jsdoc/require-jsdoc` rule, because the closing comment of the nested JSDoc would close the multiline comment that contains the example.

### DIFF
--- a/src/rules/checkExamples.js
+++ b/src/rules/checkExamples.js
@@ -94,6 +94,9 @@ export default iterateJsdoc(({
     // Snippets likely too short to always include import/export info
     'import/unambiguous': 0,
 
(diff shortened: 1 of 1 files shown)


brettz9:
Thanks for this. I would guess no other jsdoc rules should apply either since other rules have to have a `/**` and `*/` to find a block on which to operate.

We do have a [PR](https://github.com/syavorsky/comment-parser/pull/89) to support the opening of jsdoc block sequences in `comment-parser`, but as you point out, a closing sequence wouldn't be possible (unless we were to bless some escaping mechanism).

PR #621, opened 2020-08-04
docs: add meta docs url for all rules

remcohaszing:
This URL is used to create links in some various places. For example VS Code makes the rule name clickable in the warnings panel if the URL is available.


brettz9:
Thank you for offering this, but I think we'd need to point to the anchors on the main README instead, as those files are the source files which need to go through further processing to be complete.

Note that #472 proposing to split up the output docs (as with the source docs) was rejected.

Would be happy to merge if you would change these to point to anchors on https://github.com/gajus/eslint-plugin-jsdoc/ .

PR #627, opened 2020-08-24
fix(`no-bad-blocks`): allow TypeScript directive comments

remcohaszing:
This allows TypeScript directive comments in block comments.

Although these comments are mostly written using line comments, block comments must be used in some situations, e.g. when inside JSX tags.

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -490,6 +490,8 @@ selector).
 {"gitdown": "include", "file": "./rules/implements-on-classes.md"}
 {"gitdown": "include", "file": "./rules/match-description.md"}
 {"gitdown": "include", "file": "./rules/newline-after-description.md"}
+{"gitdown": "include", "file": "./rules/no-bad-blocks.md"}
+{"gitdown": "include", "file": "./rules/no-defaults.md"}
 {"gitdown": "include", "file": "./rules/no-types.md"}
 {"gitdown": "include", "file": "./rules/no-undefined-types.md"}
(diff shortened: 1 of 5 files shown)


brettz9:
Ah, looks like a nice addition. Do you think you could also add an array option (with test coverage) to allow overriding the defaults with one's own list? (Our convention has been to override defaults rather than supplementing them.)

remcohaszing:
Sure, I’ll get to it later today. I’ll add a property `ignored`, which is an array that accepts ignored tag names.

What should the behaviour be for defaults?

1. Don’t add any defaults
2. Default to TypeScript comments
    1. Specifying `ignore` overrides the defaults.
    2. Specifying `ignore` appends the defaults.

brettz9:
Sounds good... I think we can go with 2.i (the project owner and other maintainer wanted this kind of behavior for options in place of the likes of 2.ii).

If you prefer, we could also check whether `settings.jsdoc.mode` is "typescript", and if so, then add the defaults and otherwise, not. But either way is ok with me.

remcohaszing:
I made it customizeable as per option in 2.i. I didn’t check for `settings.jsdoc.mode`, because I think it might be confusing the mode changes the defaults for this rule, and I don’t think anyone would use TS directive comments for anything other than TypeScript.

brettz9:
Looks great.

I just changed to expect no `@` along with the directive name (since I think the at-sign is more of an implementation detail in addition to being always present and thus redundant).

I've also added some docs for it and ensuring `no-bad-blocks` is in the README (as well as `no-defaults` which was also missing).

Issue #540, opened 2020-05-14
Destructuring improvements/fixes


brettz9:
A couple issues on the new destructuring support:

- [ ] Where a root  and grandchildren are present but not children, the fixing order is off.
- [ ] As per https://github.com/gajus/eslint-plugin-jsdoc/pull/532#pullrequestreview-409240327 , we'd like to have clearer messages for reporting destructuring issues. To do this, we need to have a proper function for stringifying a destructured parameter object.
- [ ] Report the proper line numbers

I put the issues together since for implementation, I think it will be easiest to refactor `getFunctionParameterNames` to handle both cases (perhaps avoiding flattening--need to investigate).

PR #630, opened 2020-09-05
fix(`require-param`): check deeply destructured parameters (fixes #569)

### LINKED ISSUE #569: require-param rule destructuring false errors
The `require-param` rule option `checkTypesPattern` currently doesn't work deeply for destructured parameters.

## Expected behavior

If you use `object` for an `@param` root type but then a named type (e.g. `FooBar`) for one of it's destructured properties, there should be no `jsdoc/require-param` lint errors.

## Actual behavior

There are unexpected lint errors.

## ESLint Config

```json
{
  "rules": {
    "jsdoc/require-param": "error"
  }
}
```

## ESLint sample

This has the unexpected lint error `Missing JSDoc @param "options.a.b" declaration. eslint(jsdoc/require-param)`:

```js
/**
 * Description.
 * @param {object} options Options.
 * @param {FooBar} options.a A description.
 */
function foo({ a: { b } }) {}
```

This does not:

```js
/**
 * Description.
 * @param {Foo} options Options.
 * @param {FooBar} options.a A description.
 */
function foo({ a: { b } }) {}
```

## Environment

* Node version: v14.4.0
* ESLint version v7.2.0
* `eslint-plugin-jsdoc` version: v27.0.4

hyex:
fixes [#569]([https://github.com/gajus/eslint-plugin-jsdoc/issues/569](https://github.com/gajus/eslint-plugin-jsdoc/issues/569))

Check the type of destructured parameters in the require-param rule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12169,6 +12169,20 @@ function quux ({foo: bar}) {
 module.exports = function a(b) {
   console.info(b);
 };
+
+/**
+ * Description.
+ * @param {object} options Options.
+ * @param {FooBar} options.foo A description.
+ */
+function quux ({ foo: { bar } }) {}
+
+/**
+ * Description.
+ * @param {Foo} options Options.
+ * @param {FooBar} options.a A description.
+ */
+function quux ({ foo: { bar } }) {}
 ````
 
 
--- a/README.md
+++ b/README.md
@@ -11595,6 +11595,14 @@ module.exports = class GraphQL {
 	}
 })();
 // Message: Missing JSDoc @param "param" declaration.
+
+/**
+ * Description.
+ * @param {Object} options Options.
+ * @param {Object} options.foo A description.
+ */
+function quux ({ foo: { bar } }) {}
+// Message: Missing JSDoc @param "options.foo.bar" declaration.
 ````
 
 The following patterns are not considered problems:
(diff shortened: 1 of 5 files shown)


brettz9:
Looks great from a quick look, thank you very much!

In case it will all need to be redone though, I'm a little reluctant to add much more code on top of our code until the first part of #540 might be addressed (i.e., "Where a root and grandchildren are present but not children, the fixing order is off."). Do you think this is something you could also look into?

If not, I can still see about merging as it of course is better to have something working.

hyex:
Thank you for your quick response!

Actually, It's my first PR in open source. So #540 seems like a difficult issue for me to solve for now... Can I try it when I have enough time?

brettz9:
Ok--but as we also have an `checkTypesPattern` option for `check-param-names`, could you at least see if that needs an equivalent fix (and ideally make the fix there also)?

hyex:
Of course! 
I also agree that I need to fix `check-param-names`. I will ping you after finishing the fix.

hyex:
@brettz9 
Hi, I fixed what you mentioned. 
Please let me know if there are need any changes. I will be appreciated any review. 😊

brettz9:
Great work, thank you!!

PR #631, opened 2020-09-05
Fixed typo in check-types.md

bobpaw:
It was a bad typo since the whole rule is about checking case.


brettz9:
Yeah, I think the source of it is that if I weren't blocking `function` in my own projects (in favor of always requiring a specific callback type), I'd prefer for `function` to be lower-case  (with the similar philosophy to this rule, with the difference that besides for the obvious `null` distinction, `typedef` should be used for identifying precise types, unless, like `Array`, `Date`, they have a distinct (non-`Object`) constructor. But yeah, it should be advertising how it's behaving of course... Thanks!

Issue #483, opened 2020-01-17
Implement default "never" option for `check-line-alignment` rule: linting inside comments block spaces/tab to single space


jperocho:
Just a question if this plugin has a rule to check for multiple spaces or tabs inside a comment block?

```
 /**
 * Some jsdocs for a function
 * @param   {String}   argument string - string
 * @param   {Object}   property        - property object
 * @returns {Function} callback        - callback
 */
```

Thanks in advance.

brettz9:
Not currently. One can use regexes on the description portion using `match-description`, but that wouldn't catch whitespace before or after the tag, type, or name (or default value).

l1bbcsg:
This is quite complex given the lack of concrete syntax tree or parsers that can produce one for JSDoc.
Regular expressions will quickly turn into a mess if edge cases and more scenarios were to be implemented.

Neither Doctrine, nor comment-parser can do what espee does for js, namely tokenize input and keep information about all whitespaces and non-syntax important stuff.
When we previously discussed this in my company, we figured we're going to need to write a new JSDoc parser that either keeps track of tokens or builds a concrete syntax tree, but it wasn't the task we had time for.
This would allow all kinds of validation of JSDoc codestyle – indentation, whitespace placement, you name it.

Some attempt were made in [eslint-plugin-interfaced](https://github.com/interfaced/eslint-plugin-interfaced) however, particularly [JSDocTokenizer](https://github.com/interfaced/eslint-plugin-interfaced/blob/master/lib/jsdoc-tokenizer.js) and some [helper functions](https://github.com/interfaced/eslint-plugin-interfaced/blob/master/lib/ast-utils.js). And [`jsdoc-type-spacing`](https://github.com/interfaced/eslint-plugin-interfaced/blob/master/docs/rules/jsdoc-type-spacing.md) is an example of a rule that validates whitespaces.

yvele:
What about the use case where we want to **disable multi space alignment** in JSDoc? Does that feature requires complex parsing?

Something that fixes:

```js
/**
 * @param   {string}   foo   This is foo.
 * @param   {object}   bar   This is bar.
 * @returns {boolean}        This is the returned value.
 */
```

to:

```js
/**
 * @param {string} foo This is foo.
 * @param {object} bar This is bar.
 * @returns {boolean} This is the returned value.
 */
```

The rule may be very specific like

```json
"jsdoc/no-aligment" : "error"
```

[...]

PR #636, opened 2020-09-11
Add lines alignment rule

renatho:
## The PR

When developing for WordPress, there is a [recommended standard](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/javascript/#aligning-comments) to have the comment lines aligned.

Some days ago, we created an Eslint rule for that: https://github.com/Automattic/eslint-plugin-jsdoc-alignment/
It was created with the purpose to start using it and soon submit a contribution to add it to the `eslint-plugin-jsdoc`, which makes much more sense.

If this PR is approved, the current `eslint-plugin-jsdoc-alignment` will be deprecated in favor of this rule.

It's my first contribution to `eslint-plugin-jsdoc`, so I'm sorry if I missed something. But I'll be happy to fix anything if needed.

## Example

❌ Incorrect example:

```js
/**
 * Function description.
 *
 * @param {int} sit Description multi words.
 */
```

✅ Correct example:

```js
/**
 * Function description.
 *
 */
```

### DIFF
--- a/.README/rules/check-lines-alignment.md
+++ b/.README/rules/check-lines-alignment.md
@@ -1,6 +1,7 @@
 ### `check-alignment`
 
-Reports invalid alignment of JSDoc block lines.
+Reports invalid alignment of JSDoc block lines. This is a
+[standard recommended to WordPress code](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/javascript/#aligning-comments), for example.
 
 |||
 |---|---|
--- a/.README/rules/check-lines-alignment.md
+++ b/.README/rules/check-lines-alignment.md
@@ -1,0 +1,51 @@
+### `check-alignment`
+
+Reports invalid alignment of JSDoc block lines.
+
+|||
+|---|---|
+|Context|everywhere|
+|Tags|`param`, `arg`, `argument`, `property`, `prop`|
+
+The following patterns are considered problems:
+
+````js
+/**
+ * Function description.
+ *
+ * @param {string} lorem Description.
+ * @param {int} sit Description multi words.
+ */
+const fn = ( lorem, sit ) => {}
+// Message: Expected JSDoc block lines to be aligned.
+
+/**
+ * My object.
+ *
+ * @typedef {Object} MyObject
+ *
+ * @property {string} lorem Description.
+ * @property {int} sit Description multi words.
+ */
+// Message: Expected JSDoc block lines to be aligned.
+
+The following patterns are not considered problems:
+
+````js
+/**
+ * Function description.
+ *
+ * @param {string} lorem Description.
+ * @param {int}    sit   Description multi words.
+ */
+const fn = ( lorem, sit ) => {}
+
+/**
+ * My object.
+ *
+ * @typedef {Object} MyObject
+ *
+ * @property {string} lorem Description.
+ * @property {int}    sit   Description multi words.
+ */
+````
--- a/.README/rules/check-lines-alignment.md
+++ b/.README/rules/check-lines-alignment.md
@@ -3,9 +3,17 @@
 Reports invalid alignment of JSDoc block lines. This is a
 [standard recommended to WordPress code](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/javascript/#aligning-comments), for example.
 
+#### Options
+
+This rule allows one optional string argument. If it is `"always"` then a
+problem is raised when the lines are not aligned. If it is `"never"` then
+a problem should be raised when there is more than one space between the
+lines parts. Only the non-default `"always"` is implemented for now.
+
 |||
 |---|---|
 |Context|everywhere|
+|Options|(a string matching `"always"|"never"`)|
 |Tags|`param`, `arg`, `argument`, `property`, `prop`|
 
 The following patterns are considered problems:
@@ -18,6 +26,7 @@ The following patterns are considered problems:
  * @param {int} sit Description multi words.
  */
 const fn = ( lorem, sit ) => {}
+// Options: ["always"]
 // Message: Expected JSDoc block lines to be aligned.
 
 /**
@@ -28,6 +37,7 @@ const fn = ( lorem, sit ) => {}
  * @property {string} lorem Description.
  * @property {int} sit Description multi words.
  */
+// Options: ["always"]
 // Message: Expected JSDoc block lines to be aligned.
 
 The following patterns are not considered problems:
@@ -40,6 +50,7 @@ The following patterns are not considered problems:
  * @param {int}    sit   Description multi words.
  */
 const fn = ( lorem, sit ) => {}
+// Options: ["always"]
 
 /**
  * My object.
@@ -49,4 +60,5 @@ const fn = ( lorem, sit ) => {}
  * @property {string} lorem Description.
  * @property {int}    sit   Description multi words.
  */
+// Options: ["always"]
 ````
--- a/.README/rules/check-lines-alignment.md
+++ b/.README/rules/check-lines-alignment.md
@@ -1,4 +1,4 @@
-### `check-alignment`
+### `check-lines-alignment`
 
 Reports invalid alignment of JSDoc block lines. This is a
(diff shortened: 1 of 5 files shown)


src/rules/checkLinesAlignment.js
```diff
@@ -0,0 +1,203 @@
...
+const createFixer = (comment, expectedPositions, partsMatrix, lineRegExp, tagIndentation) => {
+  return (fixer) => {
+    let lineIndex = 0;
+
+    // Replace every line with the correct spacings.
+    const fixed = comment.value.replace(lineRegExp, () => {
+      // eslint-disable-next-line unicorn/no-reduce
+      return partsMatrix[lineIndex++].reduce(
+        (acc, {string}, index) => {
+          const spacings = createSequence(' ', expectedPositions[index] - acc.length);
```
brettz9, line 102:
```suggestion
          const spacings = ''.padStart(expectedPositions[index] - acc.length, ' ');
```

I suggest using that instead.  `createSequence` could then be removed.

src/rules/checkLinesAlignment.js
```diff
@@ -0,0 +1,203 @@
...
+ * @param {string} char   The char.
+ * @param {int}    length The length.
+ */
+const createSequence = (char, length) => {
+  return new Array(length + 1).join(char);
+};
+
+/**
+ * Aux method until we consider the dev envs support `String.prototype.matchAll` (Node 12+).
+ *
```
brettz9, line 18: Though the rule unfortunately does not yet support jsdoc-style `@todo`, Unicorn has an excellent rule we are using, [`expiring-todo-comments`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/expiring-todo-comments.md) which can turn the to-do comment into an error as soon as a condition is matched--in this case when `package.json` starts pointing to Node 12 as a minimum.

You can use this by putting the following above this block instead.

```js
// todo[engine:node@>=12]: Replace with `String.prototype.matchAll`.
```

This has the advantage of ensuring that we will be reminded to make the change as soon as we can.

.README/rules/check-lines-alignment.md
```diff
@@ -0,0 +1,52 @@
+### `check-lines-alignment`
+
+Reports invalid alignment of JSDoc block lines. This is a
+[standard recommended to WordPress code](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/javascript/#aligning-comments), for example.
+
```
brettz9, line 5: As per #483 , it seems the community is more in favor of a rule to prevent this kind of spacing, so while I support having the functionality you are providing as an option, and giving the rationale related to WordPress as you have, I'd like to change the configuration a little.

In case you don't feel you have the time to add the opposite functionality (i.e., preventing extra spaces for alignment), I think we should at least design the rule to accept a string option, "never" or "always", where "never" is the default, but where we throw an exception for the time being if the option is "never" indicating that the default is not yet implemented. This can avoid a future breaking change for those who wish to use the rule now.

Until such a "never" option may be implemented, of course, we wouldn't want this rule added to our `recommended` config, but we can add it as an opt-in rule, with docs indicating that only the non-default "always" option is currently implemented.

brettz9 (changes requested):
Besides my comments, this looks good to me from an initial look.

It would be ideal if we could in the future instead use `comment-parser` for obtaining the tag, name, and description with whitespace, but `comment-parser` doesn't currently support that, so regexes should work for now.

If you could make the changes per the comments, I can see about taking a closer look locally. Thanks a lot!

src/rules/checkLinesAlignment.js
renatho: Good idea! Refactor here: https://github.com/gajus/eslint-plugin-jsdoc/pull/636/commits/98ae62cdb48facd14c04f9edc9acff860dc90ebd

src/rules/checkLinesAlignment.js
renatho: Pretty interesting! TIL =)
Comment added here: https://github.com/gajus/eslint-plugin-jsdoc/pull/636/commits/0be5a14e68026839e270d6ac6fd8df60fa83b274

.README/rules/check-lines-alignment.md
renatho: Makes sense! I implemented the option here: https://github.com/gajus/eslint-plugin-jsdoc/pull/636/commits/07e93f7ddb9a9b52d1df92e88dc6d5ca42987b39

I pushed with `--no-verify` for now, so you can see the changes. But I still need to test the throw to achieve the full coverage. I'll think of an approach to test it and I'll amend the commit.

renatho:
@brettz9, thank you for your first review! I still need to change a little my last commit to having the full coverage, but I already pushed, so you can see the other changes for now! 😉

src/index.js
```diff
@@ -47,6 +48,7 @@ export default {
         'jsdoc/check-alignment': 'warn',
         'jsdoc/check-examples': 'off',
         'jsdoc/check-indentation': 'off',
+        'jsdoc/check-lines-alignment': 'warn',
```
brettz9, line 51: This should be "off" for now (noting that when enabled, it will be for the "never" option).

src/rules/checkLinesAlignment.js
```diff
@@ -0,0 +1,210 @@
...
+ * @param {string}   string   String that will be checked.
+ * @param {RegExp}   regexp   Regular expression to run.
+ * @param {Function} callback Function to be called each iteration.
+ * @param {int}      limit    Limit of matches that we want to exec.
+ *
+ * TODO [engine:node@>=12]: Remove function and use `String.prototype.matchAll` instead.
```
brettz9, line 14: My apologies--I was mistaken about `@todo` not being supported. So please do change `TODO` to `@todo`.

brettz9 (changes requested):
One important and one minor change, though both small and easy-to-change.

All the other recent changes look great.

src/rules/checkLinesAlignment.js
```diff
@@ -0,0 +1,210 @@
...
+const createFixer = (comment, expectedPositions, partsMatrix, lineRegExp, tagIndentation) => {
+  return (fixer) => {
+    let lineIndex = 0;
+
+    // Replace every line with the correct spacings.
+    const fixed = comment.value.replace(lineRegExp, () => {
+      // eslint-disable-next-line unicorn/no-reduce
+      return partsMatrix[lineIndex++].reduce(
+        (acc, {string}, index) => {
+          const spacings = ''.padStart(expectedPositions[index] - acc.length, ' ');
```
brettz9, line 94:
```suggestion
          const spacings = ' '.repeat(expectedPositions[index] - acc.length);
```

Whoops--sorry--I had forgotten about `repeat`. :-)

brettz9 (changes requested):
And this one too if you wouldn't mind--forgot there was this...

src/rules/checkLinesAlignment.js
renatho: Hum! Much better! \o/

src/index.js
renatho: Oops! Missed that. Updated here: https://github.com/gajus/eslint-plugin-jsdoc/pull/636/commits/1c1dd685b5a16a8f3b5660b68447502500040c9a

src/rules/checkLinesAlignment.js
renatho: No worries! Thank you for check that! =)
https://github.com/gajus/eslint-plugin-jsdoc/pull/636/commits/af433fa79ec5197a76e5707784e528f57b91b8be

renatho:
@brettz9 I updated the `never` temporary warning to be a report instead of a `throw`. Do you think it makes sense? At some point in the future, I'll try to implement the `never` too. It's easier than the current check. 🙂

brettz9:
Yes, you are correct, it should have been `report`.

And my apologies, I've just now quickly renamed the rule to `check-line-alignment`, as a "fix" (`line` being a generic adjective here). Thought I had merged that version, but hadn't pushed to the branch. I also included the docs in the main README template, removing the need for including the examples (since we auto-generate those from tests).

PR #638, opened 2020-09-16
Fix check line alignment

renatho:
Some fixes for the rule `check-line-alignment`.

It skips single line comments and considers types with spaces.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1847,6 +1847,19 @@ const config = {
 // Message: Expected JSDoc block lines to be aligned.
 
 /**
+ * My object.
+ *
+ * @typedef {Object} MyObject
+ *
+ * @property {{a: number, b: string, c}} lorem Description.
+ * @property {Object.<string, Class>} sit Description multi words.
+ * @property {Object.<string, Class>} amet Description} weird {multi} {{words}}.
+ * @property {Object.<string, Class>} dolor
+ */
+// Options: ["always"]
+// Message: Expected JSDoc block lines to be aligned.
+
+/**
  * Not implemented yet.
  *
  * @param {string} lorem Description.
@@ -1961,6 +1974,22 @@ const config = {
  * @property {string} lorem Description.
  * @property {int}    sit   Description multi words.
  */
+// Options: ["always"]
+
+/**
+ * My object.
+ *
+ * @typedef {Object} MyObject
+ *
+ * @property {{a: number, b: string, c}} lorem Description.
+ * @property {Object.<string, Class>}    sit   Description multi words.
+ * @property {Object.<string, Class>}    amet  Description} weird {multi} {{words}}.
+ * @property {Object.<string, Class>}    dolor
+ */
+// Options: ["always"]
+
+/** @param {number} lorem */
+const fn = ( lorem ) => {}
 // Options: ["always"]
 
 /**
--- a/src/rules/checkLineAlignment.js
+++ b/src/rules/checkLineAlignment.js
@@ -187,6 +187,11 @@ export default iterateJsdoc(({
     return;
   }
 
+  // Skip if it contains only a single line.
+  if (!jsdocNode.value.includes('\n')) {
+    return;
+  }
+
   // `indent` is whitespace from line 1 (`/**`), so slice and account for "/".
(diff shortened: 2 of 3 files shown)


brettz9:
I can take a look at this, but I'm wondering if you might be open to adding support for an option to preserve whitespace in [comment-parser](https://github.com/syavorsky/comment-parser)--which is a small library that we are elsewhere using in our codebase and has been open to PRs?

I'd feel a lot more comfortable if the rule were to total up whitespace counts based on parsed results rather than making ever longer, if not also harder-to-maintain or imperfect, regular expressions.

This could also be potentially useful for our other rules, including fixers where we use its parser and stringifier to make modifications and where, without such whitespace-awareness, the fixers may thus drop whitespace that the user wanted to preserve.

Thanks!

renatho:
> I can take a look at this, but I'm wondering if you might be open to adding support for an option to preserve whitespace in [comment-parser](https://github.com/syavorsky/comment-parser)--which is a small library that we are elsewhere using in our codebase and has been open to PRs?
> 
> I'd feel a lot more comfortable if the rule were to instead total up whitespace counts based on parsed results rather than making ever longer, if not also harder-to-maintain or imperfect, regular expressions.
> 
> This could also be potentially useful for our other rules, including fixers where we use its parser and stringifier to make modifications and where, without such whitespace-awareness, the fixers may thus drop whitespace that the user might have wanted to preserve.
> 
> Thanks!

Interesting! Yep! I'll try to take a look at this on the next days. :)

renatho:
@brettz9, I wonder if we could merge this fix for now, and when the comment-parser changes are done, we update this rule to use it.

Considering that probably it can take a while, I think it'd be nice to already fix this issue here. WDYT?

brettz9:
@renatho : I really don't want to introduce new code that has to be maintained with its own complexities. It is not going super fast, I know, but the project maintainer has been making progress on this new release and has been receptive to changes in the past, so I'd hold out a little bit longer. We will need those `trim` and `indent` features so if they don't get added in his new release, you could help expedite things by adding a PR to add them back here (or if it turns out too inefficient or cumbersome to do them post-parsing/post-stringifying, then maybe to propose a generic hook for `comment-parser` so we can supply a callback to do the trimming, etc.).

renatho:
> @renatho : I really don't want to introduce new code that has to be maintained with its own complexities. It is not going super fast, I know, but the project maintainer has been making progress on this new release and has been receptive to changes in the past, so I'd hold out a little bit longer. We will need those `trim` and `indent` features so if they don't get added in his new release, you could help expedite things by adding a PR to add them back here (or if it turns out too inefficient or cumbersome to do them post-parsing/post-stringifying, then maybe to propose a generic hook for `comment-parser` so we can supply a callback to do the trimming, etc.).

Hey @brettz9! I get your point. But I was thinking that maybe it could be nice to merge this one only because it's more a fix than a new feature. But if you prefer to wait a little more for this, it's okay. ;)

brettz9:
I've gone ahead and applied your changes. For future changes, please note that `comment-parser` has now been updated in a way where we should be able to modify its parsed tokens and stringify them to achieve alignment changes.

Note that `comment-parser` was only recently updated, so it may yet add tweaks and improvements which make it easier to stringify. Currently if one makes changes on the `tokens` property of the objects in its parsed `Block.tags[].source` array, one has to copy these changes to the `Block.source` array (which mirrors the tag `source` along with holding non-tag source such as the jsdoc block description (so that full jsdoc block stringifying can be done).

renatho:
> I've gone ahead and applied your changes. For future changes, please note that `comment-parser` has now been updated in a way where we should be able to modify its parsed tokens and stringify them to achieve alignment changes.
> 
> Note that `comment-parser` was only recently updated, so it may yet add tweaks and improvements which make it easier to stringify. Currently if one makes changes on the `tokens` property of the objects in its parsed `Block.tags[].source` array, one has to copy these changes to the `Block.source` array (which mirrors the tag `source` along with holding non-tag source such as the jsdoc block description (so that full jsdoc block stringifying can be done).

Hey @brettz9! Thank you for the quick update of the `comment-parser` dependency! I want to update the rule to use the `comment-parser` as soon as I can!

PR #644, opened 2020-09-28
fix(`require-jsdoc`): support `TSTypeAliasDeclaration` with `publicOnly`

pauliunas:


brettz9:
Hi... Thanks for these... 

Do you expect to add any other similar changes? I think it'd be ideal if we could avoid bumping versions (since separate merges cause new releases) if you know of some similar changes that we might add to this PR before merging...

pauliunas:
> 
> 
> Hi... Thanks for these...
> 
> Do you expect to add any other similar changes? I think it'd be ideal if we could avoid bumping versions (since separate merges cause new releases) if you know of some similar changes that we might add to this PR before merging...

yeah I thought about that, sorry I forgot to add them both in the first PR. I can't think of anything we might need, but let's keep this open until tomorrow just in case something comes up.

README.md
```diff
@@ -1,14 +1,14 @@
 <a name="eslint-plugin-jsdoc"></a>
-# eslint-plugin-jsdoc
-
-[![GitSpo Mentions](https://gitspo.com/badges/mentions/gajus/eslint-plugin-jsdoc?style=flat-square)](https://gitspo.com/mentions/gajus/eslint-plugin-jsdoc)
-[![NPM version](https://img.shields.io/npm/v/eslint-plugin-jsdoc.svg?style=flat-square)](https://www.npmjs.org/package/eslint-plugin-jsdoc)
-[![Travis build status](https://img.shields.io/travis/gajus/eslint-plugin-jsdoc/master.svg?style=flat-square)](https://travis-ci.org/gajus/eslint-plugin-jsdoc)
-[![js-canonical-style](https://img.shields.io/badge/code%20style-canonical-blue.svg?style=flat-square)](https://github.com/gajus/canonical)
-[![Discord Chat](https://img.shields.io/badge/chat-on%20disord-green.svg?logo=discord)](https://discord.gg/kFFy3nc)
-
-JSDoc linting rules for ESLint.
```
pauliunas, line 10: again... sorry for the mess, I made sure VSCode is configured for LF line breaks and the file wasn't open, but still had the same problem. If I change them back to LF, the push hook still complains and asks me to run the generator again.

brettz9: No worries---I can handle all of the rest. All good, thanks for these!

pauliunas:
Alright, I think you can merge this now, I don't see us needing anything else in the foreseeable future. If we end up needing something else, I'll just go through the full list of AST definitions and add them all in one batch. But that's a more long-term goal.

brettz9:
Ok, sure, I can merge. But out of curiosity, I wonder whether you had a chance, or could, look at the items with "Declaration" in them and if there were any you could guess at their syntax, whether they could be exportable?

pauliunas:
> Ok, sure, I can merge. But out of curiosity, I wonder whether you had a chance, or could, look at the items with "Declaration" in them and if there were any you could guess at their syntax, whether they could be exportable?

yeah, I have a work in progress list of potentially exportable items:
 TSAbstractClassProperty
 TSAbstractMethodDefinition
 TSDeclareFunction
 TSEnumMember
 TSInterfaceDeclaration
 TSMethodSignature
 TSModuleDeclaration
 TSTypeAliasDeclaration
TSExportAssignment
TSIndexSignature
TSNamespaceExportDeclaration

I'm almost certain the top 8 are exportable, not really sure what the remaining 3 are. But I've decided to put this away for when I have more time, or more need, to investigate them and make sure I understand them correctly. Since typescript-eslint AST has no documentation at all, it's mostly just guess work and some playing around with AST Explorer. It also has different names than those used by Typescript compiler internally, and, subsequently, by TSLint, which doesn't help.

Issue #658, opened 2020-11-29
Support ignore-next-line inside JSDoc comment block


basickarl:
I'm unsure if this issue belongs in this repo or if I should file it over at eslint since `eslint-disable-next-line` is an eslint thing. But then again you'd never need it to work inside a JSDoc block if you aren't using JSDoc.
```
/**
 * @param {Messages} messages Object containing the messages in chronological order.
 * eslint-disable-next-line jsdoc/no-undefined-value
 * @returns {processMessage} Function that will handle incoming messages.
 */
export function processMessageWrapper(messages) {
    /**
     * Processes incoming messages.
     *
     * @param {string} message Incoming message from stdin.
     */
    function processMessage(message) {
```

## Motivation

To disable rules for specific lines.

## Current behavior

Doesn't work, it's not possible to disable next lines in JSDoc comment blocks.

## Desired behavior

Would like to disable next line inside JSDoc comment block.

brettz9:
As far as disabling lines, ESLint only offers the mechanism to do something like this:

```js
 /* eslint-disable jsdoc/no-undefined-types */
/**
 * @param {Messages} messages Object containing the messages in chronological order.
 * @returns {processMessage} Function that will handle incoming messages.
 */
export function processMessageWrapper(messages) {
   /* eslint-enable jsdoc/no-undefined-types */
```

Another option would be to add `processMessage` to the `definedTypes` array option.

However, based on your example, it looks like what might want to do is something like this, which should not report any error (and might also work better with jsdoc tooling, though I'm not sure):

```js
/**
 * Processes incoming messages.
 * @callback ProcessMessage
 * @param {string} message Incoming message from stdin.
 */

/**
 * @param {Messages} messages Object containing the messages in chronological order.
 * @returns {ProcessMessage} Function that will handle incoming messages.
 */
export function processMessageWrapper(messages) {
    /**
     * @type {ProcessMessage}
     */
    function processMessage(message) {
```

I.e., I wouldn't think that jsdoc would expect `processMessage` to be accessible by the time it is used, so the above might be the more proper way around it. I'm not fully sure about this, but it is the way I would handle it.

Closing as that should address the main question, but feel free to comment further as needed.

Yash-Singh1:
Could the plugin itself have some sort of internal way of detecting `eslint-disable` comments inside JSDoc comment blocks that will apply to all rules?

brettz9:
[...]

Issue #648, opened 2020-10-09
publicOnly doesn't recognize classes xor methods


apowers313:
Reposted from #288 

## Expected vs actual behavior
This flavor of exporting a class triggers `Missing JSDoc comment` on the `method` but not the `class`:
``` js
class Utility {
    mthd() {
        return false;
    }
}

module.exports = Utility;
```

This flavor of exporting a class triggers `Missing JSDoc comment` on the `class` but not the `method`:
``` js
module.exports = class Utility {
    mthd() {
        return false;
    }
};
```

## ESLint Config

``` js
"jsdoc/require-jsdoc": [
    "error", {
        enableFixer: false,
        publicOnly:  true,
        require:     {
            ClassDeclaration:        true,
            ClassExpression:         true,
            FunctionDeclaration:     true,
            FunctionExpression:      true,
            MethodDefinition:        true,
            ArrowFunctionExpression: true,
        },
    },
],
```

## Environment

- Node version: v14.8.0
- ESLint version v5.16.0
- `eslint-plugin-jsdoc` version: 30.6.3

PsycheSloLo811:
![gmsnet2](https://user-images.githubusercontent.com/72146486/97946611-99cefb00-1d50-11eb-8652-00c87dc95127.jpg)

PsycheSloLo811:
I have no idea what I am doing what do you think

brettz9:
@PsycheSloLo811  : Your image isn't showing anything

PR #687, opened 2021-02-04
Fix typescript interface working with `publicOnly` options

wwwzbwcom:
In typescript, when interface is export, property and method is also exposed. but unlike class method and property, they are not retrievable from AST, so they failed on both esm and AST check of public, in this PR I fix this issue.

more detail can be find in tests

### DIFF
--- a/test/rules/assertions/requireJsdoc.js
+++ b/test/rules/assertions/requireJsdoc.js
@@ -8,6 +8,96 @@ export default {
       code: `
           /** This is comment */
           export interface Foo {
+            /** This is comment x2 */
+            tom: string;
+            catchJerry(): boolean;
+          }
+      `,
+      errors: [
+        {
+          message: 'Missing JSDoc comment.',
+          type: 'TSMethodSignature',
+        },
+      ],
+      options: [
+        {
+          contexts: [
+            'TSInterfaceDeclaration',
+            'TSMethodSignature',
+            'TSPropertySignature',
+          ],
+          publicOnly: {
+            ancestorsOnly: true,
+          },
+          require: {
+            ClassDeclaration: true,
+            ClassExpression: true,
+            MethodDefinition: true,
+          },
+        },
+      ],
+      output: `
+          /** This is comment */
+          export interface Foo {
+            /** This is comment x2 */
+            tom: string;
+            /**
+             *
+             */
+            catchJerry(): boolean;
+          }
+      `,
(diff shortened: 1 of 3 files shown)


test/rules/assertions/requireJsdoc.js
```diff
@@ -4,6 +4,124 @@
...
+
+        /**
+         *
+         */
+        meow(): void;
+      }
+      `,
+      parser: require.resolve('@typescript-eslint/parser'),
+    },
     {
```
brettz9, line 125: How about a valid test too, e.g., to ensure a non-public one won't be flagged by `publicOnly`?

brettz9 (changes requested):
Looks good AFAICT. Just adding that one valid test.

FWIW, if you were feeling inspired, it'd be great if someone could take a look at https://github.com/gajus/eslint-plugin-jsdoc/issues/648 too. No worries if not...

Thanks!

test/rules/assertions/requireJsdoc.js
wwwzbwcom: There is no private in typescript `interface`.
But private method in exported class will always considered as exported, for even if the esm check consider the method is not exported, the variables map check can find the private method (cause the variable map construct dont consider any typescript feature)
In this PR I would prefer to focus on interface checking.

wwwzbwcom:
I found that the previous version will consider interface in a exported function as 'exported', but actually it's not, for example:

```ts
export function Tom() {
    interface Jerry {
        spikie: string;
    }
}
```

The interface `Jerry` should not be considered as exported.

I add the checkings for this situation in the latest commit.

test/rules/assertions/requireJsdoc.js
wwwzbwcom: Can make more PR about typescript classes, also #658 later

brettz9:
Yes, that's what I was getting at. Would be good to have a `valid` test where the latter example doesn't fail. Thanks!

wwwzbwcom:
> Yes, that's what I was getting at. Would be good to have a `valid` test where the latter example doesn't fail. Thanks!

Oh, thanks for reminds me that, any further changes for this PR needed?

brettz9:
You mentioned some additional changes, but I don't see any new commits. Just be sure to push. :-)

wwwzbwcom:
forget that, pushed

brettz9:
Looked good to me. Merged!

Issue #680, opened 2021-01-23
check-line-alignment forces a misalignment


TheJaredWilcurt:
Happened when updating from `30.7.6` to `31.2.1`

* **Reproduction Repo:** https://github.com/nwutils/create-desktop-shortcuts

* * *

My existing block
```js
/**
 * Creates OS based shortcuts for files, folders, and applications.
 *
 * @param  {object}  options  Options object for each OS.
 * @return {boolean}          True = success, false = failed to create the icon
 */
```
After running `-fix` with `check-line-alignment` enabled it outputs this:
```js
/**
 * Creates OS based shortcuts for files, folders, and applications.
 *
 * @param {object} options Options object for each OS.
 * @return {boolean}          True = success, false = failed to create the icon
 */
```
It was nice and vertically aligned, but, ironically, the linting rule that attempts to enforce such behavior, instead breaks it across all files.

* * *

Another example:

```js
    /**
     * Returns the value stored in the process.env for a given
     * environment variable.
     *
     * @param  {string} withPercents    '%USERNAME%'
     * @param  {string} withoutPercents 'USERNAME'
     * @return {string}                 'bob' || '%USERNAME%'
     */
```
is "fixed" to
```js
    /**
     * Returns the value stored in the process.env for a given
     * environment variable.
     *
     * @param {string} withPercents '%USERNAME%'
     * @param {string} withoutPercents 'USERNAME'
     * @return {string}                 'bob' || '%USERNAME%'
     */
```

* * *

```
  'rules': {
    'jsdoc/check-line-alignment': 1
  }
```

* https://github.com/nwutils/create-desktop-shortcuts/blob/master/.eslintrc.js

* * *

## Environment

* **Node.js** 14.15.4
* **ESLint:** 7.18.0
* **eslint-plugin-jsdoc**: 31.2.1

brettz9:
@TheJaredWilcurt : The rule has changed so that if you want alignment, you need to add the "always" option:

```js
  'rules': {
    'jsdoc/check-line-alignment': [1, 'always']
  }
````

However, for your case, the code still does not seem to align properly. @renatho : Do you think you could take a look for the "always" code that is the problem here? I don't think this is related to my later changes which were for the "never" option, as the issue still occurs after adding the "always" option.

(It also looks like the "never" behavior is not fully working either, both because the rule hasn't to date checked `@return` tags and because of something else I plan to take a look at tomorrow).

renatho:
[...]

PR #689, opened 2021-02-09
Update check line alignment to check by tag

renatho:
Part of https://github.com/gajus/eslint-plugin-jsdoc/issues/680

This PR updates the check line alignment to check by tag (when using the `always` option) instead of the whole block, as suggested here: https://github.com/gajus/eslint-plugin-jsdoc/issues/680#issuecomment-770314024.

It also adds support to the `tags` option.

My approach was cloning the `jsdoc`, parse that, and then get the aligned positions fixing the original one only in the _applicable tags_.


src/rules/checkLineAlignment.js
```diff
@@ -97,41 +103,93 @@ const checkNotAlignedPerTag = (utils, tag) => {
   utils.reportJSDoc('Expected JSDoc block lines to not be aligned.', tag, fix, true);
 };
 
-const checkAlignment = ({
+const filterApplicableTags = (jsdoc, applicableTags) => {
```
renatho, line 106: I'm not sure if it's a good approach with comment-parser, but it's working. Basically, we remove the tags we don't want to align to align only the ones we want.

src/rules/checkLineAlignment.js
```diff
@@ -97,41 +103,93 @@ const checkNotAlignedPerTag = (utils, tag) => {
...
+    jsdoc,
+    utils,
+  });
+
+  foundTags.forEach((tag, index) => {
+    const formattedTag = jsdocFormatted.tags[index];
+    if (!isTagSourcesEqual(tag, formattedTag)) {
+      const fix = () => {
+        const formattedSource = getFormattedSource(tag, formattedTag);
+        utils.replaceTagSource(tag, formattedSource);
```
renatho, line 175: And here we update the original tag to use the formatted one.

src/rules/checkLineAlignment.js
```diff
@@ -97,41 +103,93 @@ const checkNotAlignedPerTag = (utils, tag) => {
...
   indent,
   jsdoc,
   jsdocNode,
-  report,
   context,
   utils,
 }) => {
   const {
     tags: applicableTags = ['param', 'arg', 'argument', 'property', 'prop', 'returns', 'return'],
   } = context.options[1] || {};
```
TheJaredWilcurt, line 191: This was pretty hard to read. A cleaner way of writing it would be:
```js
  const userTags = context.options?.[1]?.tags;
  const defaultTags = [
    'param',
    'arg',
    'argument',
    'property',
    'prop',
    'returns',
    'return'
  ];
  const applicableTags = userTags || defaultTags;
```
Or if this codebase doesn't use optional chaining:
```js
  const userTags = (
    context.options &&
    context.options[1] &&
    context.options[1].tags
  );
  const defaultTags = [
    'param',
    'arg',
    'argument',
    'property',
    'prop',
    'returns',
    'return'
  ];
  const applicableTags = userTags || defaultTags;
```

renatho: Hey @TheJaredWilcurt !
Thank you for checking! Since it's not part of this PR, I'd prefer don't touch this part. If we want to refactor this part, maybe we could create a new PR for that. :)

TheJaredWilcurt: I know, I was just doin' [boy scout rule](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html) ⚜️

PR #703, opened 2021-03-14
Add line alignment tags support when using `always`

renatho:
Part of #680
Close #689

This PR adds support to the `tags` option when using the `always` option.

The first PR for this feature was the https://github.com/gajus/eslint-plugin-jsdoc/pull/689, but the approach wasn't being enough, needing a lot of hacks. So I changed the approach, creating a new transform based on the original `align`, but adding support to custom tags.

It still reports the errors for the entire block, different than what was suggested in https://github.com/gajus/eslint-plugin-jsdoc/issues/680#issuecomment-770314024.

Another idea for a new feature in the future:

- We could add a new option to force the alignment similar to the `never` for the tags that aren't in the `tags` when using the `always`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2167,6 +2167,15 @@ const fn = ( lorem, sit ) => {}
 // Message: Expected JSDoc block lines to be aligned.
 
 /**
+ * Only return doc.
+ *
+ * @return {boolean}  Return description.
+ */
+const fn = ( lorem, sit ) => {}
+// "jsdoc/check-line-alignment": ["error"|"warn", "always"]
+// Message: Expected JSDoc block lines to be aligned.
+
+/**
  * Creates OS based shortcuts for files, folders, and applications.
  *
  * @param  {object}  options  Options object for each OS.
@@ -2378,6 +2387,14 @@ const fn = ( lorem ) => {}
  *
  * @param  {object}  options Options object for each OS.
  * @return {boolean}
+ */
+ function quux () {}
+// "jsdoc/check-line-alignment": ["error"|"warn", "always"]
+
+/**
+ * Only return doc.
+ *
+ * @return {boolean} Return description.
  */
  function quux () {}
 // "jsdoc/check-line-alignment": ["error"|"warn", "always"]
--- a/README.md
+++ b/README.md
@@ -1960,6 +1960,16 @@ const fn = ( lorem, sit ) => {}
 // Message: Expected JSDoc block lines to be aligned.
 
 /**
+ * With tabs.
+ *
+ * @param {string} lorem Description.
+ * @param {int} sit Description multi words.
+ */
+    const fn = ( lorem, sit ) => {}
+// "jsdoc/check-line-alignment": ["error"|"warn", "always"]
+// Message: Expected JSDoc block lines to be aligned.
+
+/**
  * Function description.
  *
  * @param {string} lorem - Description.
@@ -2249,6 +2259,15 @@ The following patterns are not considered problems:
  * @param {int}    sit   Description multi words.
  */
 const fn = ( lorem, sit ) => {}
+// "jsdoc/check-line-alignment": ["error"|"warn", "always"]
+
+/**
+ * With tabs.
+ *
+ * @param {string} lorem Description.
+ * @param {int}    sit   Description multi words.
+ */
+    const fn = ( lorem, sit ) => {}
 // "jsdoc/check-line-alignment": ["error"|"warn", "always"]
 
 /**
--- a/README.md
+++ b/README.md
@@ -2091,6 +2091,19 @@ const config = {
 // Message: Expected JSDoc block lines to be aligned.
 
 /**
+ * My object.
+ *
+ * @typedef {Object} MyObject
+ *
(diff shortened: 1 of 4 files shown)


renatho:
Ah, it still needs some more tests for full coverage.

TheJaredWilcurt:
Ran it against my codebase and it worked without any bugs. I just wish that you could have 2 spaces after a variable name. without it, it becomes hard to read and looks like it's all just one sentence. example:

```js
/**
 * Mocks the process.platform to a specified value.
 *
 * @param {string} platform 'win32', 'linux', or 'darwin'.
 */
```

I'd prefer a two space buffer before and after a variable name

```js
/**
 * Mocks the process.platform to a specified value.
 *
 * @param {string}  platform  'win32', 'linux', or 'darwin'.
 */
```

renatho:
@TheJaredWilcurt, thank you for testing! About these other requisites, I think it could be part of a different PR.

I also tested it against the Gutenberg project and I identified only one issue. Basically, if we have only the `@return` tag, it'll add 2 spaces after the tag because the approach used with the tokenizers. This issue is currently present already, so I think we could merge this that already have improvements, and we could fix it in another PR.

I'm not sure yet how we could fix this. Do you have any idea @brettz9? I'd like to fix it without changing the align transformer because I tried to make it more similar to the original as possible, so we could remove this if we add it to the comment parser lib.

renatho:
I checked a little more, and actually I think there is not problem with how it's using the tokenizers. So I already fixed it through the transformer in this commit: https://github.com/gajus/eslint-plugin-jsdoc/pull/703/commits/6d2961059d043ebc01e7024c77b6ab7e5af6da3f

Anyway, I'd appreciate your feedback when you have some time, @brettz9 =)

renatho:
I found another issue related to the indentation with tabs. Fixed here: https://github.com/gajus/eslint-plugin-jsdoc/pull/703/commits/8389ad31096a8078b47becb0206b11aad01740f3

renatho:
It's almost there. Now I found another weird error with the `template` tag (duplicating some content).

I think it's probably because [this](https://github.com/gajus/eslint-plugin-jsdoc/blob/master/src/iterateJsdoc.js#L54-L72), but I'll try to check it another day.

renatho:
> It's almost there. Now I found another weird error with the `template` tag (duplicating some content).
> 
> I think it's probably because [this](https://github.com/gajus/eslint-plugin-jsdoc/blob/master/src/iterateJsdoc.js#L54-L72), but I'll try to check it another day.

I took some time to debug it. The issue is that we're getting the `name` and the `description` tokens duplicated in the following case, for example:

```js
          /**
           * With template tag.
           *
           * @template T
           *
           */
```

See the result debugging [the commented part](https://github.com/gajus/eslint-plugin-jsdoc/blob/master/src/iterateJsdoc.js#L54-L72).

Screen Shot 2021-03-31 at 18 44 13

The same happens for `@template {string} T`.

renatho:
Hey @brettz9!
Do you have any idea of where the error could be in that part I mentioned in the previous comment? I suspect it could be in the `pos`, but honestly, I'm a little confused about the meanings of the variables `remainder`, `pos`, and `extra`. If you have some clues, I could try to fix it.

brettz9:
If you log against the existing tests, you can see `spec.source[0].tokens`, and how `description` gives the remainder after `@template` (there's no need for `name` as it is not populated).

`pos` is the position at which the first whitespace that is itself not preceded by whitespace or a comma is found (i.e., where the template name or sequence of names stop).

`extra` is then everything after that post-name(s) whitespace (i.e., any extra postName whitespace and any description).

brettz9:
The repo is at https://github.com/es-joy/jsdoccomment , but I can add if I can see the reason for it. However, I don't imagine we can add such a conditional because if it is something like the example used at https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#template `* @template {string} K - K must be a string or string literal`, making it conditional will shave off the postName and description. What is the concrete problem you are trying to solve in making this change?

renatho:
> The repo is at https://github.com/es-joy/jsdoccomment , but I can add if I can see the reason for it. However, I don't imagine we can add such a conditional because if it is something like the example used at https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#template `* @template {string} K - K must be a string or string literal`, making it conditional will shave off the postName and description. What is the concrete problem you are trying to solve in making this change?

Hey @brettz9 !

I'm not in the computer now, but I remember I tested with the examples of this link and it worked well.

When `pos` is not found (-1), I think it means we don't have anything after the name, right?

Basically, the problem is with a case like (when we don't have anything after the name):

`@template T`

In this case, it's setting `T` for name and for description.

brettz9:
Ah yes, sorry, @renatho . Fixed in 31.1.1

renatho:
Nice!!! Than you @brettz9 !! \o/
In this case, this PR is ready for a final review :)

brettz9:
Thanks for all your work and persistence on this, @renatho !

brettz9:
Ugh, seems my commit was not well-formed. Adding a commit to revert and yet another to add back in the proper format.

PR #727, opened 2021-05-11
feat: new rule `tag-line`; fixes #93

### LINKED ISSUE #93: Add rule to (dis)allow empty lines between tags
In our code base, there is no real consistency between having empty lines or not between the tags of docblocks. I want to have more consistency and want to disallow empty lines between then, while having the empty line after the description. So I would like to see this implemented. If I have time, I would like to try it myself, but I can't promise anything.

brettz9:
Includes a `always` vs. `never` option as well as two special options for `always` (`count` and `noEndLine`). Fixes #93.

### DIFF
--- a/.README/rules/tag-lines.md
+++ b/.README/rules/tag-lines.md
@@ -13,7 +13,7 @@ The second option is an object with the following optional properties.
 
 Use with "always" to indicate the number of lines to require be present.
 
-##### `noEndLine` (defaults to `false`)
+##### `noEndLines` (defaults to `false`)
 
 Use with "always" to indicate the normal lines after tags should not be
 added after the final tag.
@@ -24,6 +24,6 @@ added after the final tag.
 |Tags|Any|
 |Recommended|false|
 |Settings|N/A|
-|Options|(a string matching `"always" or "never"` and optional object with `count` and `noEndLine`)|
+|Options|(a string matching `"always" or "never"` and optional object with `count` and `noEndLines`)|
 
 <!-- assertions tagLines -->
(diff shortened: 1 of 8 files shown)


golopot:
What does `noEndLine: true` do? I thought it bans empty lines after last tag but there is no tests for it?

brettz9:
There are tests for it, but I can look to add some passing tests instead of only failing ones.  You might want to hold off a little, as I think I have some fixing to do as far as line numbers (and am seeking to add tests for that).

golopot:
Do you think it would be better if the always/never options is dropped, and that fee "count: 0" means no empty lines are allowed?

brettz9:
> Do you think it would be better if the always/never options is dropped, and that fee "count: 0" means no empty lines are allowed?

The one reason I thought it may be ok, as is, is because we can have a sensible defaults for "always" which avoids the need for specifying a number in that case as with "never". But I'm ok with switching it if you think it should be changed.

I've updated, btw, so the line number tests/fixes and added test should otherwise be ready for review.

.README/rules/tag-lines.md
```diff
@@ -0,0 +1,28 @@
...
+
+The second option is an object with the following optional properties.
+
+##### `count` (defaults to 1)
+
+Use with "always" to indicate the number of lines to require be present.
+
+##### `noEndLine` (defaults to `false`)
+
+Use with "always" to indicate tag lines should not be added at the end.
```
golopot, line 18: I don't understand this option. What is a tag line?

brettz9: Clarified to state:

> Use with "always" to indicate the normal lines after tags should not be
added after the final tag.

Also changed `noEndLine` to `noEndLines` as can be more than one between tags normally.

PR #729, opened 2021-05-12
feat(`check-values`): ensure `@variation` is a positive integer

brettz9:


.README/rules/check-values.md
```diff
@@ -7,9 +7,11 @@ This rule checks the values for a handful of tags:
 2. `@since` - As with `@version`
 3. `@license` - Checks that there is a present and valid SPDX identifier
     or is present within an `allowedLicenses` option.
-4. `@author` - Checks there is a value present, and if the option
+4. `@author` - Checks that there is a value present, and if the option
     `allowedAuthors` is present, ensure that the author value is one
     of these array items.
+5. `@variation` - Checks that there iis a value present, and that it is
```
golopot, line 13: But the doc say you can use any value!

https://jsdoc.app/tags-variation.html

> You can provide any value with the @variation tag, as long as the combination of the value and the longname results in a globally unique version of the longname.

brettz9: Yes, I think you're right it should be optional and perhaps off by default, but it does specify `variationNumber` in the spec.

brettz9: Added now only if a non-default `numericOnlyVariation` is set to `true`

brettz9: Plan to be merging tonight, so please let me know if you are not done reviewing (or if you are done).

test/rules/assertions/checkValues.js
```diff
@@ -206,6 +225,44 @@ export default {
...
+       * @variation
+       */
+      function quux (foo) {
+
+      }
+      `,
+      errors: [
+        {
+          line: 3,
+          message: 'Missing JSDoc @variation.',
```
golopot, line 243: Should be "Missing value for JSDoc variation."

brettz9: Thanks, fixed, along with the messages for other Missing tag values that preexisted.

PR #731, opened 2021-05-13
Add `preserveMainDescriptionPostDelimiter` option to `check-line-alignment` rule

renatho:
This introduces the new option `preserveMainDescriptionPostDelimiter ` for the `check-line-alignment` rule.

This new option will preserve the post delimiter (spacings after the `*`) in the main description when using the `always` option.

```js
      /**
       * Function description
       *           description with post delimiter.
       *
       */
```

Notice that it just applies to `always` because the `never` always preserves the spacing. I thought about doing it as default to `always` too, but since we already released it in that way, I added a new option to avoid breaking changes.

### DIFF
--- a/src/alignTransform.js
+++ b/src/alignTransform.js
@@ -58,7 +58,7 @@ const space = (len) => {
   return ''.padStart(len, ' ');
 };
 
-const alignTransform = (tags, indent) => {
+const alignTransform = (tags, indent, preserveMainDescriptionPostDelimiter) => {
   let intoTags = false;
   let width;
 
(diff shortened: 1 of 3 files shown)


renatho:
@brettz9, thank you for the super quick merge! =)
Just confirming. Is there a plan to release it soon? And thank you for all your support in all these changes!

brettz9:
For some reason, when doing squashes and adding the [Angular](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines) commit format needed for the automated semantic releases, I am not noticing there is a commit summary before it getting applied which negates the auto-release. I will try to be mindful of this in the future. Thanks very much for the feature!

PR #730, opened 2021-05-13
`no-multi-asterisks` rule

brettz9:
feat: `no-multi-asterisks` rule to prevent multiple asterisks with middle delimiters or at end of blocks

Fixes #199 (other fixes have now been merged).

PR #733, opened 2021-05-15
Multiline blocks

brettz9:
Should fix the last of #199 whenever ~#727, #728, and~ #730 may be merged.

~Builds on #727. You can review the diffs now or I can wait to rebase whenever that may be merged.~

feat(`multiline-blocks`): allow control of how and whether jsdoc blocks can be expressed as single or multiple line blocks

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -552,6 +552,7 @@ selector).
 {"gitdown": "include", "file": "./rules/empty-tags.md"}
 {"gitdown": "include", "file": "./rules/implements-on-classes.md"}
 {"gitdown": "include", "file": "./rules/match-description.md"}
+{"gitdown": "include", "file": "./rules/multiline-blocks.md"}
 {"gitdown": "include", "file": "./rules/newline-after-description.md"}
 {"gitdown": "include", "file": "./rules/no-bad-blocks.md"}
 {"gitdown": "include", "file": "./rules/no-defaults.md"}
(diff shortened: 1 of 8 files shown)


danvk:
Thanks for accommodating my quirky JS Doc preferences with `minimumLengthForMultiline` @brettz9!

One issue I noticed is that setting `noMultilineBlocks` seems to disable `noZeroLineText`.

For example, this comment:

```ts
/** This comment is bad
 * It should not have text on line zero.
 */
```

is allowed with:

```
{
  noZeroLineText: true,
  noMultilineBlocks: true,
  minimumLengthForMultiline: 50
}
```

but with just:

```
{ noZeroLineText: true }
```

it's an error:

```
Should have no text on the "0th" line (after the `/**`). eslintjsdoc/multiline-blocks
```

The documentation hints at an interaction here ("`noZeroLineText`: `noMultilineBlocks` will have priority over this rule if it applies.") but I don't see what the connection is. I'd like to enforce that, if a JS Doc comment uses the multiline form because of its length, then it should have no text on line zero. The text should all be on lines 1, 2, etc.

brettz9:
@danvk : I hope to take a look at this maybe tomorrow or as I have energy, but could you add this as a separate issue? It's no bother to have a separate issue filed and linking back to the issue with a new one better ensures the concern won't be dropped. Thanks!

brettz9:
(However, you might clarify your comments there because it sounds like your statement "setting `noMultilineBlocks` seems to disable `noZeroLineText`" contradicts what your example shows.)

danvk:
Sure thing @brettz9, filed https://github.com/gajus/eslint-plugin-jsdoc/issues/737

Thanks for tackling all these issues! Your work is greatly appreciated.

PR #740, opened 2021-05-16
Add custom spacing to line alignment rule

### LINKED ISSUE #732: jsdoc/check-line-alignment option to take into account existing alignment
With my current settings, the linter wants to change my alignment from this
```js
  /**
   * Description
   *
   * @param  {object} options  The user's options
   * @param  {string} message  The text to be logged
   * @param  {object} error    Optional object with additional details
   */
```
to this
```js
  /**
   * Description
   *
   * @param {object} options The user's options
   * @param {string} message The text to be logged
   * @param {object} error   Optional object with additional details
   */
```

It should be able to detect that each section is starting at the same point on each line.
```js
    'jsdoc/check-line-alignment': [
      1,
      'always',
      {
        'tags': [
          'param',
          'arg',
          'argument',
          'property',
          'prop',
          'returns',
          'return'
        ]
      }
    ],
```

* [all lint settings](https://github.com/nwutils/create-desktop-shortcuts/blob/lint/.eslintrc.js)

renatho:
Fixes #732

It introduces the `customSpacings` option to the line alignment rule. You can use something like:

```js
        customSpacings: {
          postDelimiter: 2,
          postTag: 3,
          postType: 2,
        },
```

If the spacing is not defined, it uses `1` by default.

### DIFF
--- a/src/rules/checkLineAlignment.js
+++ b/src/rules/checkLineAlignment.js
@@ -65,7 +65,7 @@ const checkNotAlignedPerTag = (utils, tag, customSpacings) => {
 
     // There will be extra alignment if...
 
-    // 1. The spaces doesn't match the space it should have (1 or custom spacing) OR
+    // 1. The spaces don't match the space it should have (1 or custom spacing) OR
     return spacerPropVal.length !== spacing && spacerPropVal.length !== 0 ||
 
       // 2. There is a (single) space, no immediate content, and yet another
--- a/src/rules/checkLineAlignment.js
+++ b/src/rules/checkLineAlignment.js
@@ -97,6 +97,7 @@ const checkNotAlignedPerTag = (utils, tag) => {
 };
 
 const checkAlignment = ({
+  customSpacings,
   indent,
   jsdoc,
   jsdocNode,
@@ -105,7 +106,14 @@ const checkAlignment = ({
   tags,
   utils,
 }) => {
-  const transform = commentFlow(alignTransform(tags, indent, preserveMainDescriptionPostDelimiter));
+  const transform = commentFlow(
+    alignTransform({
+      customSpacings,
+      indent,
+      preserveMainDescriptionPostDelimiter,
+      tags,
+    }),
+  );
   const transformedJsdoc = transform(jsdoc);
 
   const comment = '/*' + jsdocNode.value + '*/';
@@ -133,6 +141,7 @@ export default iterateJsdoc(({
   const {
     tags: applicableTags = ['param', 'arg', 'argument', 'property', 'prop', 'returns', 'return'],
     preserveMainDescriptionPostDelimiter,
+    customSpacings,
   } = context.options[1] || {};
 
   if (context.options[0] === 'always') {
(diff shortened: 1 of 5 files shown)


src/rules/checkLineAlignment.js
```diff
@@ -174,6 +186,9 @@ export default iterateJsdoc(({
       {
         additionalProperties: false,
         properties: {
+          customSpacings: {
+            type: 'object',
+          },
```
brettz9, line 191: How about we get this child object specified more precisely with:
```js
additionalProperties: false,
properties: {
  postDelimiter: {
    type: 'integer',
  },
  postName: {
    type: 'integer',
  },
  postTag: {
    type: 'integer',
  },
  postType: {
    type: 'integer',
  },
},
```

src/rules/checkLineAlignment.js
```diff
@@ -61,11 +61,12 @@ const checkNotAlignedPerTag = (utils, tag) => {
     const contentProp = contentProps[idx];
     const contentPropVal = tokens[contentProp];
     const spacerPropVal = tokens[spacerProp];
+    const spacing = customSpacings?.[spacerProp] || 1;
 
     // There will be extra alignment if...
 
-    // 1. There is extra whitespace within a single spacer segment OR
-    return spacerPropVal.length > 1 ||
+    // 1. The spaces doesn't match the space it should have (1 or custom spacing) OR
```
brettz9, line 68:
```suggestion
    // 1. The spaces don't match the space it should have (1 or custom spacing) OR
```

test/rules/assertions/checkLineAlignment.js
```diff
@@ -917,6 +917,78 @@ export default {
...
+      /**
+       * Function description.
+       *
+       *  @param   {string}  lorem Description.
+       *  @param   {int}  sit Description multi words.
+       */
+      const fn = ( lorem, sit ) => {}
+      `,
+    },
   ],
```
brettz9, line 992: How about at least one valid test, and for invalid, adding one test with an explicit `postName` set to 2+ (different from the default).

brettz9 (changes requested):
Besides my comments, everything looks great to me from here... Thanks!

test/rules/assertions/checkLineAlignment.js
renatho: Sure! https://github.com/gajus/eslint-plugin-jsdoc/pull/740/commits/3d626b0e51bace5e8bc02ca8a56075720c69c86c

src/rules/checkLineAlignment.js
renatho: Ah, of course! https://github.com/gajus/eslint-plugin-jsdoc/pull/740/commits/4c973f52c8ee1a156bec02a28b7226ea5ee9eb4a

renatho:
Thank you for your review @brettz9! The changes were added. =)

brettz9:
I got the commit message fine I think, but Git is not finding the remote refs to complete the CI and release process. Trying to figure out how to resolve.

Issue #743, opened 2021-05-18
jsdoc/check-line-alignment adds extra spaces to objects


Allain55:
## Expected behavior
jsdoc/check-line-alignment should properly format objects with @param

## Actual behavior
jsdoc/check-line-alignment adds many extra spaces for objects

## ESLint Config
```
{
	'plugins': [
		'jsdoc'
	],
	'parser': 'babel-eslint',
	'parserOptions': {
		'ecmaVersion': 2021
	},
	'rules': {
		'jsdoc/check-line-alignment': ['error', 'always']
	}
};
```

## ESLint sample

```
/**
 * @param {{
 * ids: number[]
 * }} params
 */
const myMethod = ({ids}) => {}
```

The result is:
```js
/**
 * @param {{
 *                                                                       ids: number[]
 *                                                                       }}            params
 */
const myMethod = ({ids}) => {}
```
## Environment

- Node version: 16.0
- ESLint version 7.26
- `eslint-plugin-jsdoc` version: 34.7.0

brettz9:
@renatho , would you be able to take a look?

renatho:
Hey there!

I tested this example, but it gave me the result as valid.
Are you able to reproduce it @brettz9?

brettz9:
Yeah, I'm getting these results (using tabs):

```js
{
      /* eslint-disable no-tabs */
      code: `
				/**
				 * @param {{
				 * ids: number[]
				 * }} params
				 */
				const myMethod = ({ids}) => {}
			`,
      errors: [
        {
          line: 2,
          message: 'Expected JSDoc block lines to be aligned.',
          type: 'Block',
        },
      ],
      options: ['always'],
      output: `
				/**
				 * @param {{
				 *        ids: number[]
				 *        }}            params
				 */
				const myMethod = ({ids}) => {}
			`,
      /* eslint-enable no-tabs */
      parser: require.resolve('@babel/eslint-parser'),
      parserOptions: {
        ecmaVersion: 2_021,
      },
    },
```

(Btw, I got the same results with `babel-eslint` too, but it is probably a good idea to use `@babel/eslint-parser` instead as `babel-eslint` is deprecated in favor of it.)

renatho:
> I'm getting these results (using tabs or spaces)

You mean you got the issue? Or that it worked well for you? I tested the example you posted, and the tests passed.

brettz9:
[...]

PR #744, opened 2021-05-18
Add multiline workaround

renatho:
It's a workaround for #743.

Basically, it ignores multiline tags. To check it, it'll verify in the tokens if `tag` is empty and it contains `type` or `name`. The ideal solution would be to align it even being multiline.


src/alignTransform.js
```diff
@@ -101,6 +101,10 @@ const alignTransform = ({
       }
     }
```
brettz9, line 103:
```suggestion
// Todo: Avoid fixing alignment of blocks with multiline wrapping of type 
```

src/alignTransform.js
```diff
@@ -101,6 +101,10 @@ const alignTransform = ({
...
 
+    if (tokens.tag === '' && (tokens.type || tokens.name)) {
```
brettz9, line 104:
```suggestion
    if (tokens.tag === '' && tokens.type) {
```

Don't apply this if it is needed, but I think you will not get coverage for `tokens.name` because there will always be a type if there is a multiline name. `comment-parser` seems to treat the following as having only a description (i.e., `name desc`):

```js
/**
 * @tag
 * name desc
 */
```

If I am wrong, and the extra check is needed, however, please modify my to-do comment above to indicate that `name` is also checked.

brettz9 (changes requested):
Thanks for the quick fix and looks good if you could just review the comments and apply as necessary. Thanks!

renatho:
Thank you @brettz9! Changes applied! I also tested it against the Gutenberg repo and it looked okay, although `type` in multiline it's not common in that code.

brettz9:
Thanks If you haven't run `npm install` locally, please do so, as that will set up a proper environment that will do linting, and automated checks, etc.

Also for next time, if you can run `npm run create-readme` to update our README (and also make sure we have 100% coverage with `npm run test-cov` if you aren't doing that already.

Thanks!

PR #761, opened 2021-06-29
fix(no-missing-syntax): fix message

nicolashenry:


brettz9:
Thanks for the PR... I needed to add a test too, so just made the change locally. Thanks again!

PR #763, opened 2021-07-09
fix(`check-line-alignment`): preserve carriage returns; fixes #745

### LINKED ISSUE #745: check-line-alignment throwing warnings in GitHub Actions on Windows
So this is a weird one. I added `check-line-alignment` and it works locally on Windows.

* [settings file](https://github.com/nwutils/create-desktop-shortcuts/blob/main/.eslintrc.js#L24)

My repo is a cross-platform library, so I run GitHub Actions against it on Linux, OSX, and Windows with Node 14 (and Node 8, but ESLint 7 doesn't support Node 8, so just ignore that).

The weird thing is that only the Windows version is failing. And it isn't failing when ran locally on Windows.

Here is the GHA checks. you can switch between each OS and see that it's only on the Windows branch.

* https://github.com/nwutils/create-desktop-shortcuts/runs/2619667209?check_suite_focus=true

No idea what the cause is, all the code is stored in LF, so unless the GHA Windows VM is cloning in CRLF and I need to change some GHA setting, I don't think it's related to that.

brettz9:
Fixes #745.

Waiting on feedback on https://github.com/syavorsky/comment-parser/pull/132 (and possibly to use a stable update to `comment-parser` if also ready).

@renatho : You might want to see my [comment](https://github.com/syavorsky/comment-parser/pull/132#issuecomment-876814403) there. In particular, if the `delim`-setting code is unreachable, then I imagine we don't really need to expand on this line:

```js
    tokens.postDelimiter = nothingAfter.delim ? '' : space(spacings.postDelimiter);
```
to this:
```js
    tokens.postDelimiter = nothingAfter.delim ? (tokens.postDelimiter === '\r' ? '\r' : '') : space(spacings.postDelimiter);
```

### DIFF
--- a/test/rules/assertions/checkLineAlignment.js
+++ b/test/rules/assertions/checkLineAlignment.js
@@ -1057,6 +1057,36 @@ export default {
       const fn = ( lorem, sit ) => {}
       `,
     },
+    {
+      code: `\r
+        /**\r
+         * Function description.\r
+         *\r
+         * @param {string} lorem Description.\r
+         * @param {int} sit Description multi words.\r
+         */\r
+        const fn = ( lorem, sit ) => {}\r
+      `,
+      errors: [
+        {
+          line: 2,
+          message: 'Expected JSDoc block lines to be aligned.',
+          type: 'Block',
+        },
+      ],
+      options: [
+        'always',
+      ],
+      output: `\r
+        /**\r
+         * Function description.\r
+         *\r
+         * @param {string} lorem Description.\r
+         * @param {int}    sit   Description multi words.\r
+         */\r
+        const fn = ( lorem, sit ) => {}\r
+      `,
+    },
   ],
   valid: [
     {
@@ -1443,6 +1473,18 @@ export default {
          *        }}            params
          */
         const fn = ({ids}) => {}
+      `,
+      options: ['always'],
+    },
+    {
+      code: `\r
+        /**\r
+         * Function description.\r
+         *\r
+         * @param {string} lorem Description.\r
+         * @param {int}    sit   Description multi words.\r
+         */\r
+        const fn = ( lorem, sit ) => {}\r
       `,
       options: ['always'],
     },
--- a/test/rules/assertions/checkLineAlignment.js
+++ b/test/rules/assertions/checkLineAlignment.js
@@ -1064,6 +1064,8 @@ export default {
          *\r
          * @param {string} lorem Description.\r
          * @param {int} sit Description multi words.\r
+         * @param {string} sth   Multi\r
+         *                       line description.\r
          */\r
         const fn = ( lorem, sit ) => {}\r
       `,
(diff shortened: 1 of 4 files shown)


renatho:
It would be awesome if we bring the features we introduced in the copy of [alignTransform](https://github.com/gajus/eslint-plugin-jsdoc/blob/master/src/alignTransform.js) to `comment-parser` and maintain it only there. cc @syavorsky.

Currently, I'm implementing a prettier plugin to use in a project along with this eslint rule (probably I'll make it only for the project in this first time), and I'm importing the alignTransform from `eslint-plugin-jsdoc`.

brettz9:
You might want to ping @syavorsky on the `comment-parser` repo, as not sure he's listening here. That sounds fine to me, though if he isn't inclined to expand it in `comment-parser` and wishes to leave it to user-space, we could add it within [jsdoccomment](https://github.com/es-joy/jsdoccomment) where I've been gathering some essentials or utilities for dealing with JSDoc that can be of wider interest beyond the specific goals of this project (e.g., our `comment-parser` JSDoc-tag-aware parsing code is now obtained from there).

So, does this look like the right fix and do you think we need that carriage return code in both places I mentioned?

test/rules/assertions/checkLineAlignment.js
```diff
@@ -1057,6 +1057,36 @@ export default {
       const fn = ( lorem, sit ) => {}
       `,
     },
+    {
+      code: `\r
+        /**\r
+         * Function description.\r
+         *\r
+         * @param {string} lorem Description.\r
+         * @param {int} sit Description multi words.\r
```
renatho, line 1066: I'd suggest at least test manually (if you don't want to expand this test), making one of the descriptions multiline, and add an empty line and then a `@return` tag. Just to make sure we didn't miss any situation. :)

renatho (approved):
> we could add it within jsdoccomment where I've been gathering some essentials or utilities for dealing with JSDoc that can be of wider interest beyond the specific goals of this project

Makes sense! It'd be nice!!

> In particular, if the delim-setting code is unreachable, then I imagine we don't really need to expand on this line

Makes sense too! Maybe if we don't have plans to bring that back to `comment-parser`, we could remove [that if](https://github.com/gajus/eslint-plugin-jsdoc/blob/5b9f379c04bd22db4475e22619bdefcf0efa2961/src/alignTransform.js#L96-L99) and update the line you commented:

```js
tokens.postDelimiter = nothingAfter.delim ? '' : space(spacings.postDelimiter);
```

But this suggestion wouldn't be needed for this PR since it's just a cleanup. I added a comment in the code with a suggestion just to test a situation. If that's working well, this is good for me! =)

brettz9:
Test added, so merging to get a fix released sooner.

Let us know if you decide to file a PR for `comment-parser` or likewise with `jsdoccomment`.

PR #775, opened 2021-08-11
[tag-lines] Schema fix for tags

DMartens:
The current schema of `tag-lines` for the property tags is incorrect as `patternProperties` is inside `properties` instead of only `patternProperties`.
As such only `tags.patternProperties` is checked and not as wanted `tags.*`.


brettz9:
Good find! Thanks!

PR #797, opened 2021-10-21
Fix a few small typos

timgates42:


brettz9:
Strange, as it seems like the README should be incorporating the updated rule docs and test case change. Nevermind then, I can take a look, thanks!

PR #808, opened 2021-11-26
refactor: remove the `lodash` dependency

jaydenseric:
This significantly reduces the package install size, as `lodash` has a 1.35 MB install size:

https://packagephobia.com/result?p=lodash@4.17.21

While the current total install size for `eslint-plugin-jsdoc` is 4.75 MB:

https://packagephobia.com/result?p=eslint-plugin-jsdoc@37.0.3

Also, presumably the native methods (`Array.filter`, `Array.flatMap`, `Array.some`, etc.) perform better than `lodash` functions.

Some of the more complicated helpers have been refactored to minimal dependencies, most or which were able to be dev dependencies to correctly ensure a minimal production installation for consumers.


src/rules/requireDescription.js
```diff
@@ -1,6 +1,13 @@
-import _ from 'lodash';
 import iterateJsdoc from '../iterateJsdoc';
 
+const checkDescription = (description) => {
+  return description
+    .trim()
+    .split('\n')
+    .filter(Boolean)
+    .length;
+};
```
jaydenseric, line 9: This was moved to the outer scope at the recommendation of a lint error.

src/rules/requireDescriptionCompleteSentence.js
```diff
@@ -164,7 +165,7 @@ export default iterateJsdoc(({
 
   const abbreviationsRegex = abbreviations.length ?
     new RegExp('\\b' + abbreviations.map((abbreviation) => {
-      return _.escapeRegExp(abbreviation.replace(/\.$/ug, '') + '.');
+      return escapeStringRegexp(abbreviation.replace(/\.$/ug, '') + '.');
```
jaydenseric, line 168: Maybe look into if that extra replacing is necessary.

src/jsdocUtils.js
```diff
@@ -312,11 +316,15 @@ const getPreferredTagName = (
...
   // https://github.com/eslint/eslint/issues/13289
   // https://github.com/gajus/eslint-plugin-jsdoc/issues/537
-  const tagPreferenceFixed = _.mapKeys(tagPreference, (_value, key) => {
-    return key.replace('tag ', '');
-  });
+  const tagPreferenceFixed = Object.fromEntries(
+    Object
+      .entries(tagPreference)
+      .map(([key, value]) => {
+        return [key.replace('tag ', ''), value];
```
jaydenseric, line 323: Technically this would be more correct, I didn't include it in the PR to avoid scope creep.

```suggestion
        return [key.replace(/^tag /u, ''), value];
```

jaydenseric (commented):
Sorry about the earlier seperate comment notifications; I thought I was queueing them for a single comment review but accidentally was spamming them as I went.

brettz9:
Thanks! Looks good. Hope to have energy to review in the next few days... Re: scope creep, I don't mind a few small things--better to get things right, imo.

brettz9:
I pushed a couple minor tweaks and also fixed testing and linting. Could you handle the removal of the remaining `lodash` import in `src/rules/noUndefinedTypes.js`?

brettz9:
Actually, I think I've handled that last file all right--maybe just review all of my changes to see if you feel all right to move forward

src/rules/requireJsdoc.js
```diff
@@ -140,11 +139,15 @@ const OPTIONS_SCHEMA = {
...
+  if (context.options[0] && option in context.options[0] &&
+    // Todo: boolean shouldn't be returning property, but tests currently require
+    (typeof context.options[0][option] === 'boolean' ||
+    key in context.options[0][option])
+  ) {
+    return context.options[0][option][key];
   }
 
-  return context.options[0][option][key];
+  return baseObject.properties[key].default;
```
jaydenseric, line 150: I'm not aware of the context for these changes, but trust it makes sense.

jaydenseric (commented):
Thanks for the polish, LGTM!

src/rules/requireJsdoc.js
brettz9: Yeah, although it is likely a bug (some issues reported which may suggest it), just using the boolean value of `context.options[0][option]` when found to be a boolean didn't work either, so would need investigation and a separate PR.

What this fix does is just restore the lodash behavior of avoiding erring when trying `in` checking on `context.options[0][option]` (as it will do when it is a boolean).

Trott:
I'm excited to see this change has landed. Is there any chance that a new release will be published soon? I'm asking because we're [waiting for this change to make it into a release over in Node.js core](https://github.com/nodejs/node/pull/41027#issuecomment-982358209).

brettz9:
@Trott : As per the automated release, this is now released in v37.1.0. And thanks for the heads up on its use on the Node.js codebase! Hope it works well for y'all...

PR #814, opened 2021-12-15
WIP: Add new assertions that should be valid with a fixed parser

### LINKED ISSUE #813: jsdoc/valid-types Rule Errors when namepath contains `readonly` (all lowercase)
## Expected behavior

The following should be valid and should not generate an error.

```js
/**
 * @param {object} params
 * @param {boolean} params.readonly
 */
```

## Actual behavior

```js
/**
 * @param {object} params
 * @param {boolean} params.readonly
 */
```

Errors with: "Syntax error in namepath: params.readonly  jsdoc/valid-types".

## ESLint Config

```js
module.exports = {
  root: true,
  extends: ["plugin:jsdoc/recommended"],
  rules: {
    "jsdoc/valid-types": "error",
  },
};
```

## ESLint sample

```js
/**
 * @param readonly
 */
```

```js
/**
 * @param {boolean} readonly
 */
```

```js
/**
 * @param {object} params
 * @param {boolean} params.readonly
 */
```

## Environment

- Node version: 16.6.1
- ESLint version: 8.4.1
- `eslint-plugin-jsdoc` version: 37.2.0

## Possible Cause

I believe the cause of this is due to the following PR: https://github.com/jsdoc-type-pratt-parser/jsdoc-type-pratt-parser/pull/98

Although the cause may be due to the parser (I filed an issue there too (https://github.com/jsdoc-type-pratt-parser/jsdoc-type-pratt-parser/issues/104)), I'm thinking it would be good to improve the tests here to help catch some of these errors.

dhurlburtusa:
Here are some assertions that would be useful to have to ensure the underlying parser is behaving correctly.

Closes #813 .

### DIFF
--- a/test/rules/assertions/validTypes.js
+++ b/test/rules/assertions/validTypes.js
@@ -736,6 +736,72 @@ export default {
     {
       code: `
           /**
+           * @param {{ foo: string }} bar
+           */
+          function quux() {
+
+          }
+      `,
+    },
+    {
+      code: `
+          /**
+           * @param {{ "foo": string }} bar
+           */
+          function quux() {
+
+          }
+      `,
+    },
+    {
+      code: `
+          /**
+           * @param {{ readonly "foo": string }} bar
+           */
+          function quux() {
+
+          }
+      `,
+      settings: {
+        jsdoc: {
+          mode: 'typescript',
+        },
+      },
+    },
+    {
+      code: `
+          /**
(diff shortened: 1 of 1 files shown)


brettz9:
As mentioned in #813 , can just track with jsdoc-type-pratt-parser, as generally keeping up to date, and would like to pare down our issue and PR lists.

PR #821, opened 2021-12-20
chore: update Husky 7 config

nschonni:
Husky 7 uses hook files in the `.husky` folder that are currently ignored.
Unignored the files and used `npx mrm@2 lint-staged` to rescaffold the `lint-staged` setup


brettz9:
I like the Husky changes, but:

1. I don't see `mrm` anywhere in your PR
2. I prefer our lint-args script as the ESLint cache can, after changes to the ESLint config, for example, still remain stale. Our lint-arg script ensures lint-staged passes on only the files of relevance rather than any others (the eslint cache may do that, but still has to check across all files).

nschonni:
> 1. I don't see `mrm` anywhere in your PR

Sorry, that was unclear. It's the recommended scaffold tool from `lint-staged` in their docs https://www.npmjs.com/package/lint-staged but it's just run as a CLI to setup, rather than as another tool that's added

nschonni:
OK, I've dropped the `--cache` flag from the eslint call. `lint-staged` only passes the touched files, and skips them if they're ignored by the `.eslintignore`. I think the extra `.eslintignore` script you have is a neat idea, but I think it might be missing a call to `git add .`. `lint-staged` will auto-add the files it touches, but won't for the extra files that your larger call flags

package.json
```diff
@@ -57,10 +57,8 @@
   ],
   "license": "BSD-3-Clause",
   "lint-staged": {
-    "./*.js": "npm run lint-arg --",
     ".eslintignore": "npm run lint",
-    "src/**/*.js": "npm run lint-arg --",
-    "test/**/*.js": "npm run lint-arg --"
+    "*.js": "eslint --fix"
```
brettz9, line 61: I'd prefer not to have these changes. I like to refer to `lint-arg` so we can keep the actual linting directives in one place (including calling `--report-unused-disable-directives`).

I am ok, however, with reducing it to a single line pointing to `npm run lint-arg --` if @gajus doesn't mind us linting all files by default, leaving the ignore file to settle which ones should be ignored.

brettz9 (changes requested):

package.json
nschonni: Yeah, `--report-unused-disable-directives` is useful, so I added it to the call, and just dropped the `lint-arg` script

brettz9: > `lint-staged` only passes the touched files, and skips them if they're ignored by the `.eslintignore`.

Ah yes, sorry, forgot you were actually not getting rid of lint-staged either.

> I think the extra `.eslintignore` script you have is a neat idea, but I think it might be missing a call to `git add .`. `lint-staged` will auto-add the files it touches, but won't for the extra files that your larger call flags

The idea is merely that we should relint everything, and although `lint-staged` doesn't pass in all files as part of its file args (such as we leverage with our `lint-arg` calls), in running our `npm run lint` with `.eslintignore`, that script should, I believe, cause *all* files to be re-linted due to the nature of `npm run lint` running everything.

package.json
```diff
@@ -57,10 +57,8 @@
   ],
   "license": "BSD-3-Clause",
   "lint-staged": {
-    "./*.js": "npm run lint-arg --",
     ".eslintignore": "npm run lint",
-    "src/**/*.js": "npm run lint-arg --",
-    "test/**/*.js": "npm run lint-arg --"
+    "*.js": "eslint --report-unused-disable-directives --fix"
```
brettz9, line 61: As mentioned, I'd like to keep `lint-arg` so that we have all of our lint commands together, better ensuring they remain in sync.

package.json
nschonni: Right, I was mixing up the current call to `lint` instead of `lint-fix`, which might have changes files to re-add 😄

package.json
brettz9: We do have `lint-fix`. I'm not always a fan of auto-fixing, but if you want to switch to `lint-fix`, I'm ok with that.

package.json
brettz9: Ah. Well, I guess that's not a bad idea, so welcome to apply `lint-fix` for the ignore file too.

But as mentioned in the review comments, I do want to reference npm scripts.

package.json
brettz9: Probably good in a group project as some folks making minor fixes might get frustrated and not know what to do with some linting directives.

nschonni: `lint-fix` for this command doesn't make sense, because `lint-staged` individually passes in the files, rather than calling the global lint job. I did go back to using your `lint-arg` consistently across all the calls though

brettz9: > `lint-fix` for this command doesn't make sense, because `lint-staged` individually passes in the files, rather than calling the global lint job. 

Sure, but we could try adding some `lint-*` script for the sake of `.eslintignore` which did run things globally with a fix. It appears `lint-staged` nowadays auto-runs `git add .` so maybe it would work? Anyways, it's not really that important here.

> I did go back to using your `lint-arg` consistently across all the calls though

Cool. I also really like what you did with `lint-fix` and `lint`.

nschonni: Do you want me to try and add the `lint-fix` in this PR, or keep it for later?

brettz9 (changes requested):
Besides my comments just now, it all looks good. Just make sure there aren't any other directories or files to ignore. Should I think just be 14 warnings only after linting.

.eslintignore
```diff
@@ -0,0 +1 @@
+.ncurc.js
```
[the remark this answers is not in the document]
gajus: I usually just `ln -s .gitignore .eslintignore` and then commit the symbolic link.

.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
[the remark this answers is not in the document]
nschonni: When Husky installs now, it puts its internal scripts in the `.husky/_/` folder. To ignore churn on scripts that shouldn't be checked in, it adds the extra `.gitignore`

.eslintignore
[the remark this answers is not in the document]
nschonni: The link is an interesting approach, but I don't know if you the untracked files are always the ones you don't want linted. I updated this to just `dist/`, since that's the only folder that has JS files in that list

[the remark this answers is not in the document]
brettz9: Normally at least `coverage` should be included, but as we're not using the HTML or iconv reporter (and hopefully keeping 100% coverage), we shouldn't need to dig into such reports with the text ones being sufficient.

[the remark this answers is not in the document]
brettz9: I also meant to have `!.*.js` added to ensure linting the hidden files since they are ignored in ESLint by default.

We should at least add that one and also `.ignore` so people have a place to put JavaScript tests and junk without getting them linted.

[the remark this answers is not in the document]
nschonni: OK, haven't really come across the `.ignore` before, but I've added it in with the other one. Add some comments so it's more explicit for the next person. I haven't come across the `coverage` folder in this repo, just the `.nyc_output`, but that is all JSON files

package.json
brettz9: I can merge what you have and see if it works. `lint-staged` warns against using `git add`, but will see if it can work.

.eslintignore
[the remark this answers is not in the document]
brettz9: Re: `.ignore`, `.idea` would work fine too--just to have a place for experiments or such. Re: `coverage` you'll only see it if you run against an html or iconv reporter, but as we don't have scripts for that and don't need it, that's fine.

PR #824, opened 2021-12-21
ci: Convert Travis to GitHub Actions

nschonni:
Will still need some work in setting up the secrets to allow the autopublish. Took baseline from htmlthint, but skipped the CodeCov setup, since it isn't used here


.github/workflows/ci.yml
```diff
@@ -0,0 +1,107 @@
...
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node }}
+
+      - name: 📥 Download deps
+        run: npm install
+
+      - name: Install ESLint 7
+        if: matrix.eslint == 7
+        run: npm install --no-save eslint@7 eslint-config-canonical@24.4.4
```
nschonni, line 69: Don't really understand the need for the `eslint-config-conanical` here, but it was used in Travis, and 2 tests failed without it here https://github.com/nschonni/eslint-plugin-jsdoc/runs/4591072042?check_suite_focus=true#step:8:63

nschonni (commented):
Run can be seen here https://github.com/nschonni/eslint-plugin-jsdoc/actions/runs/1605404810
The build badge on the readme should be updated too, but can happen afterwards from getting the markup from the Actions tab

.github/workflows/ci.yml
brettz9: The `check-examples` rule (which has had to be disabled for ESLint 8 tests, but should work for ESLint 7 still) has a behavior of checking the actual current package's eslintrc (i.e., that of `eslint-plugin-jsdoc`).

I don't recall, however, whether it fails because of more recent expectations breaking the tests (I don't think so) or because of some ESlint interaction problem. Sorry not of more help.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,107 @@
...
+  build:
+    name: Build
+    runs-on: ${{ matrix.os }}
+    needs: lint
+    strategy:
+      fail-fast: false
+      matrix:
+        node:
+          - 12
+          - 14
```
brettz9, line 39:
```suggestion
          - 14.17.0
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,107 @@
...
+
+  build:
+    name: Build
+    runs-on: ${{ matrix.os }}
+    needs: lint
+    strategy:
+      fail-fast: false
+      matrix:
+        node:
+          - 12
```
brettz9, line 38:
```suggestion
          - 12.22.0
```

.github/workflows/ci.yml
brettz9: Could you take a look at the current `.travis.yml` file--which resolves some issues we were having--and see about applying the conditional logic there?

`eslint-config-canonical` can now be pegged to `32.1.0`, but we also need to peg `eslint-plugin-mocha` to version `9` because it [no longer supports Node 12](https://github.com/lo1tuma/eslint-plugin-mocha/releases/tag/10.0.0)

I also added the `--engine-strict` check on the latest Node 16 version to ensure at least our development tools could all run on latest Node, but I couldn't add this to the earlier versions because as mentioned the latest linting tools won't work on Node 12. That also means we don't want linting to run in every environment--Node 16 should be enough (and also `npm run build`) is only really needed in Node 16 I think, as the critical item to check is that the tests work in all environments.

brettz9 (changes requested):
Thanks for your work on this. Now that we've resolved in Travis which versions were needed to help us resume passing tests, do you think you could update the conditionals with the matrix to use the same basic logic in the Actions file? Thanks!

.github/workflows/ci.yml
brettz9: You can see if these version bumps are really necessary to pass the tests, but IIRC some downstream tools at one point at least were going to need this version, or possibly some of the linting or testing wouldn't work without the changes.

brettz9:
@nschonni : Think you might be able to take a look? I think GH Actions is probably the way to go for auto-publishing. Another repo I worked with did this recently at https://github.com/ianpogi5/mocha-badge-generator/pull/40/files if it may help to see a pattern.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,107 @@
...
+        os:
+          - ubuntu-latest
+          - macos-latest
+          - windows-latest
+        eslint:
+          - 7
+          - 8
+
+    steps:
+      - name: 🛑 Cancel Previous Runs
```
gajus, line 51: Can we remove this?

Even if there is a previous build, I would still expect them to be released separately.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,107 @@
...
+on:
+  pull_request:
+  push:
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: 🛑 Cancel Previous Runs
```
gajus, line 12: Can we remove this?

Even if there is a previous build, I would still expect them to be released separately.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,107 @@
...
+      - name: 🏗 Build
+        run: npm run build
+        env:
+          NODE_ENV: production
+
+      - name: 🚢 Release
+        uses: cycjimmy/semantic-release-action@v2
+        with:
+          extra_plugins: |
+            @semantic-release/git
```
gajus, line 100: I believe this is missing npm plugin if we want to release the package.

nschonni:
@gajus please just push the changes you want

gajus:
Closing in favor of #943

PR #825, opened 2021-12-21
test: move Mocha and NYC config from CLI

nschonni:
Cleans up the scripts so all the CLI arguments are in the config sections in package.json. This allows the CLI to be overloaded while calling the scripts if needed.
Also allows IDEs to hook into with the debugger, since the config is loaded by Mocha.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
   },
   "nyc": {
     "branches": 100,
-    "check-coverage": false,
+    "check-coverage": true,
     "exclude": [
       "src/rules/checkExamples.js"
     ],
--- a/package.json
+++ b/package.json
@@ -61,6 +61,14 @@
   },
   "main": "./dist/index.js",
   "name": "eslint-plugin-jsdoc",
+  "mocha": {
+    "require": [
+      "@babel/register"
+    ],
+    "reporter": "dot",
+    "recursive": true,
+    "timeout": 12000
+  },
   "nyc": {
     "branches": 100,
     "check-coverage": false,
@@ -73,6 +81,7 @@
     ],
     "instrument": false,
     "lines": 100,
+    "reporter": "text-summary",
     "require": [
       "@babel/register"
     ],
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -100,10 +109,9 @@
     "lint-arg": "eslint --report-unused-disable-directives",
     "lint-fix": "npm run lint-arg -- --fix .",
     "prepare": "husky install",
-    "test": "cross-env BABEL_ENV=test nyc --reporter text-summary mocha --reporter dot --recursive --require @babel/register --timeout 12000",
-    "test-cov": "cross-env BABEL_ENV=test nyc mocha --reporter dot --recursive --require @babel/register --timeout 12000",
-    "test-index": "cross-env BABEL_ENV=test mocha --recursive --require @babel/register --reporter progress --timeout 12000 test/rules/index.js",
-    "test-no-cov": "cross-env BABEL_ENV=test mocha --reporter dot --recursive --require @babel/register --timeout 12000"
+    "test": "mocha",
+    "test-cov": "cross-env BABEL_ENV=test nyc npm run test",
```
brettz9, line 113: The purpose of `test-cov` is to see a table of line numbers indicating failing coverage, so we should override the reporter in that case (otherwise, I kept `test` in case @gajus didn't want that large summary always showing).

```suggestion
    "test-cov": "cross-env BABEL_ENV=test nyc --reporter text npm run test",
```

FWIW, I switched from `progress` to `dot` for the Mocha reporter in anticipation of ESLint supporting parallel tests for Mocha (still hasn't happened, but should be coming) as `progress` doesn't work with parallel ones. We could use it for `test-index`, but I guess we may as well be uniform.

brettz9 (changes requested):
Very good idea and nicely done. I just have one addition as a commit suggestion. Given the problems of `cross-env` with our set-up, this PR might also help us resume testing other Node versions by calling without `cross-env` from within CI.

package.json
nschonni: Should this be set in the config like the others, instead of the current `text-summary`?

brettz9: I got the impression @gajus might like the minimalism of `text-summary` normally. But I'd like to have one script to get the full results.

nschonni: OK, but with this change, it doesn't run coverage as the default anymore, only for this task, so it's already minimal as the default, if you wanted to have the verbose for this coverage task

brettz9: I mean I think he prefers the minimal coverage reporter--just to see the summary that coverage is 100% or whatever.

package.json
```diff
@@ -74,6 +82,7 @@
     ],
     "instrument": false,
     "lines": 100,
+    "reporter": "text-summary",
```
nschonni, line 85: @brettz9 what I mean is that this would be the place to change it, since no other test script displays coverage at all

brettz9: I see. But if we are going to keep accommodating both of our requirements, we need an extra one. That's why I had a `no-cov` before and a `test` that did do coverage and `test-cov` which did a more complete report.

That reminds me, are you ok for me to reenable coverage checks, @gajus ?

brettz9:
I've reverted to using `test-no-cov` so that `test` still does a form of coverage, and also kept out use of the `text` reporter on `test-cov`, but otherwise the gist of your PR is preserved in terms of keeping the config moved to `package.json` and with the scripts interdependent. I've also resumed with `check-coverage` as we have earlier Node tests passing again, and I believe it should be good to enforce as before. Thanks!

PR #833, opened 2022-02-02
feat: add `sort-tags` rule

brettz9:
Welcome to give feedback on:
# Tag grouping and order
# Other aspects (options, tests)

To-dos:
- [ ] finish implementation including fixer
- [ ] ensure coverage

Although it might be nice in some ways to try to promote a pseudo-standard in adding this rule to `recommended`, I figure it may be too annoying, especially for those who don't use `--fix` to have to remember the complex sorting algorithm.

- feat: add `sort-tags` rule

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -589,5 +589,6 @@ selector).
 {"gitdown": "include", "file": "./rules/require-throws.md"}
 {"gitdown": "include", "file": "./rules/require-yields.md"}
 {"gitdown": "include", "file": "./rules/require-yields-check.md"}
+{"gitdown": "include", "file": "./rules/sort-tags.md"}
 {"gitdown": "include", "file": "./rules/tag-lines.md"}
 {"gitdown": "include", "file": "./rules/valid-types.md"}
--- a/.README/rules/sort-tags.md
+++ b/.README/rules/sort-tags.md
@@ -1,0 +1,198 @@
+### `sort-tags`
+
+Sorts tags by a specified sequence according to tag name.
+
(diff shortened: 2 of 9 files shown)


Shinigami92:
I will sponsor one-time `10$` to @brettz9, when this feature is released and stable, via GitHub-Sponsor.

brettz9:
Hi @Shinigami92 ... Want to take it for a test drive? Passing the test cases now...

Note that it probably won't work correctly for these non-typical styles, however:

```js
/** @tagOnZeroLine
 * @tagOnFirstLine
 */
```

```js
/**
 * @tagOnFirstLine
 * @tagOnSecondLineWithEndOnSameLine */
```

Shinigami92:
Both cases would be "invalid" JSDoc anyways :stuck_out_tongue_winking_eye: 
So I would be totally fine if this gets ignored

I will try to find a time today to test it out and will report back

brettz9:
I think technically they should be supported as the only requirements with JSDoc as far as I know are just the two-asterisks and two only beginning, but probably pretty uncommon, and rather just get this out the door now anyways once tested a little.

Shinigami92:
Seems that the rule is working quite well
Only the message when it's triggered is very long

brettz9:
Could you show how the long message appears in your UI? Or is it the command line?

Shinigami92:
> Could you show how the long message appears in your UI? Or is it the command line?

CLI

brettz9:
Ok, plan to release then, as I think it'd be a helpful thing to do if possible so people can trace what to do without having to dig through the config. If people only use a small set of tags, they can supply a shortened list too.

PR #851, opened 2022-03-12
adding recommended-all-contexts config

Zamiell:


src/index.js
```diff
@@ -48,65 +48,92 @@ import sortTags from './rules/sortTags';
...
+ * apply in certain contexts, like when they are attached to a function. In this stricter config, the
+ * rules will apply all of the time.
+ */
+const recommendedAllContextsRules = {};
+for (const [key, value] of Object.entries(recommendedRules)) {
+  const ruleProperties = {
+    contexts: ['any'],
+  };
+  recommendedAllContextsRules[key] = [value, ruleProperties];
+}
```
brettz9, line 116: I believe this will create problems given that each rule has a schema and if `contexts` is not on that schema when used, an error will be reported to the user. I think you will need to add only to rules which explicitly allow `contexts`.

brettz9 (changes requested):
Looking good, but see the comment.

PR #856, opened 2022-03-26
perf: improve performance for rules

golopot:
Benchmark on the [ESLint project](https://github.com/eslint/eslint) with `time npx eslint .` (real time) :

|  | before | after |
| --- | -- | -- |
| run 1 | 47.7s | 31.0s |
| run 2 | 47.8s | 31.2s |
| run 3 | 48.7s | 31.2s |


brettz9:
Good find, thanks!

PR #865, opened 2022-04-07
removed regextra library to reduce dependencies

ChrisDryden:
We currently have three seperate security advisories for this package due to the sub packages of regextras. I noticed that there is only one use of the regextras package and it could be replaced with only a few lines and would signifigantly reduce the size of the package and the impact of the downstream dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ os: linux
 dist: xenial
 language: node_js
 node_js:
-  - "12.22.0"
   - "14.17.0"
   - "16"
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ language: node_js
 node_js:
   - "14.17.0"
   - "16"
+  - "17"
 install:
   - npm install --legacy-peer-deps
 before_script: >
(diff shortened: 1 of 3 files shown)


brettz9:
Ok, but can we see about using [matchAll](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll) instead of `exec`? Node 12 is being end-of-lifed by the end of this month, so I think we can move to drop support ourselves for Node 12 now, and the feature would then be available for use by our targeted versions.

ChrisDryden:
Sounds good to me, thanks for the fast response! Updating it now

brettz9:
and so that the tests pass, feel free to add the change to `engines` and travis...

ChrisDryden:
Removed the node12 support and made the matchAll changes. One of the four builds is failing and it appears to be unrelated to the changes, would it be possible to take a quick look or rerun the CI?

brettz9:
I've filed https://github.com/gajus/eslint-plugin-canonical/issues/4 as such a fix should get our tests working again.

brettz9:
I discovered that we could probably live with `matchAll` with Node 12, as it was only buggy apparently by not throwing when a non-global regex was supplied, but I still think we may as well, allowing us to use other Node 14+-only features now.

Thanks for the PR!

PR #866, opened 2022-04-07
Triggering CI build without changes

ChrisDryden:
Verifying that changes in other PR already existed before


brettz9:
Yeah, this was due to a breaking change introduced in `eslint-plugin-canonical` in a minor release update. Good check. Let's discuss in #865.

PR #867, opened 2022-04-07
fix: add `enableFixer` option to `require-example`

arinthros:
Currently there is no way to disable the fixer for `require-example`. This adds that option following the pattern from other rules.

- Add test coverage for option
- Add option to rule
- Wrap fixer logic in a conditional

### DIFF
--- a/.README/rules/require-example.md
+++ b/.README/rules/require-example.md
@@ -47,6 +47,11 @@ A value indicating whether getters should be checked. Defaults to `false`.
 
 A value indicating whether setters should be checked. Defaults to `false`.
 
+##### `enableFixer`
+
+A boolean on whether to enable the fixer (which adds an empty `@example` block).
+Defaults to `true`.
+
 #### Fixer
 
 The fixer for `require-example` will add an empty `@example`, but it will still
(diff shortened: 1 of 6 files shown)


arinthros:
I manually added the option to the readme, but it was deleted when manually running the readme generator. Not 100% sure how to get it to fully show up in the readme - but I'm happy to rebase with the change if someone can point it out.

brettz9:
Thanks for the PR!

For changes to the README, you need to modify the files in `/.README/`. I've added a commit to this PR to clarify how that is done.

If you want to make the manual addition, I can review, and then squash all these commits. (It's fine we include this slightly irrelevant addition I've made.)

arinthros:
Thanks for the quick response. I made the documentation updates and rebased them into my initial commit to keep the history clean. Please let me know if there's anything else I need to do!

.README/rules/require-example.md
```diff
@@ -57,7 +62,7 @@ report a missing example description after this is added.
 |Context|`ArrowFunctionExpression`, `FunctionDeclaration`, `FunctionExpression`; others when `contexts` option enabled|
 |Tags|`example`|
 |Recommended|false|
-|Options|`exemptedBy`, `exemptNoArguments`, `avoidExampleOnConstructors`, `contexts`|
+|Options|`exemptedBy`, `exemptNoArguments`, `contexts`, `checkConstructors`, `checkGetters`, `checkSetters`, `enableFixer`|
```
arinthros, line 65: This removes a deprecated option and adds the options that were missing from this table.

brettz9: Oh, good find. Doing a search, I also see `avoidExampleOnConstructors` is within the main README. Would you mind updating that too?

arinthros: Done!

brettz9:
Thanks for finding this and helping out! Should be posting a release here shortly...

PR #875, opened 2022-04-18
fix(`require-param-type`): report column=1

chearon:
Hello, I've been dealing with an annoying bug using NeoVim + LSP + eslint. If I'm typing a function parameter from scratch, like `/** @param {nu`, this plugin rightfully reports an error (`require-param-type`) since I haven't finished typing the type.

But it reports column number = 0 which causes NeoVim's LSP client to crash, interrupting my typing.

Presumably it would be impossible to error on an empty line, so I'm hoping adding column = 1 when the upstream parser doesn't report a column number is okay as a quick fix. I'm sure there is a better way to get the real column number but I don't have time to dig any deeper right now.

Not sure if other rules are affected or not, this is the only one I ran into.

Thanks!

---

`x.js`
```js
/**
 * @param {a xxx
 */
```

`yarn eslint x.js` before commit
```
2:0   error  Missing JSDoc @param "" type     jsdoc/require-param-type
```

`yarn eslint x.js` after commit
```
2:1   error  Missing JSDoc @param "" type     jsdoc/require-param-type
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -14868,6 +14868,13 @@ function quux (foo) {
 // Message: Missing JSDoc @param "foo" type.
 
 /**
+ * @param {a xxx
+ */
+function quux () {
+}
+// Message: Missing JSDoc @param "" type.
+
+/**
  * @param foo
  */
 function quux (foo) {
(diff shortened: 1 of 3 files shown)


brettz9:
LGTM--thanks for the report!

PR #878, opened 2022-04-20
chore(.husky): remove unnecessary .gitignore

Fdawgs:
`.husky/.gitignore` is no longer required as of [husky 7.0.0](https://github.com/typicode/husky/releases/tag/v7.0.0), and the devDep is currently at ^7.0.4.


brettz9:
Thanks for the PR!

brettz9:
For some reason, Atom's `.gitignore` is showing things as though there has been a change.

image

Even if this is an Atom bug, I find it strange that this particular commit would cause problems. Is the underscore directory no longer needed too, do you know?

Fdawgs:
`.husky/_/` is created when `husky install` is ran and contains the script that the hooks use, so it is needed yes.
Seems like an Atom bug. 😞

PR #899, opened 2022-08-02
fix(`check-tag-names`): work with the constructor tag

### LINKED ISSUE #898: Use of tagNamePreference breaks all @constructor tags
## Expected behavior

I would like to use the `jsdoc/check-tag-names` with `tagNamePreference` in my settings and still use tags like `@constructor` even though they are named like `Object.prototype` properties.

## Actual behavior

```
Invalid JSDoc tag (preference). Replace "constructor" JSDoc tag with "function Object() { [native code] }"  jsdoc/check-tag-names
```

## ESLint Config

In `package.json`, here is my `eslintConfig`:
```json
{
  "plugins": [
    "jsdoc"
  ],
  "rules": {
    "jsdoc/check-tag-names": "error"
  },
  "settings": {
    "jsdoc": {
      "tagNamePreference": {
        "returns": "return"
      }
    }
  }
}
```

## ESLint sample

```js
/**
 * @constructor
 */
function Test() {
  this.works = false;
}
```

## Environment

- Node version: v18.4.0
- ESLint version: v8.21.0
- `eslint-plugin-jsdoc` version: v39.3.4

This is different than the issue described in #785, where the user was trying to provide a `tagNamePreference` for the `@constructor` tag.  In my case, I'm trying to provide a `tagNamePreference` for any tag (not only one that conflicts with `Object.prototype` names), and this makes it so I can no longer use the `@constructor` tag.

tschaub:
This makes it so the `check-tag-names` rule works when JSDoc tag names match `Object.prototype` properties (like `@constructor`).

Fixes #898.

### DIFF
--- a/test/jsdocUtils.js
+++ b/test/jsdocUtils.js
@@ -10,6 +10,12 @@ describe('jsdocUtils', () => {
         it('returns the primary tag name', () => {
           expect(jsdocUtils.getPreferredTagName({}, 'jsdoc', 'return')).to.equal('returns');
         });
+        it('works with the constructor tag', () => {
+          expect(jsdocUtils.getPreferredTagName({}, 'jsdoc', 'constructor')).to.equal('class');
+        });
+      });
+      it('works with tags that clash with prototype properties', () => {
+        expect(jsdocUtils.getPreferredTagName({}, 'jsdoc', 'toString')).to.equal('toString');
       });
       it('returns the primary tag name', () => {
         expect(jsdocUtils.getPreferredTagName({}, 'jsdoc', 'returns')).to.equal('returns');
--- a/README.md
+++ b/README.md
@@ -4500,6 +4500,15 @@ function quux (foo) {}
 /** @jsxImportSource preact */
 /** @jsxRuntime automatic */
 // Message: Invalid JSDoc tag name "jsx".
+
+/**
+ * @constructor
+ */
+function Test() {
+  this.works = false;
+}
+// Settings: {"jsdoc":{"tagNamePreference":{"returns":"return"}}}
+// Message: Invalid JSDoc tag (preference). Replace "constructor" JSDoc tag with "class".
 ````
 
 The following patterns are not considered problems:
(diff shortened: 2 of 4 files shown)


test/jsdocUtils.js
```diff
@@ -10,6 +10,12 @@ describe('jsdocUtils', () => {
         it('returns the primary tag name', () => {
           expect(jsdocUtils.getPreferredTagName({}, 'jsdoc', 'return')).to.equal('returns');
         });
+        it('works with the constructor tag', () => {
+          expect(jsdocUtils.getPreferredTagName({}, 'jsdoc', 'constructor')).to.equal('class');
```
tschaub, line 14: Without the `src` change, `jsdocUtils.getPreferredTagName()` returns `[Function Object]` in this case.

test/jsdocUtils.js
```diff
@@ -10,6 +10,12 @@ describe('jsdocUtils', () => {
...
+        });
+      });
+      it('works with tags that class with prototype properties', () => {
+        expect(jsdocUtils.getPreferredTagName({}, 'jsdoc', 'toString')).to.equal('toString');
```
tschaub, line 18: Without the `src` change, this test fails with `expected [Function Object] to equal 'class'`.

test/jsdocUtils.js
```diff
@@ -10,6 +10,12 @@ describe('jsdocUtils', () => {
         it('returns the primary tag name', () => {
           expect(jsdocUtils.getPreferredTagName({}, 'jsdoc', 'return')).to.equal('returns');
         });
+        it('works with the constructor tag', () => {
+          expect(jsdocUtils.getPreferredTagName({}, 'jsdoc', 'constructor')).to.equal('class');
+        });
+      });
+      it('works with tags that clash with prototype properties', () => {
+        expect(jsdocUtils.getPreferredTagName({}, 'jsdoc', 'toString')).to.equal('toString');
```
tschaub, line 18: Without the `src` change, this test fails with `expected [Function toString] to equal 'toString'`.

PR #902, opened 2022-08-09
Fix README link in require-returns-type.md

bluefuton:
The README link to "ASTs and Selectors" in rules/require-returns-type.md does not currently work.

This pull request updates the link to use the correct URL and anchor.

### DIFF
--- a/.README/rules/require-returns-type.md
+++ b/.README/rules/require-returns-type.md
@@ -14,7 +14,7 @@ for finding function blocks not attached to a function declaration or
 expression, i.e., `@callback` or `@function` (or its aliases `@func` or
 `@method`) (including those associated with an `@interface`).
 
-See the ["AST and Selectors"](#eslint-plugin-jsdoc-advanced-ast-and-selectors)
+See the ["AST and Selectors"](../#ast-and-selectors)
 section of our README for more on the expected format.
 
 |||


brettz9:
Thanks for the PR. However, I am unclear where you are seeing breakage (the link works for me when viewing both on the main GitHub repo README and on npmjs.com) and in what environment your change fixes things.

We have a script in `src/generateReadme.js` which compiles `./README/README.md` (which grabs in the items in `./README/rules/`), and although we could support allowing relative paths in source and then converting this to our GitHub/npmjs.com-friendly paths, I'm curious to see where you saw the issue.

It appears that GitHub (with or without the Octolinker add-on) turns the relative paths into links which at least lead to the right page (though the anchor doesn't get preserved there). Is that where you saw the issue?

bluefuton:
Hi @brettz9 👋 Yes, I saw it on Github. I believe I searched for the rule name and landed up here:

https://github.com/gajus/eslint-plugin-jsdoc/blob/master/.README/rules/require-returns-type.md

The link there doesn't currently take you to the correct page - you remain on `require-returns-type.md`.

brettz9:
Cool. I've added b029eb985d6dfa86a6fb09e8c0bb195b70d315a4 which should address. Note that it still is only linking to the main README source rather than generated file, but that seems more consistent with what these files are anyways. The real authority should be the final generated file, but this was a good idea to make it work on the individual pages.

Issue #701, opened 2021-03-01
Add a config with ALL options turned on ("plugin:jsdoc/all")


mkvlrn:
## Motivation

Enabling all rules without having to add each rule or the ones missing from the recommended config

## Current behavior

There is no extensible "all" config like there is a recommended one "plugin:jsdoc/recommended"

## Desired behavior

I would like to be able to enable all rules just by extending such configuration

## Alternatives considered

Adding each missing rule from the recommended config or creating/publishing a custom config would do, but I rather stick to the tried and true, reliable configs out there

brettz9:
This is a good idea, but we should note that we will really need environment-specific "all" configs, e.g., we wouldn't want `no-types` applying for regular jsdoc.

Note that this issue also ties into #615 where environment-specific "recommended" configs were requested. So this can I think essentially become an issue requesting environment-specific "all" configs.

PR #906, opened 2022-08-19
feat: Add `require-throws-type` rule

jomarquez21:


brettz9:
Thanks for the PR.

One can currently get this functionality by using:

```js
      settings: {
        jsdoc: {
          structuredTags: {
            throws: {
              required: [
                'type',
              ],
            },
          },
        },
      },
```

I think this approach keeps things simpler (one might also want to ensure `@typedef` has a type too.

We might offer a config in the future along the lines of #701 which offered a mode with such reasonable stricter default settings.

Issue #894, opened 2022-06-18
`check-line-alignment` with "always" breaks with @param, @return, and no types


TGCrystal:
This issue is very similar to #891, except it applies when both the param and return (or returns) tags are present. Rather than aligning the return description with the param description, extra whitespace before the return description is added.

## Expected behavior
No errors are given.

## Actual behavior
The following error is given for either test case below:
```
Expected JSDoc block lines to be aligned  jsdoc/check-line-alignment
```

## ESLint Config

```js
{
    "plugins": [
        "jsdoc"
    ],
    "rules": {
        "jsdoc/check-line-alignment": [
            2,
            "always"
        ]
    }
}
```

## ESLint sample

```js
/**
 * Function description.
 *
 * @param lorem Description.
 * @param sit   Description multi words.
 * @returns     Return description.
 */
const fn = ( lorem, sit ) => {};
```
and
```js
/**
 * Function description.
 *
 * @param lorem Description.
 * @param sit   Description multi words.
 * @return      Return description.
 */
const fn = ( lorem, sit ) => {};
```

## Environment

- Node version: 16.13.2
- ESLint version 8.17.0
- `eslint-plugin-jsdoc` version: 39.3.3

brettz9:
Hmm. I see that the current behavior is to recommend:

```js
      /**
       * Function description.
       *
       * @param    lorem Description.
       * @param    sit   Description multi words.
       * @returns        Return description.
       */
      const fn = ( lorem, sit ) => {};
```

There is at least one extra space after the tag names for sure (as the default is to have just one sapace). But I can see why a project might want the names in its own column.

@renatho : Do you have thoughts about the intended result for this? The WordPress standard you pointed to, in their extended example (one has to click to expand it), under https://developer.wordpress.org/coding-standards/inline-documentation-standards/javascript/#functions , does not seem to even push the `@return` description past the `@param` names to align with the `@param` description, so I'm not sure that we want to be pinning to this standard (or, if we are, we are far from it given that that spec states "*Related* comments should be spaced so that they align to make them more easily readable.").

@TGCrystal is requesting a behavior still different from your approach or the WordPress spec, in aligning across different tag groups, but not apparently wishing to add space for a name column when no name is present. I'm not sure such a suggestion fits with the rule given it differs both from current behavior and the WordPress spec, but it'd be nice if you could respond to it.

[...]

PR #908, opened 2022-08-25
fix: Improve Windows contributing experience

TGCrystal:
While working on #894, I ran into the following issues while developing on Windows:
1. The eslint config had the `linebreak-style` rule enabled and set to LF. This can run into false positives with git's `autocrlf` config property.
2. The `npm run create-readme` command was giving the error `No assertions available for rule "checkAccess".` 

These issues were addressed respectively by the following changes in this PR:
1. Use .gitattributes to force all files to be given LF line endings. They can still show up locally as CRLF, but anything committed will be LF. This can be seen with the `git ls-files --eol` command, the line ending for the index is prefaced by `i/` and the line ending for the workspace is prefaced by `w/`. ~~The only potential issue with this is that tests with `\r` before the newline now have that character removed in the generated README.md, and this seems to be the cause for the changes in the readme's diff in this PR.~~ (see note below) It doesn't seem like they show up in the readme on GitHub anyways though, so I don't think that should be an issue.
2. The `path.resolve` function will generate the path with `\`, as Windows normally does. Even on Windows, though, `glob.sync` expects `/` so replacing all `\` with `/` seems to have fixed the issue.

Note: The initial .gitattributes change seemed to be causing differing behavior on Windows vs Linux (tested in WSL). The `run create-readme` command was actually removing the `\r` from tests in Windows, but not in Linux. There was no visual difference, though, so I figured the best way to fix this would be to have the readme just replace the carriage returns with `\\r` so that they will now be explicitly shown on the README. Tested on Windows and WSL again, and now they consistently generate the same output.


brettz9:
Thanks for the PR!

PR #918, opened 2022-10-18
fix: support node 19

### LINKED ISSUE #917: Support node 19
## Motivation

Support node 19 now that it has been officially released by node, see:
* https://nodejs.org/en/download/current/
* https://github.com/nodejs/Release#release-schedule

## Current behavior

We have automated unit tests that run across all officially supported node versions, `maintenance`, `active lts`, and `current`. Today the ones running against `current` started failing when trying to install dependencies with the following error:
```
error eslint-plugin-jsdoc@39.3.13: The engine "node" is incompatible with this module. Expected version "^14 || ^16 || ^17 || ^18". Got "19.0.0"
```

## Desired behavior

Installing `eslint-plugin-jsdoc` when using node 19 should work.

## Alternatives considered

We could use something like `yarn install --ignore-engines` as a workaround but that opens us up to compatibility errors that will only show up at runtime.

RodEsp:
Fixes #917


brettz9:
Thanks for the PR!

PR #931, opened 2022-10-31
Skip root checking

### LINKED ISSUE #863: allow means to skip roots in`check-types`
## Motivation

To allow the likes of this standard JSDoc approach, which allows for documentation per parameter:

```js
/**
 * @param {object} cfg
 * @param {boolean} cfg.isChecked
 * @param {string} cfg.path
 */
```

...but without complaining about "object" not being a specific type if one's `preferredTypes` normally discourage the use of this object. One can already exempt it within a `@typedef`, but one shouldn't wish to normally exempt it in `@param`--only if it is the "root" param. This would therefore need to be a new kind of `exemptedBy`, allowing probably for an object and possibly AST `comment` and `context`.

## Current behavior

Using `object` in a root param when it is prevented in `preferredTypes` will err.

## Desired behavior

Don't err when using `object` in a root param.

## Alternatives considered

N/A

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/113278050-allow-means-to-skip-roots-in-check-types?utm_campaign=plugin&utm_content=tracker%2F23037809&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23037809&utm_medium=issues&utm_source=github).
</bountysource-plugin>

brettz9:
feat(`check-types`): add `skipRootChecking` option to `preferredTypes` setting; fixes #863

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -391,6 +391,8 @@ The format of the configuration is as follows:
       - a string type to be preferred in its place (and which `fix` mode
         can replace)
       - `false` (for forbidding the type)
+    - an optional key `skipRootChecking` (for `check-types`) to allow for this
+      type in the context of a root (i.e., a parent object of some child type)
 
 Note that the preferred types indicated as targets in
 `settings.jsdoc.preferredTypes` map will be assumed to be defined by
--- a/README.md
+++ b/README.md
@@ -480,6 +480,8 @@ The format of the configuration is as follows:
       - a string type to be preferred in its place (and which `fix` mode
         can replace)
       - `false` (for forbidding the type)
+    - an optional key `skipRootChecking` (for `check-types`) to allow for this
+      type in the context of a root (i.e., a parent object of some child type)
 
 Note that the preferred types indicated as targets in
 `settings.jsdoc.preferredTypes` map will be assumed to be defined by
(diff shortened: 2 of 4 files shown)


golopot:
I don't see why `skipRootChecking` is needed. This allows you to use `object` in a "root" parameter, and `Object` in other places. But why do you need it?

brettz9:
The default is `object` actually; that example prevents `object` except as a root (since it is listed in `preferredTypes` and without a replacement).

The reason one may wish to do so is because "object" can be an overly generic type. The same is true with arrays or functions, but those can be controlled with the `preferredTypes` setting. But prior to this PR one could not make an exception and allow "object" as the root where the object typically is nothing more or need be nothing more than a plain object. And since JSDoc requires that roots be listed, one has to use workarounds (e.g., defining via `typedef`) just to be able to use this common pattern without warnings—if one still wants to get warnings about overly generic "object"'s.

brettz9:
The default is `object` actually; that example prevents `object` except as a root (since it is listed in `preferredTypes` and without a replacement).

The reason one may wish to prevent `object` normally is because "object" can be an overly generic type. The same is true with arrays or functions, but those can be controlled with the `preferredTypes` setting. But prior to this PR one could not make an exception and allow "object" as the root where the object typically is nothing more or need be nothing more than a plain object. And since JSDoc requires that roots be listed, one has to use workarounds (e.g., defining via `typedef`) just to be able to use this common pattern without warnings—if one still wants to get warnings about overly generic "object"'s.

golopot:
I get it. It make sense

PR #933, opened 2022-11-14
chore(.husky): shell path and arg updates

Fdawgs:
v8.0.0 of Husky fixed shell path and args for newly created git hooks, but doesn't update existing ones, so have done that here.

See https://github.com/typicode/husky/pull/1051 and https://github.com/typicode/husky/pull/1132


brettz9:
Thanks for the PR!

PR #951, opened 2023-01-20
allow `*` and `any` types to avoid reporting of missing return

### LINKED ISSUE #950: require-returns-check doesn't handle `*` types correctly
## Expected behavior
I think this one might be a bug. I expect this to be ok:

```js
/*
 * @param {boolean} bar A fun variable.
 * @returns {*} Anything at all!
 */
function foo( bar ) {
  if ( bar ) {
    return functionWithUnknownReturnType();
  }
}
```

 The `*` type means the function can return anything, which IMO should include undefined/void/etc. This can be fixed with `{*|undefined}`, but that's pretty weird :P

## Actual behavior
This error happens: `JSDoc @returns declaration present but return expression not available in function`

## ESLint Config

```js
'jsdoc/require-returns-check': 'error'
```

## ESLint sample

See above

## Environment

- Node version: happens in both 16 and 18
- ESLint version: 8.14.0
- `eslint-plugin-jsdoc` version: 39.6.6 (latest)

brettz9:
fix(`require-returns-check`, `require-yields-check`); allow `*` and `any` types to avoid reporting of missing return; fixes #950

### DIFF
--- a/README.md
+++ b/README.md
@@ -17875,7 +17875,7 @@ function quux () {
 }
 
 /**
- * @returns {*} Foo.
+ * @returns {SomeType} Foo.
  */
 const quux = () => foo;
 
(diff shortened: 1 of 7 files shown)


noahtallen:
Thank you for the quick fix!!

Issue #779, opened 2021-08-22
jsdoc/require-description-complete-sentence: Autofix adds dot in wrong place


ilyub:
## Expected behavior

Should add dot here:
```ts
/**
 * Description.
 *
 * @param str - String.
 */
export function f(str: string): void {}
```

## Actual behavior

Adds dots like this:
```ts
/**.
 * .
 * .
 * .
 * Description
 *
 * @param str - String.
 */
export function f(str: string): void {}
```

## ESLint Config

```ts
"jsdoc/require-description-complete-sentence": "warn"
```

## ESLint sample

```ts
/**
 * Description
 *
 * @param str - String.
 */
export function f(str: string): void {}
```

```ts
// Format JS code here
```
## Environment

"eslint-plugin-jsdoc": "^36.0.7",

brettz9:
Here's it in test case form for any who might be able to take a look at this:

```js
    {
      code: `
      /**
       * Description
       *
       * @param str - String.
       */
      export function f(str: string): void {}
      `,
      errors: [
        {
          line: 3,
          message: 'Sentence must end with a period.',
        },
      ],
      output: `
      /**
       * Description.
       *
       * @param str - String.
       */
      export function f(str: string): void {}
      `,
      parser: require.resolve('@typescript-eslint/parser'),
      parserOptions: {
        sourceType: 'module',
      },
    },
```

vxcamiloxv:
same issue with `jsdoc/require-description-complete-sentence`  and auto --fix `eslint-plugin-jsdoc` version 36.1.0 and `eslint` version 7.32.0

RedGuy12:
```js
/**
 * ...
 * @returns {Promise<
 * 	(T extends true ? import("./src/types").ShallowGallery : import("./src/types").Gallery)[]
 * >}
 *   - Array of gallery data.
 */
```
This rule autofixes as
```js
// /**
//  * .
//  * ...
//  * @returns {Promise<
//  * 	(T extends true ? import("./src/types").ShallowGallery : import("./src/types").Gallery)[]
//  * >}
//  *   - Array of gallery data.
//  */
```

maribethb:
I am also having issues with long record types and this rule. For example

```
/**
 * Render the playground layout.
 * @param {HTMLElement} container The container to render the playground in.
 * @return {{
 *     blocklyDiv: HTMLElement,
 *     minimizeButton: HTMLElement,
 *     monacoDiv: HTMLElement,
[...]

PR #959, opened 2023-01-27
fix: complete sentence period placement

le-elvidge:
This fixes https://github.com/gajus/eslint-plugin-jsdoc/issues/779. 

The previous implementation was matching an empty line following the paragraph and performing a regex replacement on it. Added an extra test to verify the fix.


brettz9:
Thanks very much for the fix!

PR #964, opened 2023-01-28
feat: `wrapIndent` option

### LINKED ISSUE #715: Google style indentation
## Motivation

Indentation matters for readability, especially having a consistent indentation of documentation. I want to be able to use google style indentation for `@parameter` and `@return` as an example.

## Current behavior

This might be part of `check-indentation` but I could not get it work with it.

## Desired behavior

I want this to be invalid:
```js
  /**
   * Do something.
   *
   * @param {string} fooA - Some text
   * @param {string} fooB - If defined, it foos
   * a lot all day long.
   */
```

I want this to be valid (with the new rule)
```js
  /**
   * Do something.
   *
   * @param {string} fooA - Some text
   * @param {string} fooB - If defined, it foos
   *     a lot all day long. // This has been double-indented (4 spaces, we use 2 spaces for indents)
   */
```

This is called google style indent. It makes it easier to read the documentation.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/98492595-google-style-indentation?utm_campaign=plugin&utm_content=tracker%2F23037809&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23037809&utm_medium=issues&utm_source=github).
</bountysource-plugin>

brettz9:
fixes #715

Also adds missing docs for `preserveMainDescriptionPostDelimiter`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2031,10 +2031,10 @@ for example.
 This rule allows one optional string argument. If it is `"always"` then a
 problem is raised when the lines are not aligned. If it is `"never"` then
 a problem should be raised when there is more than one space between each
-line's parts. Defaults to `"never"`.
+line's parts. If it is `"any"`, no alignment is made. Defaults to `"never"`.
 
-Note that in addition to alignment, both options will ensure at least one
-space is present after the asterisk delimiter.
+Note that in addition to alignment, the "never" and "always" options will both
+ensure that at least one space is present after the asterisk delimiter.
 
 After the string, an options object is allowed with the following properties.
 
@@ -2042,8 +2042,7 @@ After the string, an options object is allowed with the following properties.
 <a name="eslint-plugin-jsdoc-rules-check-line-alignment-options-3-tags"></a>
 ##### <code>tags</code>
 
-Use this to change the tags which are sought for alignment changes. *Currently*
-*only works with the "never" option.* Defaults to an array of
+Use this to change the tags which are sought for alignment changes. Defaults to an array of
 `['param', 'arg', 'argument', 'property', 'prop', 'returns', 'return']`.
 
 <a name="user-content-eslint-plugin-jsdoc-rules-check-line-alignment-options-3-customspacings"></a>
@@ -2060,10 +2059,24 @@ An object with any of the following keys set to an integer. Affects spacing:
 
 If a spacing is not defined, it defaults to one.
 
+<a name="user-content-eslint-plugin-jsdoc-rules-check-line-alignment-options-3-preservemaindescriptionpostdelimiter"></a>
+<a name="eslint-plugin-jsdoc-rules-check-line-alignment-options-3-preservemaindescriptionpostdelimiter"></a>
+##### <code>preserveMainDescriptionPostDelimiter</code>
+
+A boolean to determine whether to preserve the post-delimiter spacing of the
+main description. If `false` or unset, will be set to a single space.
+
+<a name="user-content-eslint-plugin-jsdoc-rules-check-line-alignment-options-3-wrapindent"></a>
+<a name="eslint-plugin-jsdoc-rules-check-line-alignment-options-3-wrapindent"></a>
+##### <code>wrapIndent</code>
+
+The indent that will be applied for tag text after the first line.
+Default to the empty string (no indent).
+
 |||
 |---|---|
 |Context|everywhere|
-|Options|(a string matching `"always" or "never"` and optional object with `tags` and `customSpacings`)|
+|Options|(a string matching `"always"`, `"never"`, or `"any"` and optional object with `tags`, `customSpacings`, `preserveMainDescriptionPostDelimiter`, and `wrapIndent`)|
 |Tags|`param`, `property`, `returns` and others added by `tags`|
 |Aliases|`arg`, `argument`, `prop`, `return`|
 |Recommended|false|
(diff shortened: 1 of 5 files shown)


brettz9:
@thernstig Often one can just give the GitHub URL with the branch after the hash, but sometimes, as in this repo, it is necessary to remove the dist files from gitignore and add them to its own branch (since the `dist` files would not be available to GitHub otherwise). To simplify things, I've created such a branch for you, so you can use this in your `package.json` without needing to create your own branch with the dist files:

```json
"devDependencies": {
  "eslint-plugin-jsdoc": "git@github.com:brettz9/eslint-plugin-jsdoc.git#wrapIndent-with-dist"
}
```

thernstig:
Cheers @brettz9 - maybe you could consider adding npm preleases to the workflow to avoid having to create custom git branches. Not sure of the pros and cons, but that is a topic for another thread.

README.md
```diff
@@ -2794,6 +2839,53 @@ const fn = ( lorem, sit ) => {};
...
+
+/**
+ * Function description.
+ *
+ * @param {string} lorem Description.
+ * @param {int}    sit   Description multi
+ *                         line with asterisks.
+ */
+const fn = ( lorem, sit ) => {}
+// "jsdoc/check-line-alignment": ["error"|"warn", "always",{"wrapIndent":"  "}]
```
thernstig, line 2867: Does it make sense indenting like this when using "always"? Maybe users want to indent from the start?

thernstig (commented):
I tried with this:

```json
"jsdoc/check-line-alignment": ["warn", "never", { "wrapIndent": "   " }],
```

And it changed this:
```js
 * @property {boolean} tls_verify_client_certificate - Whether our API should
 *   enable TLS client authentication
```

To this:
```js
 * @property {boolean} tls_verify_client_certificate - Whether our API should
 *      enable TLS client authentication
```

Is that intentional? I would think it should remove current indentation and just add what is configured in our case i.e. the end result should have been:

```js
 * @property {boolean} tls_verify_client_certificate - Whether our API should
 *    enable TLS client authentication
```

README.md
brettz9: If users will go to the trouble of wanting alignment of columns, I figure they will wish to continue this as text wrapping across the columns would be distracting.

brettz9:
Thanks for checking it out. Hope to investigate soon, as that does not seem intentional.

README.md
thernstig: @brettz9 yes most likely. For me alignment of columns is a bad practice for many reasons (e.g. git history), but each to his own.

brettz9:
@thernstig : I've updated if you might try again. I amended and force-rebased my changes, so you may need to delete your lock file and delete `node_modules` (if `npm update` doesn't work).

thernstig:
@brettz9 with the same config, it did not add any white space at all now. Did your unit tests work?

**Before**
```js
/**
 * @property {object[]} services - Service list include its tags and selected tag.
 *     [{service: 'service-test', tags: ['tag1', 'tag2'], selectedTag: 'tag2'}]
```

**After**
```js
/**
 * @property {object[]} services - Service list include its tags and selected tag.
 * [{service: 'service-test', tags: ['tag1', 'tag2'], selectedTag: 'tag2'}]
```

brettz9:
@thernstig : Yes, I've run the tests. I changed the behavior to remove whitespace (the default) so one now has to set `wrapIndent` if one wishes an indent. But then it will be consistent. Did you have a `wrapIndent` set?

thernstig:
@brettz9 you are correct, apologize for that. I have multiple .eslintrc since it is a monorepo, changed `wrapIndent` in one then checked the results in a file not having that setting 😱 

It works! But there is some kind of bug. After running it the first time with `"jsdoc/check-line-alignment": ["warn", "never", { "wrapIndent": "   " }]` and `eslint --max-warnings=0 . --fix` it changed/fixed to this:

```js
...
 * @property {object} state.openedMetricRow - The metric table row to display in the flyout details
 *    panel.
 * @property {string[]} state.timelineMetrics - The queries of the selected metric table rows.
 * @property {boolean} state.timelineMaxReached - Whether the max amount of timelines have
 *      been reached.
 * @property {object} state.timelinesVisibility - An object containing the timelines
 *      name as key and its visibility as value.
 */
```

Before running `--fix` it looked like:

```js
...
 * @property {object} state.openedMetricRow - The metric table row to display in the flyout details
 * panel.
 * @property {string[]} state.timelineMetrics - The queries of the selected metric table rows.
 * @property {boolean} state.timelineMaxReached - Whether the max amount of timelines have
 *      been reached.
 * @property {object} state.timelinesVisibility - An object containing the timelines
 *      name as key and its visibility as value.
 */
```

The last two properties (it was a long list of properties for this JSDoc) it has extra spacing for the last two.

Running `eslint --max-warnings=0 path/to/file.js --fix` a second time fixed it though.

brettz9:
I think the issue there may relate to ESLint's fixer only running something like 6 times recursively. The standard behavior is for the plugin to only make one fix and let ESLint handle the rest, so I think this isn't really a bug on our part.

colinrotherham:
Would it be good to have an integer config value similar to [`{ paddedIndent: 4 }`](https://github.com/gajus/eslint-plugin-jsdoc#paddedindent)?

thernstig:
> Would it be good to have an integer config value similar to [`{ paddedIndent: 4 }`](https://github.com/gajus/eslint-plugin-jsdoc#paddedindent)?

Would that mean spaces or tabs for indent?

colinrotherham:
It's only spaces from the documentation. I haven't seen tab-indented JSDoc but it might exist!

>##### [`paddedIndent`](https://github.com/gajus/eslint-plugin-jsdoc#paddedindent)
>
>This integer property allows one to add a fixed amount of whitespace at the
>beginning of the second or later lines of the example to be stripped so as
>to avoid linting issues with the decorative whitespace. For example, if set
>to a value of `4`, the initial whitespace below will not trigger `indent`
>rule errors as the extra 4 spaces on each subsequent line will be stripped
>out before evaluation.

thernstig:
Then in my world, having the current solution here is better.

thernstig:
@brettz9 we think this looks good from our side! Feel free to merge, but please check @colinrotherham's comment above if you think the input is valid.

davidmurdoch:
Is there any way to revert to behavior before this commit?

brettz9:
Which behavior do you want to revert to / what was your rule config?

davidmurdoch:
@brettz9

it was just:

```
"jsdoc/check-line-alignment": [
  "error"
],
```

But due to inconsistent JSDoc wrap indentation (in this 230,000+ line repo 😅 ) the updated rule now causes linting to fail, but I don't want to have to `--fix` any of them.

brettz9:
I would suggest then just disabling the rule. I'd accept a PR to make the indent wrapping enforcement optional, but not really inclined to add this myself. I think it generally makes sense for there to be consistent wrapping (or no wrapping) when one is already handling aligning.

davidmurdoch:
Thanks, for a PR, would overloading to allow for `wrapIndent: false` be an acceptable solution? Or do you have something else in mind?

brettz9:
Gajus has expressed a preference for separate config rather than such overloading, so I think it should be as a new boolean config option like `disableWrapIndent`.

thernstig:
> I would suggest then just disabling the rule. I'd accept a PR to make the indent wrapping enforcement optional, but not really inclined to add this myself. I think it generally makes sense for there to be consistent wrapping (or no wrapping) when one is already handling aligning.

Agree, it feels strange using a linter and then still want to allow different indentation in different places. The purpose of a linter is to avoid such discussions so using `wrapIndent` would only help your developers in the end (you just need to agree on what the indent should be).

davidmurdoch:
Yeah, I agree. But in this case the rule changed without a way of reverting to the old behavior. Now it will force exactly the discussions that a linter is supposed to help avoid, and relinting the codebase now will also cause conflict, and if I have to start these discussions about this rule bikeshedding about all the other rules will start again. I'm in a tough place. Haha

thernstig:
Agree with that sentiment as well 😄

davidmurdoch:
I guess that brings us back to the question about making `wrapLines` something that can be opted out of. Would you still approve a PR for it just for backwards compatibility (even though opting out would typically be considered bad practice?).

PR #1007, opened 2023-03-27
feat: added typed option to check-tag-names

### LINKED ISSUE #1005: check-tag-names: targeting JSDocs unnecessary with a type checker (e.g. TypeScript)
## Motivation

There are quite a few tags that are unnecessary when using TypeScript, such as `@abstract`, `@access`/`@private`, and `@class`. Other tags are unnecessary when there's no comment, such as `@this`. It'd be nice to have a lint rule that flags those unnecessary tags.

This actually exists today in [DefinitelyTyped-tools](https://github.com/microsoft/definitelytyped-tools), the repo that contains dtslint rules for [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped). Its tracking issue for porting rules from TSLint to ESLint/typescript-eslint is https://github.com/microsoft/definitelytyped-tools/issues/648. 

## Current behavior

The [`no-types` rule](https://github.com/gajus/eslint-plugin-jsdoc/blob/f1bf68f9e5a8b8632e1978bcf5a3f8123fb64ba3/.README/rules/no-types.md) is similar in intent, but only covers unnecessary types. It doesn't cover other tags.

## Desired behavior

A port of the dtslint rule `no-redundant-jsdoc2`: https://github.com/microsoft/DefinitelyTyped-tools/blob/d5429a4fdb0832e89510ad35e44c7b32fa0149da/packages/dtslint/src/rules/noRedundantJsdoc2Rule.ts

## Alternatives considered

We could keep the implementation inside DefinitelyTyped-tools... but the plumbing in this project around JSDoc parsing & reporting is quite useful! ❤️ - 

This seems like a universally applicable rule for TypeScript projects. I'd want to add it to the `recommended-typescript(-error)` presets.

JoshuaKGoldberg:
Fixes #1005.

I believe this recreates all the logic from the [`no-redundant-jsdoc2` rule](https://github.com/microsoft/DefinitelyTyped-tools/blob/d5429a4fdb0832e89510ad35e44c7b32fa0149da/packages/dtslint/src/rules/noRedundantJsdoc2Rule.ts) in [DefinitelyTyped-tools](https://github.com/microsoft/DefinitelyTyped-tools).

~Does not add auto-fixing yet, as I had a hard time figuring out what the right util is. 😅~ ✅


src/rules/checkTagNames.js
```diff
@@ -54,12 +110,83 @@ export default iterateJsdoc(({
...
+    return true;
+  };
+
+  const tagRedundantlyChecksParamOrReturnTyped = (jsdocTag) => {
+    if (!typedTagsParamReturn.has(jsdocTag.tag)) {
+      return false;
+    }
+
+    if (!jsdocTag.description) {
+      report(`'@${jsdocTag.tag}' without a description is redundant when using a type system.`, null, jsdocTag);
```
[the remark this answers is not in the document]
brettz9: In the case of `report`, the second argument should be a fixer function. It normally is used with a fixer argument which can then have its `replaceText` method called to perform string manipulations.

However, I would instead recommend using `utils.reportJSDoc()` which allows one to manipulate the AST, and have it be stringified back.

You can remove a tag by its index with `utils.removeTag()`.

Note also that in some cases, you may need to use the `specRewire` option of `reportJSDoc` if rewriting the tags or otherwise they won't be reflected in the stringification.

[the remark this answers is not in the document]
JoshuaKGoldberg: Thanks for the info. I'm still stuck though 😞 sorry.

d624e9f adds `utils.reportJSDoc` with `specRewire=true` and `utils.removeTag(tagIndex)` usage. But the `output` in the spec is still not being updated:

```plaintext
 1) check-tag-names
       invalid
         
        /** @type {string} */
        let a;
      :

      AssertionError [ERR_ASSERTION]: Output is incorrect.
      + expected - actual

       
      -        /** @type {string} */
               let a;
             
```

Am I missing something still?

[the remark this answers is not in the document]
brettz9: There was an issue with rewiring, but I wouldn't think that should apply here. Could you add this commit to a branch to let me work with it more easily? It might have to do with a bug in `removeTag` in regards to 0-line tags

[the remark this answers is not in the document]
brettz9: I'm guessing that we have two problems: one with removing a tag on what we call the 0th line, and two the desire to have a utility to remove the entire block if it is empty (since the utility to remove a tag should really just remove the tag and leave the block in place, even if that means leaving just `/** */`).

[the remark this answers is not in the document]
brettz9: I've pushed two commits, one of which fixes (and the other is an unrelated nit I wanted to fix in testing)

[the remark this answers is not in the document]
JoshuaKGoldberg: Awesome, thanks! I'll fill out the rest of the fixers and mark this as ready once that's done.

src/iterateJsdoc.js
```diff
@@ -136,6 +136,10 @@ const getUtils = (
       handler();
       const replacement = utils.stringify(jsdoc, specRewire);
 
+      if (!replacement) {
+        return fixer.removeRange(jsdocNode.range);
```
JoshuaKGoldberg, line 140: I changed this in 5c6e9ac42351fe549a310a34894a76cc1ab13fa5 because the `0` index was relative to the source file.

brettz9: Ouch, yeah. I've added f4459a56b0d46bdc7fb62be70239ecc9b52116fc to restore the whitespace removal. How does that look?

JoshuaKGoldberg: Oop sorry forgot to reply here - looks good, thanks!

.README/rules/check-tag-names.md
```diff
@@ -200,6 +200,62 @@ jsxRuntime
...
+name
+namespace
+override
+property
+requires
+static
+this
+```
+
+Additionally, for `@param` and `@return` tags, the rule will flag unnecessary type descriptions (e.g. `@param {string}`).
```
brettz9, line 258: We already have this functionality within our `jsdoc/no-types` rule, so I don't see a reason to add this functionality to this rule.

src/rules/checkTagNames.js
```diff
@@ -54,12 +110,103 @@ export default iterateJsdoc(({
...
+      context.getFilename().endsWith('.d.ts') :
+      Boolean(subNode.declare) || isInAmbientContext(subNode.parent);
+  };
+
+  const tagIsRedundantWhenTyped = (jsdocTag) => {
+    if (!typedTagsUnnecessaryOutsideDeclare.has(jsdocTag.tag)) {
+      return false;
+    }
+
+    if (jsdocTag.description.trim() === 'default' || jsdocTag.name === 'default') {
```
brettz9, line 124: With `default` as a tag description, is this some recognized TS behavior? I'd rather we not add features for idiosyncratic conventions.

src/rules/checkTagNames.js
```diff
@@ -54,12 +110,103 @@ export default iterateJsdoc(({
...
+    if (typedTagsAlwaysUnnecessary.has(jsdocTag.tag)) {
+      reportWithTypeRemovalFixer(`'@${jsdocTag.tag}' is redundant when using a type system.`, jsdocTag, tagIndex, {
+        postTag: '',
+        tag: '',
+      });
+      return true;
+    }
+
+    if (tagIsRedundantWhenTyped(jsdocTag)) {
+      reportWithTypeRemovalFixer(`'@${jsdocTag.tag}' is generally redundant outside of \`declare\` contexts when using a type system.`, jsdocTag, tagIndex);
```
brettz9, line 183: Is it generally redundant or always redundant? If *generally*, when is it not redundant?

test/rules/assertions/checkTagNames.js
```diff
@@ -674,6 +972,95 @@ export default {
...
+        /** @template name */
+        let a;
+      `,
+      options: [
+        {
+          definedTags: [
+            'template',
+          ],
+          typed: true,
+        },
```
brettz9, line 1050: This is fine, but I think I'd prefer here the following over `definedTags`:

```js
settings: {
        jsdoc: {
          mode: 'typescript',
        },
      },
```

This should better telegraph the desired approach when using TypeScript (enabling the mode or using the typescript-eslint parser which would also work here). One generally shouldn't need to configure `definedTags`.

src/rules/checkTagNames.js
JoshuaKGoldberg: I'll defer to @sandersn as dtslint person - but I believe this is because people tend to do things like:

```ts
/** @default 0 */
let value = 0;
```

src/rules/checkTagNames.js
JoshuaKGoldberg: Yeah I struggled with this phrasing 😕. In some situations, such as "ambient" (type declaration) contexts (the `"declare"` comment), they are allowed. https://github.com/gajus/eslint-plugin-jsdoc/pull/1007#discussion_r1154616106

test/rules/assertions/checkTagNames.js
JoshuaKGoldberg: Err, this stuff predates me I think. Is there a change to the code added in this PR you're asking for?

src/rules/checkTagNames.js
brettz9: I'm specifically asking about `jsdocTag.description.trim() === 'default'`, i.e.,. where `default` is in the description area like this example:

```js
/** @abstract - default */
```

src/rules/checkTagNames.js
brettz9: The phrasing already mentions being redundant outside of `declare` contexts, so if that is the only case, I'd think we can drop "generally". If it is not redundant in other contexts, I'd hope we could either catch those contexts or if necessary not remove those particular tags.

test/rules/assertions/checkTagNames.js
brettz9: There may be another example like this that we do want to keep for coverage, but I'd rather not add it. Per the diffs, this looks like it is a new addition, esp. since it has `typed: true` on it. I guess you added it because without anything else, the rule will complain, but I'd just prefer we add the `settings` instead.

brettz9: I'm only referring to this example just above:

```js
{
      code: `
        /** @template name */
        let a;
      `,
      options: [
        {
          definedTags: [
            'template',
          ],
          typed: true,
        },
```

brettz9:
By the way, I think we should probably add an `enableFixers` option (default true is ok) since some might not want to lose all the information, e.g., about some code meaning to be abstract, until such time as the user can add the language feature themselves. Of course, if our fixer could handle this on behalf of the user, by adding the missing TS language feature, that would be ideal, but if it is not practical to do so, then some may just wish to get informed about there being a problem.

src/rules/checkTagNames.js
JoshuaKGoldberg: Hmm yeah good point - other than the conversation in https://github.com/gajus/eslint-plugin-jsdoc/pull/1007#discussion_r1154487772, these are all ambient/declaration contexts... I'll remove that for now since I don't see any non-DefinitelyTyped usage for that logic.

src/rules/checkTagNames.js
JoshuaKGoldberg: Yeah good point, I don't know why this would be the case. Pending information from @sandersn / anybody with more context, I'll remove that point. Thanks!

test/rules/assertions/checkTagNames.js
JoshuaKGoldberg: Ah got it, thank you for the explanation. Sorry this took me a bit to parse 😄.

What finally clicked was removing `definedTags: ['template']` from the example. Then we get `Invalid JSDoc tag name "template".`. So, yes, these valid tags should be added to the list of valid ones if we're in `typed: true` mode! 💡

brettz9: Ok, I'd think `settings.jsdoc.mode` set to "typescript" would do the trick, as that should be set to "typescript" anyways for TS users (or auto-detected by use of `@typescript-eslint/parser` as parser, and although you could add the tag for `typed: true`, maybe you should keep `@template` disallowed there without the `mode`/parser in case someone is using some other type annotation system and wants to stick to pure JSDoc syntax?

JoshuaKGoldberg:
> add an `enableFixers` option (default true is ok)

That makes sense to me! If enabling fixers is the default (+1!), can I suggest naming it `disableFixers` and default to `false`? I've always found it confusing to have options that default to a truthy value. It's a ~standard-ish I've been pushing a bit in typescript-eslint.

brettz9:
> > add an `enableFixers` option (default true is ok)
> 
> That makes sense to me! If enabling fixers is the default (+1!), can I suggest naming it `disableFixers` and default to `false`? I've always found it confusing to have options that default to a truthy value. It's a ~standard-ish I've been pushing a bit in typescript-eslint.

That makes sense, but we're currently using `enableFixers`, so I'd like to be consistent across rules and stick with that for now.

JoshuaKGoldberg:
> currently using `enableFixers`

Err, just clarifying, where is this?

brettz9:
Sorry, `enableFixer`. In `check-param-names`, `require-example`, etc.

JoshuaKGoldberg:
It looks good to me! I tried out locally on all in [`DefinitelyTyped/DefinitelyTyped`](https://github.com/DefinitelyTyped/DefinitelyTyped) and it only had one crash: . Otherwise it  it reports about 4,260 violations. I scanned through a few dozen of them at random and they all look like things that should be removed! ✨ 

It'd be useful to have a version of this public so I can really QA. Would you be open to publishing a beta version of `eslint-plugin-jsdoc`? Alternately, are you happy with fast follow PRs if they come up?

test/rules/assertions/checkTagNames.js
```diff
@@ -936,5 +1282,26 @@ export default {
...
+          /**
+           * Converts the images specified by the indices to base64 asynchronously.
+           * @function WebTwain#ConvertToBase64
+           * @returns {boolean}
+           */
+          ConvertToBase64(): boolean;
+        }
+      `,
+      parser: require.resolve('@typescript-eslint/parser'),
+    },
```
JoshuaKGoldberg, line 1305: This is extracted from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c8089e91d1be9af95c604ef234180903593a07b3/types/dwt/v13/index.d.ts. `node` is `null` in the lint rule:

```plaintext
TypeError: Cannot read properties of null (reading 'parent')
Occurred while linting /Users/josh/repos/DefinitelyTyped/types/dwt/v13/index.d.ts:23
Rule: "jsdoc/check-tag-names"
    at tagIsRedundantWhenTyped (/Users/josh/repos/eslint-plugin-jsdoc/dist/rules/checkTagNames.js:66:57)
    at checkTagForTypedValidity (/Users/josh/repos/eslint-plugin-jsdoc/dist/rules/checkTagNames.js:97:9)
    at _default.iterateAllJsdocs (/Users/josh/repos/eslint-plugin-jsdoc/dist/rules/checkTagNames.js:113:18)
    ...
```

JoshuaKGoldberg:
It looks good to me! I tried out locally on all in [`DefinitelyTyped/DefinitelyTyped`](https://github.com/DefinitelyTyped/DefinitelyTyped) and it only had one crash: [gajus/eslint-plugin-jsdoc@`95b212b` (#1007)](https://github.com/gajus/eslint-plugin-jsdoc/pull/1007/commits/95b212bd12be5008c21be0a70107072d5f45f343) -> https://github.com/gajus/eslint-plugin-jsdoc/pull/1007/files#r1162059560. Otherwise it  it reports about 4,260 violations. I scanned through a few dozen of them at random and they all look like things that should be removed! ✨ 

It'd be useful to have a version of this public so I can throw up a DefinitelyTyped PR. Would you be open to publishing a beta version of `eslint-plugin-jsdoc`? Alternately, are you happy with fast follow PRs if they come up?

brettz9:
A fast follow PR should be fine...

brettz9 (approved):

brettz9:
I forgot for us to add the typed option as part of the typescript configs...

brettz9:
added in #1013

JoshuaKGoldberg:
Woop, thanks @brettz9! 🙌

PR #1017, opened 2023-04-12
Sort tags lines

### LINKED ISSUE #782: Line break before @returns
## Motivation

Line breaks matters for readability, especially having a consistent line break style in documentation. I want to be able to enforce line breaks before `@returns` statements.

## Current behavior

There is no configuration for it.

## Desired behavior

I want this to be invalid:
```js
  /**
   * Do something.
   *
   * @param {string} fooA - Some text
   * @returns {string} fooB - Some text
   */
```

I want this to be valid (with the new rule)
```js
  /**
   * Do something.
   *
   * @param {string} fooA - Some text
   *
   * @returns {string} fooB - Some text
   */
```

Or for those who prefer, the other way around (the first example valid, the second invalid).

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/106102213-line-break-before-returns?utm_campaign=plugin&utm_content=tracker%2F23037809&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23037809&utm_medium=issues&utm_source=github).
</bountysource-plugin>

brettz9:
feat: allow specification of tag groups and lines between (and before and after) them and removal of whitespace within them; fixes #782

BREAKING CHANGE:

`tagSequence` array of string tags argument must be specified instead as array of `{tags: []}` objects, with each object representing groups which can gain optional whitespace between them; if none are needed, just put all tags into the array of a single `{tags: []}`


thernstig:
@brettz9 to test this, I would need something similar to what you described here right? So that you build a branch:

https://github.com/gajus/eslint-plugin-jsdoc/pull/964#issuecomment-1410473688

brettz9:
Yes, thank you, and apologies for my absent-mindedness. :-)

You should be able to get it with:

```js
"devDependencies": {
  "eslint-plugin-jsdoc": "git@github.com:brettz9/eslint-plugin-jsdoc.git#sort-tags-lines-dist"
}
```

thernstig:
Some findings. Many might stem from the same root cause?

### Finding 1

This is rather a question. How is this expected to work with `tag-lines` and spacings defined there? It is not easy to grasp. I have currently turned off `tag-lines` completely. But my wish would be to use this with `tag-lines` to be able to specify that I want no empty line breaks at the end of the entire JSDoc block. (And I could not understand the difference between `noEndLines` and `dropEndLines` reading their descriptions, but that is a later issue).

### Finding 2

I find that the entire description for `tagSequence` is somewhat unwieldy. It should possibly be re-written from scratch. Or I could comment on each part individually if you prefer that?

### Finding 3

I expected this to place `@see` with a line before it, since `@param` came before. I am also not sure why it added an empty line at the end. I also did actually did not expect this to rearrange `@see` at all since I thought that should only be done if `alphabetizeExtras` was set?

```json
    "jsdoc/sort-tags": [
      "error",
      {
        "tagSequence": [
          { "tags": ["since", "access"] },
          { "tags": ["class", "augments", "mixes"] },
          { "tags": ["alias", "memberof"] },
          { "tags": ["see", "link", "global"] },
          { "tags": ["fires", "listens"] },
          { "tags": ["param"] },
          { "tags": ["yields"] },
          { "tags": ["returns"] }
        ]
      }
    ],
```

```js
  // BEFORE
  /**
   * Foo
   *
   * @see https://github.com/jaredhanson/passport/blob/v0.4.1/lib/middleware/authenticate.js#L217
   *
   * @param {(
   *  req: express.Request,
   *  done: (error: any, user?: any, info?: any) => void
   * ) => void} verify - callback to excute custom authentication logic
   */
   
   // AFTER
   /**
   * Foo
   *
   * @param {(
   *  req: express.Request,
   *  done: (error: any, user?: any, info?: any) => void
   * ) => void} verify - callback to excute custom authentication logic
   * @see https://github.com/jaredhanson/passport/blob/v0.4.1/lib/middleware/authenticate.js#L217
   *
   */
```

### Finding 4

Notice here how `@tutorial` and `@license` ends up grouped with `@returns` with no line break between. Same as finding 3?

```json
    "jsdoc/sort-tags": [
      "error",
      {
        "tagSequence": [
          { "tags": ["since", "access"] },
          { "tags": ["class", "augments", "mixes"] },
          { "tags": ["alias", "memberof"] },
          { "tags": ["see", "link", "global"] },
          { "tags": ["fires", "listens"] },
          { "tags": ["param"] },
          { "tags": ["yields"] },
          { "tags": ["returns"] }
        ]
      }
    ],
```

```js
// BEFORE
/**
 * Summary. (use period)
 *
 * Description. (use period)
 *
 * @since      1.0.1
 * @access     private
 *
 * @class
 * @mixes    mixin
 *
 * @alias    realName
 * @memberof namespace
 *
 * @see  Function/class relied on
 * @global
 *
 * @tutorial Asd
 * @license MIT
 *
 * @fires   eventName
 * @fires   className#eventName
 * @listens event:eventName
 * @listens className~event:eventName
 *
 * @tutorial Asd
 * @license MIT
 *
 * @yields {string} Yielded value description.
 *
 * @param {string} var1 - Description.
 * @param {string} var2 - Description of optional variable.
 * @param {string} var3 - Description of optional variable with default variable.
 * @param {object} objectVar - Description.
 * @param {string} objectVar.key - Description of a key in the objectVar parameter.
 *
 * @returns {string} Return value description.
 */
   
// AFTER
/**
 * Summary. (use period)
 *
 * Description. (use period)
 *
 * @since      1.0.1
 * @access     private
 *
 * @class
 * @mixes    mixin
 *
 * @alias    realName
 * @memberof namespace
 *
 * @see  Function/class relied on
 * @global
 *
 * @fires   eventName
 * @fires   className#eventName
 * @listens event:eventName
 * @listens className~event:eventName
 *
 * @param {string} var1 - Description.
 * @param {string} var2 - Description of optional variable.
 * @param {string} var3 - Description of optional variable with default variable.
 * @param {object} objectVar - Description.
 * @param {string} objectVar.key - Description of a key in the objectVar parameter.
 *
 * @yields {string} Yielded value description.
 *
 * @returns {string} Return value description.
 * @tutorial Asd
 * @license MIT
 *
 * @tutorial Asd
 * @license MIT
 *
 */
```

### Finding 5

Notice the line breaks after `@license MIT`. Maybe that is intentional though, since I am disabled `tag-lines` and those spaces where there in the original?

```json
    "jsdoc/sort-tags": [
      "error",
      {
        "tagSequence": [
          { "tags": ["since", "access"] },
          { "tags": ["class", "augments", "mixes"] },
          { "tags": ["alias", "memberof"] },
          { "tags": ["see", "link", "global"] },
          { "tags": ["fires", "listens"] },
          { "tags": ["param"] },
          { "tags": ["yields"] },
          { "tags": ["returns"] }
        ]
      }
    ],
```

```js
// BEFORE
/**
 * Summary. (use period)
 *
 * Description. (use period)
 *
 * @since      1.0.1
 * @access     private
 *
 * @class
 * @mixes    mixin
 *
 * @alias    realName
 * @memberof namespace
 *
 * @see  Function/class relied on
 * @global
 *
 * @tutorial Asd
 * @license MIT
 *
 *
 *
 *
 * @fires   eventName
 * @fires   className#eventName
 * @listens event:eventName
 * @listens className~event:eventName
 *
 * @yields {string} Yielded value description.
 *
 * @param {string} var1 - Description.
 * @param {string} var2 - Description of optional variable.
 * @param {string} var3 - Description of optional variable with default variable.
 * @param {object} objectVar - Description.
 * @param {string} objectVar.key - Description of a key in the objectVar parameter.
 *
 * @returns {string} Return value description.
 */
   
// AFTER
/**
 * Summary. (use period)
 *
 * Description. (use period)
 *
 * @since      1.0.1
 * @access     private
 *
 * @class
 * @mixes    mixin
 *
 * @alias    realName
 * @memberof namespace
 *
 * @see  Function/class relied on
 * @global
 *
 * @fires   eventName
 * @fires   className#eventName
 * @listens event:eventName
 * @listens className~event:eventName
 *
 * @param {string} var1 - Description.
 * @param {string} var2 - Description of optional variable.
 * @param {string} var3 - Description of optional variable with default variable.
 * @param {object} objectVar - Description.
 * @param {string} objectVar.key - Description of a key in the objectVar parameter.
 *
 * @yields {string} Yielded value description.
 *
 * @returns {string} Return value description.
 * @tutorial Asd
 * @license MIT
 *
 *
 *
 *
 */
```

### Finding 6

Notice placement of `-other` and how `@listens` gets a line break before it.

```json
    "jsdoc/sort-tags": [
      "error",
      {
        "tagSequence": [
          { "tags": ["since", "access"] },
          { "tags": ["class", "augments", "mixes"] },
          { "tags": ["alias", "memberof"] },
          { "tags": ["see", "link", "global"] },
          { "tags": ["fires", "-other", "listens"] },
          { "tags": ["param"] },
          { "tags": ["yields"] },
          { "tags": ["returns"] }
        ]
      }
    ],
```

```js
// BEFORE
/**
 * Summary. (use period)
 *
 * Description. (use period)
 *
 * @since      1.0.1
 * @access     private
 *
 * @class
 * @mixes    mixin
 *
 * @alias    realName
 * @memberof namespace
 *
 * @see  Function/class relied on
 * @global
 *
 * @fires   eventName
 * @fires   className#eventName
 * @tutorial Asd
 * @license MIT
 *
 * @yields {string} Yielded value description.
 *
 * @listens event:eventName
 * @listens className~event:eventName
 *
 * @param {string} var1 - Description.
 * @param {string} var2 - Description of optional variable.
 * @param {string} var3 - Description of optional variable with default variable.
 * @param {object} objectVar - Description.
 * @param {string} objectVar.key - Description of a key in the objectVar parameter.
 *
 * @returns {string} Return value description.
 */
   
// AFTER
/**
 * Summary. (use period)
 *
 * Description. (use period)
 *
 * @since      1.0.1
 * @access     private
 *
 * @class
 * @mixes    mixin
 *
 * @alias    realName
 * @memberof namespace
 *
 * @see  Function/class relied on
 * @global
 *
 * @fires   eventName
 * @fires   className#eventName
 * @tutorial Asd
 * @license MIT
 *
 * @listens event:eventName
 * @listens className~event:eventName
 *
 * @param {string} var1 - Description.
 * @param {string} var2 - Description of optional variable.
 * @param {string} var3 - Description of optional variable with default variable.
 * @param {object} objectVar - Description.
 * @param {string} objectVar.key - Description of a key in the objectVar parameter.
 *
 * @yields {string} Yielded value description.
 *
 * @returns {string} Return value description.
 */
```

brettz9:
> @brettz9 Some findings. Many might stem from the same root cause?
> 
> ### Finding 1
> This is rather a question. How is this expected to work with `tag-lines` and spacings defined there? It is not easy to grasp. I have currently turned off `tag-lines` completely. But my wish would be to use this with `tag-lines` to be able to specify that I want no empty line breaks at the end of the entire JSDoc block.

The behavior with `tag-lines` is undefined and should probably be explicitly discouraged (unless using "any" as the main option and using `dropEndLines`—see below).

As mentioned, we need some rule to enforce spacing or no spacing before the first tag and after the last tag. This could only be done within `tag-lines` if we would require use of the main string "any", which effectively ignores the main parts of the rule for all tags (but would let the rules' options apply before the first tag and after the final one).

> (And I could not understand the difference between `noEndLines` and `dropEndLines` reading their descriptions, but that is a later issue).

Sure, the definitions could definitely be improved.

`noEndLines` if set to `true` will fail to report any discrepancy between the expected `count` if the tag is a final one.

`dropEndLines` if changed to `true` will report any trailing lines after the final tag. (This option effectively allows a 0 count on the final tag, whatever it is, distinct from the counts used elsewhere. You may wish to use this option with an "any" option along with `sort-tags`, but I think we should really change this to also allow enforcing a positive number of lines on the final tag since `sort-tags` deliberately does not handle this as it is of potentially more broad interest.)
 
> ### Finding 2
> I find that the entire description for `tagSequence` is somewhat unwieldy. It should possibly be re-written from scratch. Or I could comment on each part individually if you prefer that?

I'd prefer if you commented individually. It was a best effort, so it'd be helpful to see which parts needed work.

> ### Finding 3
> I expected this to place `@see` with a line before it, since `@param` came before. I am also not sure why it added an empty line at the end. I also did actually did not expect this to rearrange `@see` at all since I thought that should only be done if `alphabetizeExtras` was set?
> 
> ```json
>     "jsdoc/sort-tags": [
>       "error",
>       {
>         "tagSequence": [
>           { "tags": ["since", "access"] },
>           { "tags": ["class", "augments", "mixes"] },
>           { "tags": ["alias", "memberof"] },
>           { "tags": ["see", "link", "global"] },
>           { "tags": ["fires", "listens"] },
>           { "tags": ["param"] },
>           { "tags": ["yields"] },
>           { "tags": ["returns"] }
>         ]
>       }
>     ],
> ```
> 
> ```js
>   // BEFORE
>   /**
>    * Foo
>    *
>    * @see https://github.com/jaredhanson/passport/blob/v0.4.1/lib/middleware/authenticate.js#L217
>    *
>    * @param {(
>    *  req: express.Request,
>    *  done: (error: any, user?: any, info?: any) => void
>    * ) => void} verify - callback to excute custom authentication logic
>    */
>    
>    // AFTER
>    /**
>    * Foo
>    *
>    * @param {(
>    *  req: express.Request,
>    *  done: (error: any, user?: any, info?: any) => void
>    * ) => void} verify - callback to excute custom authentication logic
>    * @see https://github.com/jaredhanson/passport/blob/v0.4.1/lib/middleware/authenticate.js#L217
>    *
>    */
> ```

The odd thing is that I could not replicate. When I test this, there is no difference in the "after", as it is already following the expected sequence and with the expected line breaks between.

I'm wondering if I may need a minimal test repo to figure out what could be happening.

> ### Finding 4
> ### Finding 5

Let's tackle these after figuring out finding 3.

brettz9:
@thernstig : I did see some code which needed to be safer for counting lines and I've updated the branches including the distribution one, but again, this didn't make a difference in my testing. Still, might be safer for you to do any further testing against the new copy.

thernstig:
Sounds good to handle `tag-lines` later and leave the behavior "undefined" for now.

I tested finding 3 and could not replicate anymore.

But finding 4 still replicates. Try that one.

brettz9:
You can retest if you like, as I think the fix should have fixed Finding 4.

However, I still need to fix Findings 5 and 6, so no need for further exploration of that now.

These are due to the fact that a tag includes all whitespace after it, so it is indeed currently moved with its whitespace. I only have code so far to reduce (or increase) whitespace between formally discovered tag groups, not to drop existing whitespace as part of "-other" (or drop trailing whitespace in the middle of a tag grouping).

thernstig:
@brettz9 I re-tested finding 4 but still same fault, assume maybe you have not re-built the branch. But I can wait till you looked at 5 and 6 and then test them all.

Great job so far, this is going to be very promising!

brettz9:
Hmmm. I did rebuild it, yet I am seeing whitespace added between the `@returns` and `@tutorial`... Any chance you hadn't gotten a fresh copy?

brettz9:
I've updated and created a minimal test repo at https://github.com/brettz9/eslint-plugin-jsdoc-sort-tags-1017 to demo that things are, I think, now working as expected in relation to your findings (3-6).

Finding 5 indeed shows lines at the end, but that is only because I think we should have another compatible rule which handles line breaks after all tags (as well as before them) since that is a useful rule regardless of whether someone wants `sort-tags`. So I've avoided handling the final tag in `sort-tags`.

thernstig:
@brettz9 verified it. All works great!

brettz9:
What rule config were you using?

brettz9:
I'm unfortunately not able to replicate using the likes of:

```js
      {
        "tagSequence": [
          { "tags": ["public"] },
          { "tags": ["param"] },
        ]
      }
```

thernstig:
```js
    "jsdoc/sort-tags": [
      "error",
      {
        "tagSequence": [
          { "tags": ["since", "access"] },
          { "tags": ["class", "augments", "mixes"] },
          { "tags": ["alias", "memberof"] },
          { "tags": ["public", "protected", "private", "override"] },
          { "tags": ["override", "async"] },
          { "tags": ["see", "link", "global"] },
          { "tags": ["param"] },
          { "tags": ["yields"] },
          { "tags": ["returns"] },
          { "tags": ["fires", "-other", "listens"] }
        ]
      }
    ],
```

Maybe it only happens when being run on many files for some reason?

brettz9:
Yeah, I'm not sure. There have occasionally been issues, e.g., with VSC, where recursive fixing can lead to relics like this, but there have been issues trying to replicate there too.

I agree that we should first have the rule for removing whitespace at the end (and beginning). `dropEndLines` in `tag-lines` mostly does the trick, but it would ideally allow a fixed number of lines, e.g., 0+ (and then also before tags).

.README/rules/sort-tags.md
```diff
@@ -8,24 +8,32 @@ Sorts tags by a specified sequence according to tag name.
 
 ##### `tagSequence`
 
-An array of tag names indicating the preferred sequence for sorting tags.
+An array of tag group objects indicating the preferred sequence for sorting tags.
+
+Each item in the array can be an object with a `tags` property set to an array
```
thernstig, line 13: "can be an object" -> "Should be an object" (or "is an")?

.README/rules/sort-tags.md
```diff
@@ -8,24 +8,32 @@ Sorts tags by a specified sequence according to tag name.
...
+of tag names.
 
 Tag names earlier in the list will be arranged first. The relative position of
 tags of the same name will not be changed.
 
+Earlier groups will also be arranged before later groups, but with the added
```
thernstig, line 19: "that additional line breaks may be added between (or before or after) such groups." -> "that one line break will be added between groups"?

It is just one line break, and it will be always be added in its current form, right?

.README/rules/sort-tags.md
```diff
@@ -8,24 +8,32 @@ Sorts tags by a specified sequence according to tag name.
...
+feature that additional line breaks may be added between (or before or after)
+such groups.
+
 Tags not in the list will be sorted alphabetically at the end (or in place of
```
thernstig, line 23: I think this paragraph could be rephrased. Is this an improvement to get better flow?

"Tag names not in the list will be grouped together at the end. The psuedo-tag `-other` can be used to place them anywhere else if wanted. The tags will be placed in their order of appearance, or alphabetized if the `alphabetizeExtras` is enabled, see more below about that option."

.README/rules/sort-tags.md
```diff
@@ -8,24 +8,32 @@ Sorts tags by a specified sequence according to tag name.
...
 the pseudo-tag `-other` placed within `tagSequence`) if `alphabetizeExtras` is
 enabled and in their order of appearance otherwise (so if you want all your
 tags alphabetized, supply an empty array with `alphabetizeExtras` enabled).
 
-Defaults to the array below.
+Defaults to the array below (noting that it is just a single tag group with
```
thernstig, line 28: noting that -> note that

thernstig (commented):
I added some minor, opinionated changes to the wording here. Nothing huge, you be the judge if they are improvements or not.

Other than that, I think this is good to merge.

brettz9:
Plan to still get to your doc suggestions, but wanted to mention I've filed #1018 to add the discussed support to `tag-lines`.

.README/rules/sort-tags.md
brettz9: Sure, changed to "should be"

.README/rules/sort-tags.md
brettz9: The default is 1 line, but it can be controlled with `linesBetween`. I realized the latter was missing from the docs, so just added that too.

I'm not sure what you mean "always be added in its current form". If you mean that the line break follows the asterisk and spacing of the surrounding context, then yes.

.README/rules/sort-tags.md
brettz9: Yes, thanks, I have borrowed from this in a new commit.

ST-DDT:
I ran it locally for one of my projects and it seems to erase some parts from the documentation after the second newline:

**Example 1**

````ts
// Original
   * @param options.mode The mode to generate the birthdate. Supported modes are `'age'` and `'year'` .
   *
   * There are two modes available `'age'` and `'year'`:
   * - `'age'`: The min and max options define the age of the person (e.g. `18` - `42`).
   * - `'year'`: The min and max options define the range the birthdate may be in (e.g. `1900` - `2000`).
   *
   * Defaults to `year`.
   *
   * @example
// Formatted
  * @param options.mode The mode to generate the birthdate. Supported modes are `'age'` and `'year'` .
   *
   * There are two modes available `'age'` and `'year'`:
   * - `'age'`: The min and max options define the age of the person (e.g. `18` - `42`).
   * - `'year'`: The min and max options define the range the birthdate may be in (e.g. `1900` - `2000`).
   * @example
// Lost parts
   *
   * Defaults to `year`.
   *
````

**Example 2**

````ts
// Original
   * @example
   * enum Color { Red, Green, Blue }
   * faker.helpers.enumValue(Color) // 1 (Green)
   *
   * enum Direction { North = 'North', South = 'South'}
   * faker.helpers.enumValue(Direction) // 'South'
   *
   * enum HttpStatus { Ok = 200, Created = 201, BadRequest = 400, Unauthorized = 401 }
   * faker.helpers.enumValue(HttpStatus) // 200 (Ok)
   *
   * @since 8.0.0
// Formatted
   * @example
   * enum Color { Red, Green, Blue }
   * faker.helpers.enumValue(Color) // 1 (Green)
   *
   * enum Direction { North = 'North', South = 'South'}
   * faker.helpers.enumValue(Direction) // 'South'
   * @since 8.0.0
// Lost parts
   *
   * enum HttpStatus { Ok = 200, Created = 201, BadRequest = 400, Unauthorized = 401 }
   * faker.helpers.enumValue(HttpStatus) // 200 (Ok)
   *
````

Summary:

````ts
@tag Text1
    <-- Newline
Text 2
    <-- Newline
Lost Text...
@nextTag (no newline in between)
````

Repo: https://github.com/faker-js/faker
Command: `pnpm eslint --cache --cache-strategy content . --fix`
Note: Inconsistent formatting issues exist before formatting it
Config:
````ts
        'jsdoc/sort-tags': [
          'error',
          {
            tagSequence: [
              { tags: ['internal'] },
              { tags: ['template', 'param'] },
              { tags: ['returns'] },
              { tags: ['throws'] },
              { tags: ['see'] },
              { tags: ['example'] },
              { tags: ['since'] },
              { tags: ['deprecated'] },
            ],
          },
        ],
````

**Comment by brettz9** [comment]
@ST-DDT : Really appreciate your guys' testing this. It's very helpful to have this to go by since I'm having to roll our own utilities for manipulating comment blocks. I plan to take a closer look soon.

**Comment by Shinigami92** [comment]
I'm a bit confused :confused: 
`sort-tags` does not imply to me that it manages newlines between `tagSequence`...
Couldn't we have just two different rules?
One for saying: do newlines between each tag-group, and one for sort-tags?

Then we can use the default for sorting defined by this plugin or define our own, like ```['param', 'returns', 'throws', ...]```
And tag-newline maybe e.g. with `{ doNewlineBefore: 'always', doNewlineAfter: 'always' }` or something like this (and it's the default to make newlines between each tag group)
So all `@param` are grouped together, as well as `@template`, but they are separated by one newline

**Comment by brettz9** [comment]
@ST-DDT : The bug causing both the problems should now be fixed, with the dist branch updated too.

**Comment by brettz9** [comment]
> `sort-tags` does not imply to me that it manages newlines between `tagSequence`... 
> Couldn't we have just two different rules? One for saying: do newlines between each tag-group, and one for sort-tags?

The spacing is directly related to there being a grouping, so I think it fits in `sort-tags`. However, I think we could add a boolean option to `checkSpacing` (or `checkGroupSpacing` and `checkItemSpacing`) so as to allow disabling the spacing (but it may end up seeming messy with the newlines all moved). `tag-lines` doesn't assume any groupings whatsoever, so I don't think it fits there.
 
> Then we can use the default for sorting defined by this plugin or define our own, like `['param', 'returns', 'throws', ...]`

If you don't want distinct groupings with line breaks, you can just put all the tags like:

```js
tagSequence: {
  tags: ['param', 'returns', 'throws', ...]
}
```

...which doesn't add much of a burden, in my estimation.

Or you can set `linesBetween` to 0 which drops lines between groups if you don't want those.

> And tag-newline maybe e.g. with `{ doNewlineBefore: 'always', doNewlineAfter: 'always' }` or something like this (and it's the default to make newlines between each tag group)

The `tag-newline` rule is not aware of groups. It acts and has acted strictly on a per-tag basis. (My recent PR #1018 does add more control after the last tag and before the first, but that isn't really grouping, even though it is compatible with this rule when not using the "always" or "never" options (i.e., using "any").)

> So all `@param` are grouped together, as well as `@template`, but they are separated by one newline

This can already be done per the current PR (the default). Sorry if I'm not following.

thernstig:
A future possible scenario for keeping spacing between tag groups in `sort-tags` is that it does give the option to add something like `linesBefore` and `linesAfter`, if ever wanted per tag group. Having that config separately would be unwieldy imo.

Shinigami92:
@brettz9 Oh 😯 I'm sorry I confused you
I was not aware that there is a PR that requested tag-lines as a real rule :eyes: 
All I wrote in my comment was pseudo named and not expected to be real or requested already

I just wanted to make clear that `sort` something does not imply to separate this "something"

So I think @thernstig understood me correctly and hopefully made it a bit more clear with they words

ST-DDT:
Tested the new version again. Works as expected.

brettz9:
> @brettz9 Oh 😯 I'm sorry I confused you I was not aware that there is a PR that requested tag-lines as a real rule 👀 All I wrote in my comment was pseudo named and not expected to be real or requested already

Ok, sure. And it is not merely a PR, but a preexisting rule. The PR just adds some behavior to it.

> I just wanted to make clear that `sort` something does not imply to separate this "something"

I understood that, sure.

> So I think @thernstig understood me correctly and hopefully made it a bit more clear with they words

I think @thernstig is offering additional support for my thought that the spacing should really stay within `sort-tags`. Just because "sort tags" does not inherently refer to whitespace between groupings does not mean that it doesn't make the most sense to keep the spacing options within the "sort tags" rule. After performing the sorting, it is a reasonable question as to whether spaces should be added in the process. If another rule does such spacing, how would it know where the groupings were to add spacing if the groupings weren't already sorted and in place? If the user had:

```js
@param
@returns
@param
```

...where would the spacing be applied?

The cases where it makes sense to have spacing in a separate rule are handled (or proposed to be handled) in `tag-lines`: namely, always having spacing after a particular tag based on its name only, and adding spacing after all tags or before all tags.

.README/rules/sort-tags.md
brettz9: Marking as resolved though feel free to comment further as needed.

brettz9:
I've gone ahead and added `reportIntraTagGroupSpacing` and `reportTagGroupSpacing` options so that one can disable the reporting/fixing of line breaks between or within tag groups, and I've tried to note in the docs where it may be problematic to also use `tag-lines`.

thernstig:
LGTM. I did not verify the last commits, since you added tests. I can verify if you want.

brettz9:
> LGTM. I did not verify the last commits, since you added tests. I can verify if you want.

No need, thanks. I did amend the docs just now though if you wanted to check those.

ST-DDT (approved):
I didn't exactly check the impl, but the new implementation works as expected.

brettz9:
Ok, sounds good to pull, but we're blocked until #1018 is done. I think it's almost ready.

thernstig:
LGTM

PR #1018, opened 2023-04-15
feat(`tag-lines`): add `endLines` in place of `dropEndLines`; add `startLines`

brettz9:
feat: for `tag-lines`, add `endLines` in place of `dropEndLines` and `applyToEndTag` in place of `noEndLines`; add `startLines` in place of `newline-after-description` rule

BREAKING CHANGE:

1. Removes `jsdoc/newline-after-description` rule in favor of `jsdoc/tag-lines` with option `startLines: 0` for "never" and `startLines: 1` for "always". Defaults now to `startLines: 0`
2. Removes `dropEndLines: true` from `jsdoc/tag-lines` in favor of option `endLines: 0`
3. Drops `jsdoc/tag-lines` rule's `noEndLines: true` in favor of `applyToEndTag: false`


.README/rules/tag-lines.md
```diff
@@ -21,9 +21,13 @@ Use with "always" to indicate the number of lines to require be present.
 Use with "always" to indicate the normal lines to be added after tags should
 not be added after the final tag.
 
-##### `dropEndLines` (defaults to `false`)
+##### `startLines` (defaults to `undefined`)
```
thernstig, line 24: Does `startLines: 1` together with `newline-after-description` in total add two empty lines after a description block with this?

.README/rules/tag-lines.md
```diff
@@ -21,9 +21,13 @@ Use with "always" to indicate the number of lines to require be present.
...
 
-If defined, will drop end lines for the final tag only.
+If defined, will enforce end lines to the given count before the first tag only.
+
+##### `endLines` (defaults to `undefined`)
```
thernstig, line 28: Is `noEndLines` needed now, or can this just be set to 0?

The description for `noEndLines` could probably be re-written, it reads a bit strange and if both are needed maybe mention there how it relates to `endLines`.

.README/rules/tag-lines.md
brettz9: Good question, lol. That rule should probably also be dropped with this PR. :-)

Setting `startLines` to 0 is like that rule's `"never"` option and setting it to 1 is like the rule's `"always"` option, so no need for the rule anymore.

.README/rules/tag-lines.md
brettz9: `noEndLines` means that the normal behavior of the rule doesn't apply to the very last tag.

This could probably be renamed to `applyToEndTag` (with the boolean value flipped).

Although the option might be dropped in favor of disallowing application of the rule to the final tag unless users use `endLines` to enforce the count there regardless of tag, I think that is less than ideal because then users cannot decide to have different final whitespace depending on the final tag's name as they can currently.

.README/rules/tag-lines.md
brettz9: Although we could move the behavior to `newline-after-description` to keep more backward compatibility, we would then need to refer people using `sort-tags` to two additional rules if they wanted initial tag and final tag whitespace control (`tag-lines` and `newline-after-description`). So I think it would be best to remove the `newline-after-description` rule since we're making breaking changes here anyways.

.README/rules/tag-lines.md
thernstig: Tricky. A name change might be good. It is a bit tricky to get it right with all three rules `sort-tags`, `tag-lines` and `newline-after-description`. So dropping one or more options and renaming a bit might clear this confusion up.

thernstig: I also wonder if `tag-linebreaks` would also be more explanatory, but that is a minor thing and probably not worth a rename.

thernstig:
Should you cross-reference from this rule towards `sort-tags` explaining that some line breaks are added there between tag groups. Something like:

> The rule `sort-tags` also has certain affects on line breaks, see the rule for details.

.README/rules/tag-lines.md
brettz9: The dropping of `newline-after-description` is now a part of the PR.

brettz9:
> Should you cross-reference from this rule towards `sort-tags` explaining that some line breaks are added there between tag groups. Something like:
> 
> > The rule `sort-tags` also has certain affects on line breaks, see the rule for details.

I've added something to this effect, mentioning that the rule with the "always" or "never" options should not be used together with the linebreak-setting features of `sort-tags`.

.README/rules/tag-lines.md
brettz9: Yeah, I went ahead and dropped `noEndLines: true` in favor of `applyToEndTag: false`. Hate to be making breaking changes, no less 3 breaking changes (dropping `newline-after-description` in favor of `startLines` and dropping `dropEndLines` in favor of `endLines: 0`), but I think it's all for the best.

`tag-line-breaks` would be more descriptive, but yeah, I'd rather not be doing any more breakage for such kinds of gains.

thernstig:
LGTM

.README/rules/tag-lines.md
```diff
@@ -2,28 +2,39 @@
...
+The "always" or "never" options of this rule should not
+be used with the linebreak-setting options of the `sort-tags` rule as both
+may try to impose a conflicting number of lines.
+
 #### Options
 
 The first option is a single string set to "always", "never", or "any"
 (defaults to "never").
 
 "any" is only useful with `tags` (allowing non-enforcement of lines except
```
thernstig, line 17: On line 9 above it is said that this rule `tag-lines` should only use `"any"`. But this line 17 here does not mention that. At the same time, it is not good to repeat what is said above. Would it be proper to move line 5-11 to under here instead, or are you afraid users will miss those important details then?

thernstig (commented):
@brettz9 it just struck me. If one use `sort-tags`, this rule here is still part of the `recommended` configs meaning the first option of this rule `tag-lines` defaults to `"never"`.

If using `sort-tags`, it would be very nice if this implicitly defaults to `"any"`, and for my preference, with `startLines: 0` and `endLines: 0`. But it is subjective.

Maybe also unwieldy changing the defaults depending on other rules configured or not, even though it would be nice in this instance.

Thoughts?

brettz9:
> @brettz9 it just struck me. If one use `sort-tags`, this rule here is still part of the `recommended` configs meaning the first option of this rule `tag-lines` defaults to `"never"`.

Yes, good point. The docs needed to draw that out more, so I've made clear in the docs here that that is the case.

> If using `sort-tags`, it would be very nice if this implicitly defaults to `"any"`, and for my preference, with `startLines: 0` and `endLines: 0`. But it is subjective.
> 
> Maybe also unwieldy changing the defaults depending on other rules configured or not, even though it would be nice in this instance.

Firstly a tangent...We're of course left with some ugly choices. Besides stuffing `tag-lines` and `sort-tags` into one rule as would be untenable conceptually, the only way we could coordinate between the rules would be to have a setting which would be looked at by both rules. This does offer the strong advantage of preventing users from a foot-gun, but it comes at the cost of requiring out-of-rule settings for functionality pretty basic to the rules, and so I'm inclined to do what we've done so far in warning users.

Hopefully with the `sort-tags` docs now prepared to mention that the `tag-lines` default behavior (for the recommended rule) is problematic for `sort-tags`, the foot-gun may be avoided for those who seek to apply `sort-tags`.

So I agree with the spirit of your suggestion that we do consider rules that may interact with one another, especially with default settings, but I think this may be the best of the possible solutions.

While it would be convenient for `sort-tags` users to switch the default of `tag-lines`, besides "never" being preexisting behavior for `tag-lines`, I think it is a safer default to help enforce such control of line breaks (yet without sorting tags into some partly arbitrary order).

> `startLines: 0` and `endLines: 0`.

This makes sense, I think, and though the preexisting behavior was to apply `newline-after-description` (a `startLines: 1` equivalent), I think it fits with the breaking change if we switch the behavior to 0 instead of what I have now, applying nothing. `endLines: 0` is also a reasonable default, so I think I'm inclined to agree on these two values. It also fits with the spirit of "never" and `tag-lines` I think.

.README/rules/tag-lines.md
thernstig: LGTM

thernstig:
> So I agree with the spirit of your suggestion that we do consider rules that may interact with one another, especially with default settings, but I think this may be the best of the possible solutions.

Agree. I also cannot come up with a better option and this might be left best as documentation to keep rules "orthogonal" as much as possible.

LGTM.

fb55:
I have a config with `jsdoc/newline-after-description` and am now trying to replace it with

```json
"jsdoc/tag-lines": [2, "always", { "startLines": 1 }],
```

This adds newlines between all tags (as well as trailing newlines), which is not what I'm looking for. Please let me know how the migration is supposed to look like.

brettz9:
> I have a config with `jsdoc/newline-after-description` and am now trying to replace it with
> 
> ```json
> "jsdoc/tag-lines": [2, "always", { "startLines": 1 }],
> ```
> 
> This adds newlines between all tags (as well as trailing newlines), which is not what I'm looking for. Please let me know how the migration is supposed to look like.

If you had the "always" option with the `newline-after-description` rule, you can replace it with:

```json
"jsdoc/tag-lines": [2, "any", { "startLines": 1 }],
````

Be sure to also be using the latest, 43.0.6.

thernstig:
@brettz9 how did we reason around the default `{ "startLines": 0 }`, should it maybe have been `{ "startLines": 1 }`?

brettz9:
@thernstig : While it was breaking in switching behavior, to me, it fits more with the tighter approach of `tag-lines` defaulting to `never`.

thernstig:
@brettz9 did having `newline-after-description` as a recommended rule before together with how `tag-lines` defaulted not end up with a line break after the description, or maybe I am just not remembering correctly. In that sense I thought `{ "startLines": 1 }` would mimic that.

thernstig:
@brettz9 yes I definitely said that, but when checking it more now think that having `startLines: 1` would make sense. I think I grouped `endLines: 0` and `startLines: 0` together in an unfortunate line of thought.

Not sure there is a consensus more than maybe you and me agreeing that `startLines: 1` does make sense? 😆 I think so.

brettz9:
Well, I'm not sure it makes the most sense, imho. Again, `tag-lines` enforces no lines between tags. It stands to reason for me that the default also does not enforce lines after the description either, keeping a consistently tight default up to the endLines: 0 too.

Unless perhaps it appears that the overwhelming practice (outside of the enforcing of lines this plug-in has been doing!) is to add such lines in the case of the description only. But I'm not so sure about that.

davidmurdoch:
Really would be nice if this PR explained how to migrate from the removed rule to the new one.

brettz9:
@davidmurdoch : Did you take a look at https://github.com/gajus/eslint-plugin-jsdoc/pull/1018/commits ?

PR #1034, opened 2023-04-20
Update the node engine in package.json

remcohaszing:
This is a follow up of https://github.com/es-joy/jsdoccomment/commit/3c6b512c0326a6bcdbc69d648212789fb70ce65c#r109800324.


brettz9:
I think if you could bump the minimum to Node 16, it would be best. Node 14 is going to be end-of-lifed April 30th, and in doing so, we can also bump the `.babelrc.json` `targets` to Node 16, allowing potentially better performance by avoiding kludges being added to fix a missing feature. The latter change can therefore also be added to the PR.

remcohaszing:
Do you also want me to update to `package-lock.json` version 3? That was also being blocked by Node.js 14.

.github/workflows/feature.yaml
```diff
@@ -33,7 +33,6 @@ jobs:
       fail-fast: false
       matrix:
         node_js_version:
-          - '14.17.0'
           - '16'
           - '18'
```
brettz9, line 37: Could you also add '20' here?

remcohaszing: Sure! I also updated the version for other CI jobs.

remcohaszing:
I understand why ESLint is failing, but not how to resolve it. What is `.ncurc.js`?

brettz9:
> Do you also want me to update to `package-lock.json` version 3? That was also being blocked by Node.js 14. (Version 2, the current, is literally just both version 1 and 3 combined.)

Yes, that'd be great.

brettz9:
> I understand why ESLint is failing, but not how to resolve it. What is `.ncurc.js`?

It is the config file for `npm-check-updates`. It had been preventing updating two packages until Node 16 was the minimum. You don't have to update those packages with this PR (the next time I run `npm-check-updates` or `ncu` on the command-line, it will auto-update those packages for me without need for dependabot, etc.)

remcohaszing:
Already figured it out and removed those. :smile:

brettz9:
I think I'll sit on this until the 30th in case there are updates between now and then. I know other projects transition earlier, but I'd rather avoid any further issues for users esp. with the other breaking changes.

brettz9:
Btw, how should we get past the fact that this repo is still expecting a Node 14 test?

brettz9:
ah, maybe this is something @gajus has to do...

remcohaszing:
Maybe it’s because the original commit still had that job? Maybe it helps if I make a new PR.

brettz9:
You could try rebasing into a single commit and force-pushing, or sure, if you find that too cumbersome or think it'll be necessary, a new PR...

remcohaszing:
That doesn’t seem to work.

PR #1041, opened 2023-04-21
fix: tag removal for 0-line tags (if other tags present, only blank out the tag; otherwise, remove to end of block)

### LINKED ISSUE #1040: Autofix breaks comments removing zero/final line
## Expected behavior

When removing first or last line of the comment by autofixing, the `/*` / `*/` should remain

## Actual behavior

Opening and closing of the comment can be removed by autofixer

## ESLint Config

```js
{
    plugins: ['jsdoc'],
    extends: ['plugin:jsdoc/recommended-typescript'],
    rules: {
      'jsdoc/multiline-blocks': ['warn', {
        noZeroLineText: false,
        noFinalLineText: false,
      }],
}
```

## ESLint sample

The `@typedef` and `@module` lines get removed with comment open/close, completely breaking the code
```js
/** @typedef {Object} MyObject
 * @property {string} id - my id
 */

/** some text
  @module */
function f() { }
```

## Environment

- Node version: v18.12.1
- ESLint version 8.36.0
- `eslint-plugin-jsdoc` version: 43.0.7

brettz9:
fixes #1040

@JoshuaKGoldberg : Could you take a look at the test cases and see if they are as you would expect? I find it a little strange that we are wiping out whole tags including their descriptions unless only one is left, in which case, we leave one description. Should we not wipe it out entirely if that is the goal of the option?


JoshuaKGoldberg:
Hi sorry for the delay - test cases look great to me! Thank you for fixing this so quickly! 🙏

PR #1044, opened 2023-04-22
feat(`no-blank-blocks`): add new rule

### LINKED ISSUE #1042: `no-empty-blocks` rule
## Motivation

1. I'm moving some JS code with types described as JSDoc into TS
All those param types are being removed and become empty
Making all the functions to have descriptions is not planned.
Thus I need a way to cleanup empty code blocks.
2. I'm working on https://github.com/Shinigami92/eslint-define-config
Code generation sometimes causes empty JSDoc blocks to appear here and there.
Some non-needed blocks like `@minItems` can also appear.

I am looking for an easy and straigntforward way to remove the blocks I don't need
First step would be removing unwanted tags by autofixer
Second step would be removing blocks which became empty by autofixer

## Current behavior

The documentation is too long to read it completely
Nothing searchable there clearly states that it can be used to remove lines or blocks, other then builtin autofixers

## Desired behavior

- A rule that says that it bans empty blocks in its name
- A rule that says that it bans certain tags in its name

## Alternatives considered

I have spend an hour reading documentation, understood nothing on the case

brettz9:
fixes #1042

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -608,6 +608,7 @@ selector).
 {"gitdown": "include", "file": "./rules/multiline-blocks.md"}
 {"gitdown": "include", "file": "./rules/no-bad-blocks.md"}
 {"gitdown": "include", "file": "./rules/no-blank-block-descriptions.md"}
+{"gitdown": "include", "file": "./rules/no-blank-blocks.md"}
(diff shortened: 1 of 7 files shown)


Dimava:
typo

brettz9:
Thanks, fixed in commit also...

PR #1047, opened 2023-04-23
fix(informative-docs): trim description text

### LINKED ISSUE #1039: jsdoc-informative-docs generates a report whenever you use @class
## Expected behavior
```js
/** Some Description
 *
 * @class 
 *
 * @param {number} value - Some useful text
 */
function MyAmazingThing(value)
{
    this.thing = value;
}
```

should not produce a "This description only repeats the name it provides" warning on the `@class` line. The jsdoc documentation on `@class` doesn't require anything else on the line, and to be honest, the only description that is applicable to some classes is the short one.

## Actual behavior

The `@class` line generates the above warning.

## ESLint Config

```
---
plugins:
- jsdoc

rules:

  "jsdoc/informative-docs": warn
```

## ESLint sample

See above

## Environment

- Node version: v18.16.0

- ESLint version v8.38.0

- `eslint-plugin-jsdoc` version: 43.0.7

JoshuaKGoldberg:
Right now, `informative-docs` isn't trimming the text it passes to `are-docs-informative`. The `Boolean(text)` check that's supposed to filter out blank text isn't accounting for whitespace. That means when there's a tag like `@class`, it calls the equivalent of:

```js
areDocsInformative('\n', 'MyAmazingThing')
```

...which is false.

This `.trim()`s the text so that whitespace is ignored.

Fixes #1039.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7649,6 +7649,13 @@ function takesOne(param) {}
  * takes abc param
  */
 function takesOne(param) {}
+
+/**
+ * @class 
+ *
+ * @param {number} value - Some useful text
+ */      
+function MyAmazingThing(value) {}
 ````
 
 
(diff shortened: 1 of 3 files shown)


brettz9:
Thanks for this... If you could also take a look at #1041 , @JoshuaKGoldberg ?

Issue #858, opened 2022-03-27
Rule to mark types in JSDoc as used


Josh-Cena:
## Motivation

Consider the following code:

```ts
import type { ComponentType } from "react";

/** @see {@link ComponentType} */
export type Foo = number;
```

TypeScript compiler understands that `ComponentType` is used in JSDoc and hence will not flag it as unused; TS-ESLint, on the other hand, reports [`@typescript-eslint/no-unused-vars`](https://typescript-eslint.io/rules/no-unused-vars), because it's unaware of JSDoc.

ESLint offers the `markVariableAsUsed` API that we can leverage to mark this type as used. `eslint-plugin-react` takes advantage of this to [mark `React` as used when there's JSX](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-uses-react.md). I don't know if it's portable to types, but hopefully it is.

## Current behavior

Reports error.

## Desired behavior

Should not give `@typescript-eslint/no-unused-vars`

## Alternatives considered

This could also be an extension in TS-ESLint, but AFAIK they never handle JSDoc, so I decided to see if you would consider it first.

brettz9:
I personally prefer to use `import()` within the types, feeling the importing into code as kind of an anti-pattern, and as per #99 we don't want to continue the buggy support we have for this in `no-undefined-types` (the bugginess relates to the fact that our approach was to handle comment blocks one-by-one, but for proper unused variable support, it needs to process them all together).

However, as a separate rule, not on by default, I'd be open to a PR, but am not really inclined to work on this, as again, I kind of view it as an anti-pattern, albeit an approach supported by TypeScript and I think meriting accommodation to projects which want to use it.

Josh-Cena:
I see. I do prefer to split the marking types as used feature into a separate rule, because then it would be purely syntactical (detecting `@see` or `@param` or similar tags), so we can probably handle the comment blocks one-by-one. I get your point about anti patterns; frankly, I like to use `import type` in TS files because the JSDoc `import()` alternative looks quite... ugly. Furthermore, I'm actually unable to get this working:

```ts
/** @typedef {import('react').ComponentType} ComponentType */

/** @see {@link ComponentType} */
export type Foo = number;
```

`ComponentType` is resolved as `any`, even when `import('react').ComponentType` seems to be correctly resolved.

brettz9:
Does this work for you?
```js
[...]

Issue #721, opened 2021-04-30
no-undefined-types should check `@type`


simonseyock:
## Motivation

Eslint complains about unused-vars if they only occur in `@type` annotations.

For example:
```javascript
import { SomeType } from 'some/where';

const val = /** @type{SomeType} */ (otherVal);
```

## Current behavior

`no-undefined-types` does not check `@type` annotations

## Desired behavior

`no-undefined-types` should check `@type` annotations

brettz9:
Is this promoting an anti-pattern though? With "typescript" `mode`, one can use `import()` with the type (keeping it within the jsdoc), and for discovery of defined types, I think #99 would be a better approach--i.e., ourselves navigate through the project's imports (or use some config) and/or its type `import()`'s and auto-define typedef's and globals found therein, while not encouraging importing code that has no purpose other than static analysis.

simonseyock:
It could be that this is considered an anti-pattern. I would argue that if I would need a variable of a certain type, that this code needs to be imported at some point in the software anyways. If I would write the same code in typescript I would need an import, too.

Also this seems to be inconsistent. If I use a type in a `@param` annotation it is also marked as used. So in my opinion this should be working the same for `@param` and `@type`.

It might be an option to make the whole 'marking as used' functionality optional so the user could decide which way of importing they would prefer.

golopot:
What are the benefits for using esm import over JSDoc `import()`? Using esm import for an unused variable leads to an unnecessary module loading at runtime (for node).

brettz9:
As per https://github.com/gajus/eslint-plugin-jsdoc/issues/721#issuecomment-837929396 , I think the argument here is for consistency rather than endorsing the practice.

simonseyock:
Yes, my main point is: Either mark all types / variables in type annotation as used or none.

I just wanted to elaborate that there might be cases in which it is possible to argue in favor of es6 imports. I think it might be nice to allow users to make this decision by themselves and make the whole 'marking as used' optional.

simonseyock:
Also the same should be true if you declare a variable / type in a file and only use it in an annotation.

brettz9:
As mentioned just now [here](https://github.com/gajus/eslint-plugin-jsdoc/pull/1058#issuecomment-1532995383), I'm not sure how this got by me, but:

1. The original issue as reported is not an issue, at least not anymore, but looking at the blame file, it seems the support for `@type` was preexisting the reporting of this issue.
[...]

Issue #507, opened 2020-03-23
no-unused-vars wrong error when using imported JSDoc definition as @property type.


ernestostifano:
`eslint-plugin-jsdoc/no-undefined-types` is correctly added to ESLint configuration.

```js
import {myTypesA} from '../internal/file.js'; // ERROR
import {myTypesB} from '../internal/file.js'; // NO ERROR

/**
* @typedef newType
* @property {myTypesA.someType} someProp - Some prop.
*/

/**
* @param {newType} arg - Arg.
*/
function myFunctionA(arg) {
    return arg;
}

/**
* @param {myTypesB.someType} arg - Arg.
*/
function myFunctionB(arg) {
    return arg;
}

```

**What did you expect to happen?**

There should not be an error.
JSDoc definition is correctly imported and used even when the error is present.
The error only appears when imported type is used as a type for a `@property` tag.
Error does not appear is used in `@param` tag.

**What actually happened? Please include the actual, raw output from ESLint.**
```js
/path/to/issue/file.js
  1:9  error  'fiberRegExpTypes' is defined but never used  no-unused-vars

```

Please see this [ESLint issue](https://github.com/eslint/eslint/issues/13075) for more details.

brettz9:
This is basically a dupe of #99 which I've just renamed and edited slightly to reflect this fact more clearly. See also #178. Closing, but feel free to comment further as needed or add to those issues as needed.

As I have a lot on my plate now, if someone else were up to implementing, my incomplete project at https://github.com/brettz9/es-file-traverse has some code comments which might give ideas for a strategy and tools that could be used.

brettz9:
Apologies, I think I misread your issue.

brettz9:
I've reduced the test case to this:

```js
{
      code: `
      import {myTypesA} from '../internal/file.js'; // ERROR
      import {myTypesB} from '../internal/file.js'; // NO ERROR

      /**
      * @typedef newType
      * @property {myTypesA.someType} someProp - Some prop.
      */

      /**
      * @param {newType} arg - Arg.
      */
      function myFunctionA(arg) {
          return arg;
      }

      /**
      * @param {myTypesB.someType} arg - Arg.
      */
      function myFunctionB(arg) {
          return arg;
      }
      export {myFunctionA, myFunctionB};
      `,
      parserOptions: {
        sourceType: 'module',
      },
      rules: {
        'no-unused-vars': ['error'],
      },
    },
```

[...]

Issue #99, opened 2018-10-31
Way to feed `no-undefined-types` additional files or have it auto-detect them (and finding unused types within the same set of files); avoid defects of tampering with `no-unused-vars`


brettz9:
For `no-undefined-types`, I'd like a way to feed one's project files, whether:

1. Detected by static or dynamic `import`/`require`, with the entry file(s) determined by:
    1. a new `eslint-plugin-jsdoc` option
    2. use of `main` or `exports` in `package.json`
    3. use of jsdoc config
2. By a file being present in one's jsdoc config
3. If eslint-plugin-jsdoc were supplied an array of files for finding typedefs

I need more context because some of my types are in:

- A purely-comment-based typedefs.js file which includes external and generic types
- Other files and modules

I'd like it to report for:

```js
/**
* @param {MyType}
*/
```

...where `MyType` is not defined anywhere within JSDoc typedefs/interfaces across all files. The corollary is thus not to report errors for:

```js
// One file, such as `typedefs.js`
/**
* @typedef {object} MyType
*/
/**
* @interface AnotherType
*/

// Another file:
/**
* @param {MyType}
*/

/**
* @param {AnotherType}
*/
```

Thanks very much for the very useful plugin!

**Update**: While traversing files of a project, also consider supporting `@inheritdoc` per #765

gajus:
https://github.com/gajus/eslint-plugin-flowtype#define-flow-type

brettz9:
Given that JSDoc has its own means of expressing user custom types such as through @typedef and @interface, wouldn't JSDoc types make sense as the default way to feed in definitions (and without extra effort to redefine such types)?

brettz9:
Btw, re: those PRs, I am quite impressed how intuitive your code is, that it looks very easy to make one... Will see if I can set aside some time for it...

bary12:
Generally, The ESLint-friendly way of doing this is exporting the types using `/* exported MyType */` and importing them as `/* global MyType */`, this is also how ESLint suggests to deal with variables defined in different files in browser context (without modules) when using [no-unused-vars](https://eslint.org/docs/rules/no-unused-vars#exported) and [no-undef](https://eslint.org/docs/rules/no-undef#rule-details). I don't think there's a general need in the community for this feature, and exported/global can be satisfactory.

brettz9:
[...]

PR #1058, opened 2023-05-02
Adding a `used-types` rule

kraenhansen:
This adds a `mark-used` rule which will mark types referenced by jsdoc as used.

## TODO

- [ ] Squash commits to a single commit before merging, as per contribution guidelines.
- [ ] Implement support for inline tags, most notably `{@link ...}` via a PR to `@es-joy/jsdoccomment` adding support for parsing of inline tags, as per [@brettz9's suggestiion](https://github.com/gajus/eslint-plugin-jsdoc/issues/858#issuecomment-1531613832).


src/rules/markUsed.js
```diff
@@ -0,0 +1,59 @@
...
+ * Extracts the type names from a type declaration string.
+ *
+ * @see https://jsdoc.app/tags-type.html
+ */
+const extractTypeNames = (type) => {
+  if (type.startsWith('(') && type.endsWith(')')) {
+    // Type union
+    return type.slice(1, type.length - 1).split('|').flatMap(extractTypeNames);
+  } else if (type.endsWith('[]')) {
+    // Arrays
```
brettz9, line 13: Rather than parsing the types manually, I would instead ask you to use `jsdoc-type-pratt-parser` to get more refined parsing of types. Any issues with its parsing can be filed with that project.

kraenhansen: I've now pushed a fix for this.

docs/rules/mark-used.md
```diff
@@ -0,0 +1,71 @@
...
+class Baz {}
+class Qux {}
+/** @type {(!Foo|?Bar|...Baz|Qux[]|foo=|obj["level1"]|{Foo?: Foo}|function(this:Foo))} */
+let foo = null;
+
+class Foo {}
+/** @type {typeof foo|import("some-package")|new(number, string): Foo|foo is Foo|{foo: Foo}} */
+let foo = null;
+// Settings: {"jsdoc":{"mode":"typescript"}}
+````
```
kraenhansen, line 70: @brettz9 Admittedly, these examples become a bit more involved for me to achieve 100% coverage. Does it make sense to split these up or could I somehow mark them to avoid pulling them into the .md?

src/rules/markUsed.js
```diff
@@ -0,0 +1,98 @@
...
+    const result = [];
+    if (parsed.element) {
+      result.push(...extractTypeNames(parsed.element));
+    }
+
+    if (parsed.elements) {
+      result.push(...parsed.elements.flatMap(extractTypeNames));
+    }
+
+    return result;
```
kraenhansen, line 60: I'm using a "fall through" case here, for all the types that conformed to having a root node in `element` or a list of root nodes in `elements`.

I basically read through https://github.com/jsdoc-type-pratt-parser/jsdoc-type-pratt-parser/blob/main/src/result/RootResult.ts and https://github.com/jsdoc-type-pratt-parser/jsdoc-type-pratt-parser/blob/main/src/result/NonRootResult.ts to handle all expected cases.

docs/rules/mark-used.md
brettz9: You can add the property `ignoreReadme: true` to the test object.

src/rules/markUsed.js
```diff
@@ -0,0 +1,98 @@
...
+import iterateJsdoc from '../iterateJsdoc';
+
+/**
+ * Extracts the type names from parsed type declaration.
+ */
+const extractTypeNames = (parsed) => {
+  if (typeof parsed !== 'object') {
+    /* istanbul ignore next */
+    return [];
+  } else if (parsed.type === 'JsdocTypeName') {
```
brettz9, line 13: Could you change everything starting with this line to a `switch`? I know it's less capable if needing expanding, but it helps convey more quickly that this is mostly comparing against a single variable.

brettz9: Yeah, go ahead for now.

kraenhansen: Pushed 👍

src/rules/markUsed.js
kraenhansen: This code went away.

.README/README.md
```diff
@@ -203,6 +204,7 @@ Problems reported by rules which have a wrench :wrench: below can be fixed autom
 |:heavy_check_mark:|:wrench:|[empty-tags](./docs/rules/empty-tags.md#readme)|Checks tags that are expected to be empty (e.g., `@abstract` or `@async`), reporting if they have content|
 |:heavy_check_mark:||[implements-on-classes](./docs/rules/implements-on-classes.md#readme)|Prohibits use of `@implements` on non-constructor functions (to enforce the tag only being used on classes/constructors)|
 |||[informative-docs](./docs/rules/informative-docs.md#readme)|Reports on JSDoc texts that serve only to restart their attached name.|
+|||[mark-used](./docs/rules/mark-used.md#readme)|Marks all types referenced in `{@link}`d tags as used.|
```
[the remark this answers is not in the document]
kraenhansen: Totally - good catch. I had the `link` as my primary objective when writing that. I'll rephrase:
```suggestion
|||[mark-used](./docs/rules/mark-used.md#readme)|Marks all types referenced from jsdoc tags as used.|
```

.README/rules/mark-used.md
```diff
@@ -0,0 +1,27 @@
+# `mark-used`
+
+{"gitdown": "contents", "rootId": "mark-used"}
+
+Marks all types referenced in `{@link}` tags as used.
```
kraenhansen, line 5:
```suggestion
Marks all types referenced from jsdoc tags as used.
```

.README/README.md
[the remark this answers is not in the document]
brettz9: Sure. If you could also do so where the other docs mention it too.

.README/rules/mark-used.md
```diff
@@ -0,0 +1,27 @@
+# `mark-used`
```
kraenhansen, line 1: I would some feedback on the name. It doesn't convey that this has to do with "types" being marked as used. An alternative would be `used-types` as a tribute to the `valid-types` rule, which already exists in the package.

.README/rules/mark-used.md
brettz9: One nit: change "jsdoc" to "JSDoc"

.README/rules/mark-used.md
brettz9: `used-types` sounds good to me

kraenhansen: Pushed a rename to "used-types".

kraenhansen:
@brettz9 oh my. I just realised that the `no-undefined-types` are already marking the types from regular (non inline) tags as used: https://github.com/gajus/eslint-plugin-jsdoc/blob/ec278a292ed343ac875359d66acc4dd81301ec23/src/rules/noUndefinedTypes.js#L214

I'm contemplating closing this PR and extending that rule to cover inline tags instead. What do you think?

brettz9:
> @brettz9 oh my. I just realised that the `no-undefined-types` are already marking the types from regular (non inline) tags as used:
> 
> https://github.com/gajus/eslint-plugin-jsdoc/blob/ec278a292ed343ac875359d66acc4dd81301ec23/src/rules/noUndefinedTypes.js#L214
> 
> I'm contemplating closing this PR and extending that rule to cover inline tags instead. What do you think?

It became a part of issue #99 (see [comment](https://github.com/gajus/eslint-plugin-jsdoc/issues/99#issuecomment-604176304) to try to avoid the `markVariableAsUsed` syntax there as it was buggy. Although a solution which fixed the bug would be ideal, I was thinking to separate the logic into its own rule. 

Re: the source of the bugginess, see the discussion up to this [comment](https://github.com/gajus/eslint-plugin-jsdoc/issues/507#issuecomment-602996831) including its reference to [this ESLint issue](https://github.com/eslint/eslint/issues/9810) which, to my knowledge, has not yet been fixed.

Btw, in #721 , there was some similar confusion, perhaps by myself as well, about `@type` not being supported, though I'm not sure why—perhaps due to the bugginess (that particular issue doesn't appear to report any unused errors currently except due to the last variable `val` not being used). Anyways, I thought the issue could be used to track moving the functionality to its own rule. WDYT?

brettz9:
> I just read the issues you referenced and I'm starting to get the big picture. To be perfectly honest, I'm mostly interested in fixing this for inline link tags (`{@link ...}`) and leave the `typedef` tags not yet implemented, since I'll personally be using this from TypeScript. Would you be okay with merging a PR for a new `used-types` rule which mark types referenced by everything except the `typedef` tags? I'd expect that to be a possible future enhancement, but is that a hard requirement for adding the rule?

I think we need to either add support for `@link` to `no-undefined-types` or, if it makes sense (i.e., it doesn't introduce too big of performance problems by having to iterate everything all over again), migrate entirely to a separate rule, including `@typedef` support and dropping the `markVariableAsUsed` within the `no-undefined-types` rule. A separate rule (or config) should have the benefit of avoiding these unused variable errors without requiring users to get reporting of undefined types. Maybe just an option to `no-undefined-types` to disable the reporting (while allowing the marking of unused variables) would be ideal for performance and features.

We wouldn't want to drop support for existing users, nor duplicate functionality across rules. I'm sorry for the ambivalence about a separate rule (and any extra work it's led to for you), but I'm just not fully sure about it either way at this point. I'm open to suggestions.

kraenhansen:
> I was thinking to separate the logic into its own rule.

Okay - I'll continue in this PR then 👍

> migrate entirely to a separate rule, including @typedef support

But, could that be a follow-up PR? It seems like a completely different approach is needed for that to work, right?
And since it doesn't scratch my itch, I don't think I'd be able to justify spending the time fixing it right now.

brettz9:
At this point, and as a result of our discussions, I'm leaning to changing the `no-undefined-types` rule. The typedef support is already there, so this would just be adding the link support there. I can add the options myself to disable the variable marking and/or the undefined types.

kraenhansen:
> I'm leaning to changing the no-undefined-types rule.

Okay - I can open another PR trying to adopt `no-undefined-types` to use the inline tags from https://github.com/es-joy/jsdoccomment/pull/12?

Somewhat related I'm seeing test failures in the `@es-joy/jsdoc-eslint-parser` tests, when I `pnpm link` it with `@es-joy/jsdoccomment` from my branch. I think it has to do with the latter having changes that the parser has not been adjusted to handle - is that correct?

brettz9:
> > I'm leaning to changing the no-undefined-types rule.
> 
> Okay - I can open another PR trying to adopt `no-undefined-types` to use the inline tags from [es-joy/jsdoccomment#12](https://github.com/es-joy/jsdoccomment/pull/12)?

Ok, and that is looking great, by the way. Quite a speed!

> Somewhat related I'm seeing test failures in the `@es-joy/jsdoc-eslint-parser` tests, when I `pnpm link` it with `@es-joy/jsdoccomment` from my branch. I think it has to do with the latter having changes that the parser has not been adjusted to handle - is that correct?

Yes, that may well be.

brettz9:
I'll go ahead and close as I've decided to keep things in the `no-undefined-types` rule now that we have config which can selectively disable the desired portions. Thanks for all your work on these PRs!

PR #1062, opened 2023-05-05
feat(`no-undefined-types`): checking inline tags

kraenhansen:
While it doesn't add a new / separate rule like https://github.com/gajus/eslint-plugin-jsdoc/pull/1058 aims to, this does solve the underlying need communicated in https://github.com/gajus/eslint-plugin-jsdoc/issues/858 by updating the `no-undefined-types` rule to incorporate the [recently added feature from jsdoccomment](https://github.com/es-joy/jsdoccomment/pull/12) of parsing inline tags. Since it already marks types from other tags as used, this seem like a valuable addition until this is refactored into a separate rule or setting on `no-undefined-types`.

### DIFF
--- a/src/getDefaultTagStructureForMode.js
+++ b/src/getDefaultTagStructureForMode.js
@@ -470,11 +470,6 @@ const getDefaultTagStructureForMode = (mode) => {
         [
           'typeOrNameRequired', true,
         ],
-
-        // "type"
-        [
-          'typeAllowed', true,
-        ],
       ]),
     ],
 
@@ -490,11 +485,6 @@ const getDefaultTagStructureForMode = (mode) => {
         [
           'typeOrNameRequired', true,
         ],
-
-        // "type"
-        [
-          'typeAllowed', true,
-        ],
       ]),
     ],
 
@@ -509,11 +499,6 @@ const getDefaultTagStructureForMode = (mode) => {
         // "namepath"
         [
           'typeOrNameRequired', true,
-        ],
-
-        // "type"
-        [
-          'typeAllowed', true,
         ],
       ]),
     ],
--- a/src/getDefaultTagStructureForMode.js
+++ b/src/getDefaultTagStructureForMode.js
@@ -11,7 +11,7 @@ const getDefaultTagStructureForMode = (mode) => {
   const isJsdocTypescriptOrPermissive = isJsdocOrTypescript || isPermissive;
 
   // Properties:
-  // `nameContents` - 'namepath-referencing'|'namepath-defining'|'namepath-or-url-referencing'|'text'|false
+  // `namepathRole` - 'namepath-referencing'|'namepath-defining'|'namepath-or-url-referencing'|'text'|false
   // `typeAllowed` - boolean
   // `nameRequired` - boolean
   // `typeRequired` - boolean
@@ -23,7 +23,7 @@ const getDefaultTagStructureForMode = (mode) => {
   //  `property`/`prop` (no signature)
   //  `modifies` (undocumented)
 
-  // None of the `nameContents: 'namepath-defining'` show as having curly
+  // None of the `namepathRole: 'namepath-defining'` show as having curly
   //  brackets for their name/namepath
 
   // Among `namepath-defining` and `namepath-referencing`, these do not seem
@@ -43,7 +43,7 @@ const getDefaultTagStructureForMode = (mode) => {
       'alias', new Map([
         // Signature seems to require a "namepath" (and no counter-examples)
         [
-          'nameContents', 'namepath-referencing',
+          'namepathRole', 'namepath-referencing',
         ],
 
         // "namepath"
@@ -56,7 +56,7 @@ const getDefaultTagStructureForMode = (mode) => {
     [
       'arg', new Map([
         [
-          'nameContents', 'namepath-defining',
+          'namepathRole', 'namepath-defining',
         ],
 
         // See `param`
@@ -75,7 +75,7 @@ const getDefaultTagStructureForMode = (mode) => {
     [
       'argument', new Map([
         [
-          'nameContents', 'namepath-defining',
+          'namepathRole', 'namepath-defining',
         ],
 
         // See `param`
@@ -95,7 +95,7 @@ const getDefaultTagStructureForMode = (mode) => {
       'augments', new Map([
         // Signature seems to require a "namepath" (and no counter-examples)
         [
-          'nameContents', 'namepath-referencing',
+          'namepathRole', 'namepath-referencing',
         ],
 
         // Does not show curly brackets in either the signature or examples
@@ -115,7 +115,7 @@ const getDefaultTagStructureForMode = (mode) => {
         // `borrows` has a different format, however, so needs special parsing;
         //   seems to require both, and as "namepath"'s
         [
-          'nameContents', 'namepath-referencing',
+          'namepathRole', 'namepath-referencing',
         ],
 
         // "namepath"
@@ -131,7 +131,7 @@ const getDefaultTagStructureForMode = (mode) => {
         //   counter-examples); TypeScript does not enforce but seems
         //   problematic as not attached so presumably not useable without it
         [
-          'nameContents', 'namepath-defining',
+          'namepathRole', 'namepath-defining',
         ],
 
         // "namepath"
@@ -145,7 +145,7 @@ const getDefaultTagStructureForMode = (mode) => {
       'class', new Map([
         // Allows for "name"'s in signature, but indicated as optional
         [
-          'nameContents', 'namepath-defining',
+          'namepathRole', 'namepath-defining',
         ],
 
         // Not in use, but should be this value if using to power `empty-tags`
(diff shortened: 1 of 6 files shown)


src/rules/noUndefinedTypes.js
```diff
@@ -191,10 +191,12 @@ export default iterateJsdoc(({
   });
 
   for (const tag of jsdocTagsWithPossibleType) {
+    const possibleType = tag.type || tag.namepathOrURL;
```
kraenhansen, line 194: I don't know if this is considered misuse of the `typeAllowed` "tag struct". Alternatively I could change the `tagMightHaveTypePosition` to a `tagMightHaveEitherTypeOrNamePosition` in the filter on line 190 and mark the `link` as such.

kraenhansen: Never mind. I tried and it failed two tests that I updated to include this case too.

test/rules/assertions/noUndefinedTypes.js
```diff
@@ -388,6 +388,10 @@ export default {
        */
       `,
       errors: [
+        {
+          line: 3,
+          message: 'The type \'SomeType\' is undefined.',
+        },
```
kraenhansen, line 394: This error got introduced when I started to test for name as well.

src/getDefaultTagStructureForMode.js
```diff
@@ -461,6 +459,50 @@ const getDefaultTagStructureForMode = (mode) => {
...
+      'link', new Map([
+        // Signature seems to require a namepath OR URL and might be checked as such.
+        [
+          'nameContents', 'namepath-referencing',
+        ],
+
+        // "namepath"
+        [
+          'typeOrNameRequired', true,
+        ],
```
brettz9, line 472: I know my comment stated `typeOrNameRequired `, but if we're treating the namepath/URL as a name, wouldn't this make more sense as `nameRequired`?

src/iterateJsdoc.js
```diff
@@ -870,7 +871,7 @@ const getUtils = (
   };
 
   utils.filterTags = (filter) => {
-    return jsdocUtils.filterTags(jsdoc.tags, filter);
+    return jsdocUtils.filterTags(jsdocUtils.getAllTags(jsdoc), filter);
```
brettz9, line 874: The `filterTags` utility is used by a number of other rules. Have you confirmed that it can work in those other contexts? I think we probably need to revisit the rules on a case-by-case basis, also ensuring they are polymorphic enough in structure and fixing rules to be viable for retrieval without further code modifications.

src/rules/noUndefinedTypes.js
```diff
@@ -187,14 +187,16 @@ export default iterateJsdoc(({
   const jsdocTagsWithPossibleType = utils.filterTags(({
     tag,
   }) => {
-    return utils.tagMightHaveTypePosition(tag) && (tag !== 'suppress' || settings.mode !== 'closure');
+    return utils.tagMightHaveEitherTypeOrNamePosition(tag) && (tag !== 'suppress' || settings.mode !== 'closure');
```
brettz9, line 190: I think this makes sense as some names can be referencing, but I think we need to use a method which checks that the tags are namepath-referencing and not namepath-defining (since the latter are by definition, defined).

src/rules/noUndefinedTypes.js
```diff
@@ -187,14 +187,16 @@ export default iterateJsdoc(({
...
   });
 
   for (const tag of jsdocTagsWithPossibleType) {
+    const possibleType = tag.type || tag.name || tag.namepathOrURL;
```
brettz9, line 194: I think we want to make and query new metadata for `namepathOrURL` from within the tag structure maps so we can know to check those tags for that metadata. I hesitate to call it a name, actually, if it may be a URL (even if a URL would be harmless in this rule, it might not be in others).

test/rules/assertions/noUndefinedTypes.js
brettz9: I think in the above case it is reasonable to err here, as we've defined the tag as namepath-referencing. But we're not sure what the other name is, so let's not fail there (as I think my above suggestion would entail).

src/getDefaultTagStructureForMode.js
kraenhansen: Probably, yeah.

src/iterateJsdoc.js
kraenhansen: Good point 👍 The other rules consider "tags" as non-inline, so it might break if these are included. I could introduce a second optional `includeInlineTags` argument (optionally `false`) or introduce another function called `filterAllTags` or make two calls, the original `filterTags` and another `filterInlineTags`? What do you think?

brettz9: An additional `includeInlineTags` argument SGTM but ok if the code demands something else...

src/rules/noUndefinedTypes.js
kraenhansen: I see two alternatives for this:
1. A new value for `nameContents`, such as `"namepath-or-url-referencing"`
2. A new `namepathOrURLRequired` boolean.

I'm probably leaning towards 2, because the links don't store this in a `name`.

brettz9: With the options being of like nature, how about we just rename `nameContents`?

kraenhansen: I mean. The "name" in `nameContents` properly refers to the `name` property on the AST node. So it feels incorrect to use it to describe the contents of the `namepathOrURL` off the inline tag AST nodes. Don't it?

brettz9: Sure, I get you, but I just mean that since inline tags are never likely to have a `name` portion anyways, we may as well merge the concepts and just rename according to that single concept--something like `role: namepath-defining`, `role: namepath-or-url-referencing`, etc.

kraenhansen: Now I understand 👍 Thanks for the clarification.

kraenhansen: Another alternative would be to refactor the `nameContents` into booleans on the structure, like `namepathDefining: true`, namepathReferencing: true` and then add `namepathOrURLReferencing: true`. But that would allow for a state of both defining and referencing and might require more changes than a simple rename. The advantage would be it would simplify the struct a bit, as it would only have boolean flags and we wouldn't have to use a name for that `role`.

brettz9: While booleans can be easier to deal with, I'd like for us to prefer strings in such cases where the choices do not overlap for conceptual clarity. (If worried about typos, we could use something like enums, but I'm not too big of a fan of that personally.)

That way also, there is no chance for two booleans to be set in such a way that it is not possible. But if they do overlap--e.g., if something could be namepath-referencing and namepath-defining, then we might need booleans.

kraenhansen: > But if they do overlap

I don't think any tags will 👍

I'll give the rename to `role` a go. Do you think we should reuse the `nameRequired` boolean for the `namespaceOrURL` on inline tags or would it be better to add a `namepathOrURLRequired` to be more explicit?  (I'm leaning towards the latter).

brettz9: That should be fine.

brettz9: Or we could rename it to `nameishRequired` :-) But whatever is fine.

brettz9:
FWIW, @kraenhansen , I've submitted #1065 to allow disabling reporting (so as to just use the marking of unused variables) as well as to disable marking of variables used (so as to just get reporting of undefined types).

src/rules/noUndefinedTypes.js
kraenhansen: I renamed to `namepathRole`, I felt `role` was a bit too implicit.

test/rules/assertions/noUndefinedTypes.js
```diff
@@ -381,13 +381,17 @@ export default {
     {
       code: `
       /**
-       * @namepathDefiner SomeType
+       * @namepathReferencing SomeType
```
kraenhansen, line 384: This seem like a better name, to reflect the settings passed.

test/rules/assertions/noUndefinedTypes.js
```diff
@@ -396,17 +400,19 @@ export default {
...
         },
       },
     },
     {
+      // An unknown tag without an namepath contents declared via settings,
+      // defaults to not having an impact on the type.
       code: `
       /**
-       * @namepathDefiner SomeType
+       * @namepathMentioning SomeType
```
kraenhansen, line 415: Since this doesn't define nor reference types, I feel "mention" is more describing.

kraenhansen:
@brettz9 I've rebased on the latest `main`.

brettz9:
Ok, thanks, I have a PR to come shortly.

brettz9:
Merged now as part of misc. items in #1087

PR #1065, opened 2023-05-07
feat: `disableReporting` and `markVariablesAsUsed` options

brettz9:


src/rules/noUndefinedTypes.js
```diff
@@ -206,11 +208,11 @@ export default iterateJsdoc(({
     }) => {
       if (type === 'JsdocTypeName') {
         const structuredTypes = structuredTags[tag.tag]?.type;
-        if (!allDefinedTypes.has(value) &&
+        if (!disableReporting && !allDefinedTypes.has(value) &&
```
kraenhansen, line 211: Won't this risk marking an undefined type as used? I don't know if that's a problem or just a no-op.

brettz9: Hmmm. Good catch. I think it is a problem. However, I was considering though whether someone just wants all used types to be trusted as used. Not sure if that deserves its own config too, or whether I should solely put the `disableReporting` inside the `if`.

kraenhansen: > put the `disableReporting` inside the `if`.

I think that would be the simplest and safe bet 👍

.README/rules/no-undefined-types.md
```diff
@@ -47,11 +47,18 @@ array's items will be considered as defined for the purposes of that tag.
...
 
 - `definedTypes` - This array can be populated to indicate other types which
   are automatically considered as defined (in addition to globals, etc.).
   Defaults to an empty array.
+- `markVariablesAsUsed` - Whether to mark variables as used for the purposes
+  of the `no-unused-vars` rule when they are not found to be undefined.
+  Defaults to `true`. May be set to `false` to enforce a practice of not
+  importing types unless used in code.
+- `disableReporting` - Whether to disable reporting of errors. Defaults to
+  `false`. This may be set to `true` in order to take advantage of marking
```
kraenhansen, line 60: I suppose this is exposed for developer that just want the marking af of types as used.

```suggestion
  `false`. This may be set to `true` in order to take advantage of only marking
```

PR #1073, opened 2023-05-10
docs: indicate available `enableFixer` option; fixes #1072

### LINKED ISSUE #1072: Question (docs?): `enableFixer` for certain rules
I just noticed that e.g. https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/no-blank-blocks.md#readme has an `enableFixer` rule.  It defaults to false.

It seems some rules have this. It is required to be set to `true` in order for `--fix` to work, at least that is how I interpret it (and when I tested it). Correct me if I am wrong.

Two questions:
1. How come it is not default set to `true`? It is what I would expect, but I bet there is a rationale to set it to default `false`.
2. Is this an option that is intended to be added to all rules to control where the rules, that can automatically fix faults, should do it or not when adding the `--fix` flag? If so should this general rule be documented in the main README.md at least close to the table showing what rules are fixable (with the wrench icon)?

With the caveat that I completely misunderstood its usage 😆

brettz9:


brettz9:
> @brettz9 I wonder if there should be a new column mentioning if it is default enabled or not? I think that would avoid a few pain points.

I considered that, but I think it is preferable not to bloat the table too much. Hopefully the note and the rules themselves will make it clear.

PR #1086, opened 2023-05-15
fix: check-tag-names crash on module comment

### LINKED ISSUE #1085: `jsdoc/check-tag-names` throws `TypeError: Cannot read properties of null (reading 'type')` on module-level comment when `typed: true`
## Expected behavior

We're using typedoc and it documents the use of a module-level (i.e. not associated with an AST node) comments: See https://typedoc.org/tags/module/

In a TypeScript codebase I expect to be able add a "free" / module-level doc string:

```typescript
/**
 * @module
 * This describes the file and not the following identifier.
 */
```

## Actual behavior

The plugin throws:

> TypeError: Cannot read properties of null (reading 'type')

from `isInAmbientContext (eslint-plugin-jsdoc/dist/rules/checkTagNames.js:56:20)`

when the `jsdoc/check-tag-names` `typed` option is `true`.

## ESLint Config

```js
{
  "plugins": [
    "jsdoc"
  ],
  "rules": {
    "jsdoc/check-tag-names": ["warn", { "typed": true }]
  }
}
```

## ESLint sample

```typescript
/**
 * @module
 * This is a module level comment
 */
```

## Environment

- Node version: v18.9.0
- ESLint version: v8.40.0
- `eslint-plugin-jsdoc` version: 44.2.3

kraenhansen:
This fixes #1085 by:
1. Removing the `module` tag from the set of `typedTagsUnnecessaryOutsideDeclare`, since typedoc promotes this, even outside declarations: See https://typedoc.org/tags/module/ 
2. Adding an early return in the check-tag-names rule's `tagIsRedundantWhenTyped` function, to avoid crashing when encountering a comment, which is not associated with an AST node.

### DIFF
--- a/docs/rules/check-tag-names.md
+++ b/docs/rules/check-tag-names.md
@@ -1122,5 +1122,11 @@ interface WebTwain {
  * @satisfies
  */
 // Settings: {"jsdoc":{"mode":"typescript"}}
+
+/**
+ * @module
+ * A comment related to the module
+ */
+// "jsdoc/check-tag-names": ["error"|"warn", {"typed":true}]
 ````
 
--- a/src/rules/checkTagNames.js
+++ b/src/rules/checkTagNames.js
@@ -143,6 +143,10 @@ export default iterateJsdoc(({
       return false;
     }
 
+    if (node === null) {
+      return false;
+    }
+
     if (context.getFilename().endsWith('.d.ts') && [
       'Program', null, undefined,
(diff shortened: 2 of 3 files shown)


brettz9:
I have less of a concept as the TS I've worked on has tended to be as JSDoc + JS.

kraenhansen:
Let's keep it at this and I'll leave my other branch available for future devs, which might disagree with this default behavior.

PR #1113, opened 2023-06-05
check peerDependencies in `imports-as-dependencies`

DetachHead:


DetachHead:
maybe all three of them should be optional like in [`import/no-extraneous-dependencies`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md)?

brettz9:
That would be fine, though I think default off for peer and default on for the others...

Issue #1130, opened 2023-07-07
@types/eslint-plugin-jsdoc or bundle .d.ts


snebjorn:
## Motivation

With the new ESLint flat config we import plugins

```js
import jsdoc from "eslint-plugin-jsdoc";
```
However that causes typescript in the VSCode to report an error due to missing type definitions.

> Could not find a declaration file for module 'eslint-plugin-jsdoc'. 'myrepo/node_modules/eslint-plugin-jsdoc/dist/index.js' implicitly has an 'any' type.
>  Try `npm i --save-dev @types/eslint-plugin-jsdoc` if it exists or add a new declaration (.d.ts) file containing `declare module 'eslint-plugin-jsdoc';`

## Current behavior

There are no types for eslint-plugin-jsdoc

## Desired behavior

`eslint-plugin-jsdoc` should bundle the typescript types.

I'm not sure this is correct, but here's an example of the d.ts

```ts
import type { ESLint } from "eslint";

interface JSDoc extends ESLint.Plugin {
  configs: ESLint.Plugin["configs"] & {
    recommended: ESLint.ConfigData;
    ["recommended-error"]: ESLint.ConfigData;
    ["recommended-typescript"]: ESLint.ConfigData;
    ["recommended-typescript-error"]: ESLint.ConfigData;
    ["recommended-typescript-flavor"]: ESLint.ConfigData;
    ["recommended-typescript-flavor-error"]: ESLint.ConfigData;
  };
}

export default JSDoc;
```

## Alternatives considered

A DefinitelyTyped package `@types/eslint-plugin-jsdoc`

snebjorn:
#1139 added
https://github.com/gajus/eslint-plugin-jsdoc/blob/e8b14756c573bc7f6eb824f2e87b67bbd6a50fab/package.json#L86
but index.d.ts is missing from the distributed package

![image](https://github.com/gajus/eslint-plugin-jsdoc/assets/1266245/ebc2f409-3c82-4776-8b0d-e1897c7ebff1)

paulshryock:
> but **index.d.ts** is missing from the distributed package

That's because `*.d.ts` is explicitly ignored in [`.npmignore`](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/.npmignore#L9).

Looks like `.npmignore` should be updated to to not ignore `index.d.ts`.

brettz9:
Yes, sorry, I inadvertently actually added that when merely considering adding it.

[...]

PR #1141, opened 2023-08-06
fix: exclude unused files from publish process

falsepopsky:
## Issue

This pull request addresses the issue of including unused files in the [publish](https://www.npmjs.com/package/eslint-plugin-jsdoc?activeTab=code) process.
ex: `tsconfig.json` and `tsconfig-prod.json`

## Solution

With this pull request, when adding the `src` and `docs` folders to the `files` section in `package.json`, the build process will automatically exclude other unused files. Note that `readme` and `license` files are always included (for documentation purposes please refer to [this link](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#files)).

### Before

filesbefore

### After

1. run `npm pack`
2. To visually identify which files are packaged, you have two options:
   - Check the log for added files.
   - Extract the tarball to see the files.

after

### Additional Comments:

- I had to add the suffix `/dist` to imports requiring `@typescript-eslint/types` because it was being inferred as `any`, leading to lint errors.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -5,3 +5,4 @@ test
 .*
 *.log
 CONTRIBUTING.md
+tsconfig*.json
--- a/package.json
+++ b/package.json
@@ -67,6 +67,10 @@
   "engines": {
     "node": ">=16"
   },
+  "files": [
+    "dist",
+    "docs"
+  ],
   "keywords": [
     "eslint",
     "plugin",
--- a/package.json
+++ b/package.json
@@ -67,10 +67,6 @@
   "engines": {
     "node": ">=16"
   },
-  "files": [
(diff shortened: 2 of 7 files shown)


Fdawgs:
Could save a few more bytes by using an `.npmignore` file instead, as `package.json` (and in turn the `files` key and its contents) are published.

brettz9:
Thanks for the PR. I submitted another one which makes some fixes to tsconfig. However, that raises another question. While #1130 is seeking for the bundling of types, I'm not sure we want to add to the package size to do so, especially since an ESLint config file may have a different import approach then the rest of the package containing it.

I'm also wondering whether we should exclude `/docs`.

PR #1196, opened 2024-02-05
fix(informative-docs): Improve wording in rule description

RDIL:
"Restart" doesn't make sense here, I'm pretty sure ["restate"](https://www.merriam-webster.com/dictionary/restate) fits the intended meaning better.


brettz9:
Thanks for the PR! If adding any more minor doc fixes, please use `docs()` instead of `fix()` in the commit. Thanks!

PR #1206, opened 2024-02-21
feat(check-param-names): Add disableMissingParamChecks option

### LINKED ISSUE #1169: Allow ignoring missing @params
## Motivation

Often many parameters are self-explanatory, but some aren't. It would be great if `jsdoc/check-param-names` had an option to skip the missing 

## Current behavior

```ts
/**
 * @param additionalChecks If true, it is checked if the argument is within the bounds 0<x<100
 */
function add10(target: number, additionalChecks: boolean) {
```

Error:
Expected @param names to be "target, additionalChecks". Got "additionalChecks".

## Desired behavior

Some parameters are self-explanatory. Adding a comment about them just increases noise and reduces readability. I understand if the default should stay as is but a config option like` "jsdoc/check-param-names":["warn",{ disableMissingParamChecks: true}]` would be great.

## Alternatives considered

The only option I saw is disabling the rule completely - but that would also remove the checks for whether the param names are correct which I like.

seanpoulter:
### Proposed Changes

- feat(check-param-names): Add disableMissingParamChecks option
- test(require-param): Add missing line numbers in errors<br>
  This ~suppressed~ fixes the output like this:
  ```
  Rule, `require-param`, missing line numbers in errors: 21
  ```

### Things to Note

- This PR closes #1169.
- I believe the tests cases should give us confidence in the behaviour when the option is not provided, and when it is set. Did you want test cases for other languageOptions?
- I haven't been able to run `create-docs` locally :thinking:

### DIFF
--- a/.README/rules/check-param-names.md
+++ b/.README/rules/check-param-names.md
@@ -83,7 +83,7 @@ that are available and actually used in the function.
 
 ### `disableMissingParamChecks`
 
-Whether to check for missing `@param` definitions. Defaults to `false`. Change to `true` if you want to be able to omit properties.
+Whether to avoid checks for missing `@param` definitions. Defaults to `false`. Change to `true` if you want to be able to omit properties.
 
 ## Context and settings
 
--- a/.README/rules/check-param-names.md
+++ b/.README/rules/check-param-names.md
@@ -81,12 +81,16 @@ item at the same level is destructured as destructuring will prevent other
 access and this option is only intended to permit documenting extra properties
 that are available and actually used in the function.
 
+### `disableMissingParamChecks`
+
+Whether to check for missing `@param` definitions. Defaults to `false`. Change to `true` if you want to be able to omit properties.
+
 ## Context and settings
 
 |||
 |---|---|
 |Context|`ArrowFunctionExpression`, `FunctionDeclaration`, `FunctionExpression`|
-|Options|`allowExtraTrailingParamDocs`, `checkDestructured`, `checkRestProperty`, `checkTypesPattern`, `disableExtraPropertyReporting`, `enableFixer`, `useDefaultObjectProperties`|
+|Options|`allowExtraTrailingParamDocs`, `checkDestructured`, `checkRestProperty`, `checkTypesPattern`, `disableExtraPropertyReporting`, `disableMissingParamChecks`, `enableFixer`, `useDefaultObjectProperties`|
 |Tags|`param`|
 |Aliases|`arg`, `argument`|
 |Recommended|true|
--- a/src/rules/checkParamNames.js
+++ b/src/rules/checkParamNames.js
@@ -227,7 +227,7 @@ const validateParameterNames = (
       funcParamName = functionParameterName;
     }
 
-    if (funcParamName !== tag.name.trim() && !disableMissingParamChecks) {
+    if (funcParamName !== tag.name.trim()) {
       // Todo: Improve for array or object child items
       const actualNames = paramTagsNonNested.map(([
         , {
@@ -247,10 +247,20 @@ const validateParameterNames = (
         return item;
       }).filter((item) => {
         return item !== 'this';
-      }).join(', ');
+      });
+
+      // When disableMissingParamChecks is true tag names can be omitted.
+      // Report when the tag names do not match the expected names or they are used out of order.
+      if (disableMissingParamChecks) {
+        const usedExpectedNames = expectedNames.map(a => a?.toString()).filter(expectedName => !!expectedName && actualNames.includes(expectedName));
+        const usedInOrder = actualNames.every((actualName, idx) => actualName === usedExpectedNames[idx]);
+        if (usedInOrder) {
+          return false;
+        }
+      }
 
       report(
-        `Expected @${targetTagName} names to be "${expectedNames}". Got "${actualNames.join(', ')}".`,
+        `Expected @${targetTagName} names to be "${expectedNames.join(', ')}". Got "${actualNames.join(', ')}".`,
         null,
         tag,
       );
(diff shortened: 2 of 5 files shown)


brettz9:
Thanks for the PR.

Re: "Rule, `require-param`, missing line numbers in errors: 21", I'd prefer we keep that showing up because otherwise we don't know which `require-param` tests are broken. By the line numbers being missing, we know that the behavior needs to be fixed so that the proper line numbers will result. Suppressing them doesn't help here, it obscures the problem.

brettz9:
Re: `create-docs`, yes, there is a bug. Awaiting @gajus to review https://github.com/gajus/gitdown/pull/67 .

seanpoulter:
> Re: "Rule, `require-param`, missing line numbers in errors: 21", I'd prefer we keep that showing up because otherwise we don't know which `require-param` tests are broken.

I found and fixed the missing `line` numbers using this dynamic import with a cache busting query parameter in the Node REPL:

```
let i = 0;
(await import(`./test/rules/assertions/requireParam.js?i=${i++}`)).default.invalid.filter(a => a.errors.filter(b => b.line === undefined).length !== 0)[0].errors
```

That finds the first `invalid` case without a `line` in the `errors` and printed the text to find/replace. Thirty iterations later it's better than when I found it. :tada:

test/rules/assertions/checkParamNames.js
```diff
@@ -1835,5 +1856,20 @@ export default {
...
+        /**
+         * @param foo
+         * @param foo.bar
+         */
+        function quux (bar, foo) {
+        }
+      `,
+      options: [
+        {
+          disableMissingParamChecks: true,
```
brettz9, line 1870: Is there a test case we can add where `disableMissingParamChecks: true` but which still fails?

brettz9:
> > Re: "Rule, `require-param`, missing line numbers in errors: 21", I'd prefer we keep that showing up because otherwise we don't know which `require-param` tests are broken.
> 
> **Edit**: I should have been more careful with my words. I ~suppressed~ fixed the output. :)

What I mean is that we have a bug in `require-param`. These line numbers should only be added if that bug is fixed by fixing some code in `require-param`. The line numbers which you added to avoid the errors are not the line numbers we actually want even though they reflect the current behavior.

test/rules/assertions/checkParamNames.js
seanpoulter: Yea, we only want to skip known params so this should fail. The implementation is way too naive.

```
        /**
         * @param foo
         * @param foo.bar
         */
        function quux (bar, baz) {
        }
```

seanpoulter: I've added two new test cases. If I understand correctly, that code is not hit when the expected or actual names include property lookup for objects or indices so it seems to be a nice and simple solution.

seanpoulter:
I reverted the changes to `require-param`. Let me know if you'd rather a comment like `// Todo: Fix line number. See issue #?` if it'll help save you time later. :)

test/rules/assertions/checkParamNames.js
brettz9: How about another test where the params are `foo, bar` and the only value present is `foo`?

seanpoulter: Of course. I ended up adding 2:
* Your suggestion - When the option is set it is invalid if there is an extra `@param`
* When the option is set it is valid to have no tags

And moved the last one:
* When the option is set it is valid to skip a tag

test/rules/assertions/checkParamNames.js
```diff
@@ -1183,6 +1183,89 @@ export default {
...
+         * @param foo
+         * @param bar
+         */
+        function quux (foo) {
+        }
+      `,
+      options: [
+        {
+          disableMissingParamChecks: true,
+        },
```
brettz9, line 1260: If the option lets one skip arguments at the beginning, why not at the end?

brettz9: Oh, nevermind, I got confused.

seanpoulter: **Done**

I've found the test data without explaining the high-level behaviours confusing too. I stopped myself from adding comments like this:

```js
// When the disableMissingParamChecks is true, all `@param`s can be omitted.

// ..., the first `@param` can be omitted.

// ..., the last `@param` can be omitted.
```

.README/rules/check-param-names.md
```diff
@@ -81,12 +81,16 @@ item at the same level is destructured as destructuring will prevent other
 access and this option is only intended to permit documenting extra properties
 that are available and actually used in the function.
 
+### `disableMissingParamChecks`
+
+Whether to check for missing `@param` definitions. Defaults to `false`. Change to `true` if you want to be able to omit properties.
```
brettz9, line 86:
```suggestion
Whether to avoid checks for missing `@param` definitions. Defaults to `false`. Change to `true` if you want to be able to omit properties.
```

src/rules/checkParamNames.js
```diff
@@ -245,10 +247,20 @@ const validateParameterNames = (
         return item;
       }).filter((item) => {
         return item !== 'this';
-      }).join(', ');
+      });
+
+      // When disableMissingParamChecks is true tag names can be omitted.
+      // Report when the tag names do not match the expected names or they are used out of order.
+      if (disableMissingParamChecks) {
+        const usedExpectedNames = expectedNames.map(a => a?.toString()).filter(expectedName => !!expectedName && actualNames.includes(expectedName));
```
brettz9, line 255:
```suggestion
        const usedExpectedNames = expectedNames.map(a => a?.toString()).filter(expectedName => Boolean(expectedName) && actualNames.includes(expectedName));
```

Just prefer to keep it more semantic...

brettz9 (changes requested):
Beyond my suggestions, I think the PR looks ready...

seanpoulter:
Thanks for your patience and reviews @brettz9!

brettz9 (approved):

brettz9:
> Thanks for your patience and reviews @brettz9!

No need for patience—was a solid PR! Thanks again!

PR #1216, opened 2024-03-25
docs: mention excludedTags in informative-docs docs

### LINKED ISSUE #1214: Missing docs for informative-docs > excludedTags option
## Motivation

`jsdoc/informative-docs` will scan the body and all tags of a JSDoc tag for informative-ness. This is good and intended in most cases. But sometimes, there are tags that are used as categorization / metadata and only coincidentally match the name of the construct they're documenting. For example:

```js
/**
 * Calculates and retrieves the compiler options for the current project.
 * @category Compiler Options
 */
function getCompilerOptions() {
  // ...
}
```

Standalone reproduction of that snippet for reference: https://github.com/JoshuaKGoldberg/repros/tree/jsdoc-informative-docs-category-tag

## Current behavior

```plaintext
  3:1  error  This tag description only repeats the name it describes  jsdoc/informative-docs
```

## Desired behavior

Can we add an option to the rule to allow a list of tags to be ignored? Say, `ignoredTags`?

```js
"jsdoc/informative-docs": ["error", {
  "ignoredTags": ["category"]
}]
```

## Alternatives considered

An inline disable comment is doable, but clutters the code and is generally inconvenient.

JoshuaKGoldberg:
Fixes #1214.

### DIFF
--- a/.README/rules/informative-docs.md
+++ b/.README/rules/informative-docs.md
@@ -35,6 +35,20 @@ The default `aliases` option is:
 }
 ```
 
+### `excludedTags`
+
+Tags that should not be checked for valid contents.
+
+For example, with `{ excludedTags: ["category"] }`, the following comment would not be considered uninformative:
+
+```js
+/** @category Types */
+function computeTypes(node) {
+  // ...
+}
+
+No tags are excluded by default.
+
 ### `uselessWords`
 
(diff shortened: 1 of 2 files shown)


brettz9:
Thanks for the PR!

PR #1230, opened 2024-05-15
chore: remove is-builtin-module

43081j:
Removes `is-builtin-module` and uses the built in function provided by node itself (via `node:module`).

As per your `engines`, all node versions you support have this.


brettz9:
Thanks for the PR! Good catch!

PR #1238, opened 2024-06-13
fix: add names to flat configs

voxpelli:
See https://eslint.org/docs/latest/use/configure/configuration-files#configuration-naming-conventions and eg https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/459

Improves presentation in the config inspector: https://github.com/eslint/config-inspector

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -128,17 +128,19 @@ const index = {
 
 /**
  * @param {"warn"|"error"} warnOrError
- * @param {boolean} [flat]
+ * @param {string} [flatName]
  * @returns {import('eslint').Linter.FlatConfig}
  */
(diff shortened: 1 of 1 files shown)


brettz9:
@voxpelli : A belated thanks for this PR. Do you think we should prefix the name with `jsdoc/` though? Seems the docs suggest incorporating the plugin name too.

voxpelli:
> For example, if you are creating a configuration object for a plugin named eslint-plugin-example, you might add name to the configuration objects with the example/ prefix

@brettz9 🙏 that's what I'm reading in the docs?

brettz9:
Ok, I can add that

brettz9:
Released in v48.2.12.

Issue #1267, opened 2024-07-19
[`sort-tags`] `@template` must be before `@typedef`


regseb:
## Expected behavior

The `@template` tag must be before the `@typedef` tag.

## Actual behavior

The `@template` tag is put after the `@typedef`. TypeScript will report these errors:

```
index.js:2:13 - error TS8021: JSDoc '@typedef' tag should either have a type annotation or be followed by '@property' or '@member' tags.

2  * @typedef Foo<T>
              ~~~

index.js:8:11 - error TS2315: Type 'Foo' is not generic.

8  * @type {Foo<string>}
```

### `package.json`

```json
{
  "name": "testcase",
  "version": "1.0.0",
  "dependencies": {
    "eslint": "9.7.0",
    "eslint-plugin-jsdoc": "48.8.0",
    "typescript": "5.5.3"
  }
}
```

### `eslint.config.mjs`

```javascript
import jsdoc from "eslint-plugin-jsdoc";

export default [
    {
        plugins: { jsdoc },
        rules: {
            "jsdoc/sort-tags": "error",
        }
    }
];
```

## `index.js`

```javascript
/**
 * @template {Object} T
 * @typedef Foo<T>
 * @prop {T} bar
 */

/**
 * @type {Foo<string>}
 */
const foo = { bar: "baz" };
```

## Steps

1. `npm install`
2. `npx eslint`

   ```
   /home/regseb/testcase/index.js
     2:1  error  Tags are not in the prescribed order: summary, [...], todo  jsdoc/sort-tags

   ✖ 1 problem (1 error, 0 warnings)
     1 error and 0 warnings potentially fixable with the `--fix` option.
   ```

3. `npx tsc --checkJs --noEmit index.js`

   > No error

4. `npx eslint --fix`
5. `cat index.js`

   ```javascript
   /**
    * @typedef Foo<T>
    * @template {Object} T
    * @prop {T} bar
    */

   /**
    * @type {Foo<string>}
    */
   const foo = { bar: "baz" };
   ```

6. `npx tsc --checkJs --noEmit index.js`

   ```
   index.js:2:13 - error TS8021: JSDoc '@typedef' tag should either have a type annotation or be followed by '@property' or '@member' tags.

   2  * @typedef Foo<T>
                 ~~~

   index.js:8:11 - error TS2315: Type 'Foo' is not generic.

   8  * @type {Foo<string>}
               ~~~~~~~~~~~


[...]

PR #1268, opened 2024-07-19
fix(`sort-tags`): `@template` must be before `@typedef`

regseb:
Fix https://github.com/gajus/eslint-plugin-jsdoc/issues/1267


brettz9 (approved):
LGTM!

PR #1297, opened 2024-08-07
fix(`lines-before-block`): remove rule from recommended

brettz9:
fix(`lines-before-block`): remove rule from recommended

BREAKING CHANGE:

Have to selectively enable the `lines-before-block` rule to have it apply now.

The rule still has some issues to work out, so dropping from recommended for now.


brettz9:
Thanks! Fixing in #1306

PR #1299, opened 2024-08-09
Enable Publish Attestation (Provenance)

elliot-huffman:
Enable provenance on the NPM publish command so that attestation data is available in the NPM public registry to reduce the risk of a supply chain attack.

This is a non-breaking change and provides additional metadata for system that know how to consume this additional package metadata. Non-supported systems ignore this.

NPM Publish Command Docs (on provenance):
https://docs.npmjs.com/generating-provenance-statements

Semantic Release support statement:
https://github.com/semantic-release/npm?tab=readme-ov-file#npm-provenance


brettz9:
@gajus : Do you want to take a look?

gajus:
This looks well intended. Not familiar with the configuration, but support it.

.github/workflows/main.yaml
```diff
@@ -21,8 +33,5 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-name: Build and release
-on:
-  push:
-    branches:
-      - main
+          NPM_CONFIG_PROVENANCE: true
```
brettz9, line 36: Out of curiosity, is this actually required? At https://docs.npmjs.com/generating-provenance-statements#using-third-party-package-publishing-tools , it mentions adding this "If you publish your packages with tools that do not directly invoke the npm publish command", but semantic-release apparently does so.

elliot-huffman: That is also a valid configuration, I chose the environmental variable way as it reduces direct code change or config change. It's more of a ci/cd change. 

I'm willing to change this to a package.json change if you really want me to put it into your project directly.

Both configuration strategies result in the same output.

brettz9: No, that's fine. I guess semantic-release uses this config to give the special `npm publish --provenance` command?

elliot-huffman: Even cooler, npm itself is what takes this environmental variable to trigger the provenance release. 
Semantic release knows nothing about what's happening. 

Semantic releases just triggering npm publish, and mpm published checks for the environmental variable, the command line argument, or the package json configuration.

PR #1332, opened 2024-10-17
fix(no-multi-asterisks): add missing `docs.description`

ntnyq:
This PR add the missing `meta.docs.description` for rule `no-multi-asterisks`.


brettz9:
Excellent, thank you... I see that one other rule has this problem. Do you think you could also add the description for the `textEscaping` rule, i.e.:

> Auto-escape certain characters that are input within block and tag descriptions

brettz9:
Nevermind, I see a separate PR... :-)

PR #1335, opened 2024-11-06
docs: make rule docs reflect their recommended status correctly

CloudNStoyan:
Currently the following rules have different recommended status than their respective status in the `README.md` recommended list: 
- `check-access`
- `lines-before-block`
- `no-defaults`
- `require-asterisk-prefix`
- `require-template`
- `require-throws`

This PR fixes that


brettz9:
Thanks for the PR!

PR #1339, opened 2024-11-18
chore(package): fix repository url; add homepage and bugs urls

Fdawgs:
The repository URL was fixed by simply running `npm pkg fix`, which is [what `npm publish` does prior to publishing](https://docs.npmjs.com/cli/v10/commands/npm-pkg#description).

The addition of the [`homepage`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json?v=true#homepage) property allows for the URL to the repo to be correctly displayed when hovered over in VS Code when using a proxy registry like in Azure DevOps:

image

Compared to eslint-plugin-regexp, which [does have the homepage set](https://github.com/ota-meshi/eslint-plugin-regexp/blob/2d90a1aca863508647115c76a7b0c23b493eb0e0/package.json#L60):

image

The [`bugs`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json?v=true#bugs) property was also added to allow for `npm bugs` to work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,7 +134,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/gajus/eslint-plugin-jsdoc"
+    "url": "git+https://github.com/gajus/eslint-plugin-jsdoc.git"
   },
   "run-if-changed": {
     "package-lock.json": "npm run install-offline"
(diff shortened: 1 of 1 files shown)


brettz9:
Thanks for the PR!

PR #1382, opened 2025-05-10
 fix(`valid-types`): fix parsing of expressions like `@returns {@link SomeType}`

### LINKED ISSUE #1381: valid-types cannot parse links right after another tag
## Expected behavior
I expected `@remarks {@link SomeType}` to parse.

## Actual behavior
I get an error saying it's invalid. I considered posting this issue on https://github.com/jsdoc-type-pratt-parser based upon the guidance in this issue format but I don't believe it's part of the concern of that library. It's probably being given an erroneous input.

## ESLint Config

```json
{
    "jsdoc/valid-types": "error"
}
```

## ESLint sample

```ts
type SomeType = number;

// Ok
/** {@link SomeType} */

// Ok
/** @remarks foo bar lorem ipsum {@link SomeType} */

// Not ok
/** @remarks {@link SomeType} */
```

## Environment

- Node version: v20.15.0
- ESLint version v9.26.0
- `eslint-plugin-jsdoc` version: ^50.6.11

LukeAbby:
Fixed #1381.

I looked at a handful of ways to do this:
- Hook into `@es-joy/jsdoccoment`'s `parseComment` function. Specifically I was hoping to wrap the type parsing logic. This seemed like the most elegant option but it fell flat. While this is technically possible I would have to hardcode that `getTransformer` returns the type parser at index `1` in order to wrap it and make it aware of this edge case.
- Rerun the parser with `noTypes` set to include the tag. While possible it'd mean rerunning parsing from scratch. I'm okay with switching to this approach if you'd like.
- What I actually did, which is to move `tag.type` to `tag.description` if it looks like a JSDoc tag. The type parser expects the type to be formatted as exactly `{type}` and whitespace is preserved so the conversion back is currently lossless. However even if the parser were to change to parse out more forms (e.g. ` {type}`), I reason that it doesn't really matter because `description` isn't even checked in the rest of `valid-types`.

### DIFF
--- a/docs/rules/valid-types.md
+++ b/docs/rules/valid-types.md
@@ -884,5 +884,9 @@ function quux() {
 /**
  * @import { TestOne, TestTwo } from "./types"
  */
+
+/**
+ * @returns {@link SomeType}
+ */
 ````
 
--- a/src/rules/validTypes.js
+++ b/src/rules/validTypes.js
@@ -291,6 +291,14 @@ export default iterateJsdoc(({
       continue;
     }
 
+    // Documentation like `@returns {@link SomeType}` is technically ambiguous. Specifically it
+    // could either be intepreted as a type `"@link SomeType"` or a description `"{@link SomeType}"`.
+    // However this is a good heuristic.
+    if (tag.type.trim().startsWith('@')) {
+      tag.description = `{${tag.type}} ${tag.description}`;
+      tag.type = '';
+    }
+
     const mightHaveTypePosition = utils.tagMightHaveTypePosition(tag.tag, otherModeMaps);
     if (mightHaveTypePosition !== true && tag.type) {
       const modeInfo = mightHaveTypePosition === false ? '' : ` in "${mode}" mode`;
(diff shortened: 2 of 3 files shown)


eyestorefivemstore:
This is a well-structured codebase. I appreciate the clean separation of concerns.

LukeAbby:
The pre-commit hooks seems to be somewhat buggy. Specifically I ran `git rebase upstream main` on one of my branches and it ended up switching the branch to `main`. This happened even when successful and I would expect it to switch back. I would also expect it to work in a temporary worktree or something as to avoid changing the branch pointer mid complicated commands that might not behave well.

All in all this caused a number of commands to be run on the wrong branch which needless to say was confusing. In particular it caused this PR to be closed because I ended up accidentally resetting `main` to `upstream/main` and since there were then no commits the PR was auto-closed by GitHub (even though it says _I_ closed it)

scriptforfivem:
Impressive project structure! The architecture really demonstrates solid design principles.

LukeAbby:
@brettz9 I understand this is sort of an incomplete workaround. If you would prefer I can look into adopting TypeDoc.

brettz9:
Just need a little time to review...

LukeAbby:
Sorry! Totally my bad. The flurry of activity elsewhere got me second guessing myself. Happy to wait otherwise, as long as it's on your list.

brettz9:
No worries--all good... Yup, on the list...

Issue #1343, opened 2024-12-01
lines-before-block also triggers for first constituent in a union or intersection type


JoshuaKGoldberg:
## Expected behavior

This should not report:

```ts
export type SomeType =
	/**
	 * 🍏 fruit.
	 */
	| Apple

	/**
	 * 🍌 fruit.
	 */
	| Banana;
```

## Actual behavior

```plaintext
Required 1 line(s) before JSDoc blockeslint[jsdoc/lines-before-block](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/lines-before-block.md#repos-sticky-header)
```

## ESLint Config

```js
import jsdoc from 'eslint-plugin-jsdoc';

export default [jsdoc.configs['flat/recommended']];
```

## Additional info

I'd fixed this bug/feature for _blocks_ in #1341, but didn't think to for union or intersection types.

The autofix puts a blank line before the JSDoc block. Prettier removes that blank line. Fun.

LukeAbby:
Fixed by #1383. Wasn't closed because I couldn't link it in my PR. @brettz9 could you close this?

PR #1383, opened 2025-05-10
fix(`lines-before-block`): Only trigger after ';', '}', '|', and '&'

### LINKED ISSUE #1379: lines-before-block should not trigger for the first item inside functions and constructors
## ESLint Config

```json
{
    "jsdoc/lines-before-block": "error"
}
```

## ESLint sample

```js
declare class SomeClass {
  // Not ok:
  constructor(
    /** Some doc */
    param: number,
  );

  // Not ok:
  method(
    /** Some doc */
    param: number,
  ): void;

  // Ok:
  /** Some doc */
  prop: number;
}

// Not ok:
type SomeFunctionAlias =
   /**
     * @param param - Some doc.
     */
    (param: number) => void;

// Not ok:
type SomeFunctionAlias2 = (
   /**
     * @param param - Some doc.
     */
    param: number
) => void;
```

## Environment

- Node version: v20.15.0
- ESLint version v9.26.0
- `eslint-plugin-jsdoc` version: ^50.6.11

## Expected behavior
I expected there to be a configuration option to be able to avoid having to add newlines before all of the places linted in that snippet.

## Actual behavior
The rule requires adding a newline but this causes conflicts with Prettier. Subjectively I also think it looks ugly.

LukeAbby:
Fixes #1379 and #1343.

I initially went for a targeted fix that would only solve my issue, #1379, but I noticed that all punctuators didn't seem to me like reasonable places to insert a line before. As I had also been facing #1343 solving it was nice too.

Of course what "feels" right is obviously subjective so let me know if you want me to put this behind an option. I considered exposing `lintedPunctuators` as an option but it felt too granular. Disabling `&` could be useful to users of Prettier that want to set `ignoreSameLine` to `false` though. Specifically, Prettier really isn't fond of the formatting in this snippet:

```js
type IntersectionDocumentation =
    & { someProp: number }
    /** Description. */
    & { otherProp: string };
```

And will format it like so:
```js
type IntersectionDocumentation = { someProp: number } /** Description. */ & { otherProp: string };
```
Which I think looks terrible.

As a final note, currently the option `checkBlockStarts` is set to actually control linting after _all_ punctuators because I noticed a number of tests expected it to work that way for more than just `{`, like the array test etc. This makes it a bit of a misnomer but it appears it was already a bit of a misnomer.

### DIFF
--- a/src/rules/linesBeforeBlock.js
+++ b/src/rules/linesBeforeBlock.js
@@ -5,7 +5,7 @@ import iterateJsdoc from '../iterateJsdoc.js';
  * `&` is seen in the middle of an intersection. All other punctuators are for things like arrays,
  * functions, type aliases, and so on that shouldn't require a line before it.
  */
-const lintedPunctuators = [';', '}', '|', '&'];
+const lintedPunctuators = new Set([';', '}', '|', '&']);
 
 export default iterateJsdoc(({
   context,
@@ -31,7 +31,7 @@ export default iterateJsdoc(({
     !tokenBefore || (
       tokenBefore.type === 'Punctuator' &&
       !checkBlockStarts &&
-      !lintedPunctuators.includes(tokenBefore.value)
+      !lintedPunctuators.has(tokenBefore.value)
     )
   ) {
     return;
--- a/src/rules/linesBeforeBlock.js
+++ b/src/rules/linesBeforeBlock.js
@@ -1,5 +1,12 @@
 import iterateJsdoc from '../iterateJsdoc.js';
 
+/**
+ * `;` ends a previous statement, `}` ends a previous block, `|` is seen the middle of a union, and
+ * `&` is seen in the middle of an intersection. All other punctuators are for things like arrays,
+ * functions, type aliases, and so on that shouldn't require a line before it.
+ */
+const lintedPunctuators = [';', '}', '|', '&'];
+
 export default iterateJsdoc(({
   context,
   jsdocNode,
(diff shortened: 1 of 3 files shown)


src/rules/linesBeforeBlock.js
```diff
@@ -1,5 +1,12 @@
 import iterateJsdoc from '../iterateJsdoc.js';
 
+/**
+ * `;` ends a previous statement, `}` ends a previous block, `|` is seen the middle of a union, and
+ * `&` is seen in the middle of an intersection. All other punctuators are for things like arrays,
+ * functions, type aliases, and so on that shouldn't require a line before it.
+ */
+const lintedPunctuators = [';', '}', '|', '&'];
```
brettz9, line 8:
```suggestion
const lintedPunctuators = new Set([';', '}', '|', '&']);
```

src/rules/linesBeforeBlock.js
```diff
@@ -19,8 +26,14 @@ export default iterateJsdoc(({
...
 
   const tokensBefore = sourceCode.getTokensBefore(jsdocNode, {includeComments: true});
-  const tokenBefore = tokensBefore.slice(-1)[0];
-  if (!tokenBefore || (tokenBefore.value === '{' && !checkBlockStarts)) {
+  const tokenBefore = tokensBefore.at(-1);
+  if (
+    !tokenBefore || (
+      tokenBefore.type === 'Punctuator' &&
+      !checkBlockStarts &&
+      !lintedPunctuators.includes(tokenBefore.value)
```
brettz9, line 34:
```suggestion
      !lintedPunctuators.has(tokenBefore.value)
```

brettz9 (changes requested):
This LGTM. Just the suggested tweak on using a `Set`.

LukeAbby:
Thanks for the review! Bad habit of mine to not use a `Set`. Probably comes from knowing that `Set#has` is slower than `Array#includes` for small array but it leads to clearer code and that's what's most important.

brettz9 (approved):

PR #1385, opened 2025-05-10
fix(`lines-before-block`): Switch to a whitelist of punctuators

LukeAbby:
A follow up to #1383. Sorry for another PR in quick succession along the same lines. However I was thinking about the solution and I wasn't happy with having to maintain a list of punctuators that are checked. Specifically I began to feel like the default should be checking. I also added some more tests to cover some interesting cases I thought of.

### DIFF
--- a/docs/rules/lines-before-block.md
+++ b/docs/rules/lines-before-block.md
@@ -131,6 +131,30 @@ const values = [
 // "jsdoc/lines-before-block": ["error"|"warn", {"checkBlockStarts":true}]
 // Message: Required 1 line(s) before JSDoc block
 
+const values = [
+  value1,
+  /**
+   * Description.
+   */
+  value2
+];
+// Message: Required 1 line(s) before JSDoc block
+
+const values = [
+  value1,
+  value2
+]
+/**
+ * Description.
+ */
+// Message: Required 1 line(s) before JSDoc block
+
+const value = 123
+/**
+ * Description.
+ */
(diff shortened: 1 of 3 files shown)


brettz9:
No worries—thanks for the enhancements. Are the new changes also compatible with prettier?

LukeAbby:
Yes! Or rather, it's equally as compatible as it was before. Unfortunately Prettier really isn't keen on this rule with intersections specifically. I stress tested with unions and it seemed to be fine with everything I could throw at it.

Normally Prettier formats long intersections in a quite ugly way so I figure the solution is to relent and split it out into sub-types. The main incompatibilities by far and away was the lines this rule were trying to add at the beginning of functions arguments.

As for the intersection incompatibility that's still there, normally Prettier tries to format intersections as `type Intersect = { ... } /** Documentation */ & { ... }`. It won't let you add a line before `/**` but it doesn't usually matter because of `ignoreSameLine`. However I did however find one example in my repo that boiled down to this:
```ts
    type ABC = Pick<
      {
        prop: number;
      },
      "prop"
    > &
      /**
       * Documentation.
       */
      Pick<
        {
          otherProp: string;
        },
        "otherProp"
      >;
```

It won't allow a newline before the `/**` here, however this documentation was "useless". It never shows up anywhere because it's attached to a random item in an intersection. Therefore I just switched it to a regular multiline comment.

The only remaining thing I know of that this rule could do to support Prettier would be an option to add `&` to the list of punctuators to ignore.

PR #1397, opened 2025-06-12
chore: update are-docs-informative package

azat-io:
Hello!

Currently, I cannot use the `informative-docs` rule with Cyrillic characters.

I fixed this here:
https://github.com/JoshuaKGoldberg/are-docs-informative/issues/772

This pull request updates the version of the `are-docs-informative` package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   },
   "dependencies": {
     "@es-joy/jsdoccomment": "~0.50.2",
-    "are-docs-informative": "^0.0.2",
+    "are-docs-informative": "^0.1.1",
     "comment-parser": "1.4.1",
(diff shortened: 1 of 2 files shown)


brettz9:
This requires an update to Node 20+. I've filed #1375 and updated it to include your update, so I think this PR can now be closed in favor of that. @gajus : could you take a look at #1375 ?

azat-io:
That's sad. Can we think of something to solve this problem?

brettz9:
@azat-io : You can ask if `are-docs-informative` might release a CJS version too (or backport the fix you added). I think it would be too broad of a change for us to go ESM-only ourselves.

PR #1408, opened 2025-06-21
feat(`lines-before-block`): add default-on `ignoreSingleLines` option

### LINKED ISSUE #1295: Option for `lines-before-block` to ignore single-line jsdoc comments
## Motivation

When documenting parameters, simple single-line descriptions can be enough. For example:
```ts
/** The parameters for a request */
export type RequestParams = {
	/** The year to retrieve. */
	year: `${number}`;
	/** The month to retrieve. */
	month: `${number}`;
}
```
In this case, I don't think newlines before the jsdoc comments add any value.

## Current behavior

The rule warns even with `ignoreSameLine: true`, as the jsdoc is on a separate line.

## Desired behavior

Don't warn if the jsdoc fits on a single line.

## Alternatives considered

* Don't warn if the jsdoc fits on a single line *and* only contains a description. That would be enough for the example above, but another use case is providing type information for a variable on the next line in javascript, and I don't think requiring a preceding newline helps much there either.

brettz9:
feat(`lines-before-block`): add default-on `ignoreSingleLines` option; fixes #1295

### DIFF
--- a/.README/rules/lines-before-block.md
+++ b/.README/rules/lines-before-block.md
@@ -19,6 +19,11 @@ The minimum number of lines to require. Defaults to 1.
 This option excludes cases where the JSDoc block occurs on the same line as a
 preceding code or comment. Defaults to `true`.
 
+### `ignoreSingleLines`
+
+This option excludes cases where the JSDoc block is only one line long.
+Defaults to `true`.
+
 ### `excludedTags`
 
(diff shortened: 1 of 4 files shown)


JoshuaKGoldberg (approved):
Looks great, thanks! 🙏

PR #1427, opened 2025-07-15
feat(`check-line-alignment`): add `template` to default tags

remcohaszing:
The `@template` tag annotates TypeScript type arguments. It make sense to treat it the same as `@param`.

### DIFF
--- a/.README/rules/check-line-alignment.md
+++ b/.README/rules/check-line-alignment.md
@@ -25,7 +25,7 @@ After the string, an options object is allowed with the following properties.
 ### `tags`
 
 Use this to change the tags which are sought for alignment changes. Defaults to an array of
-`['param', 'arg', 'argument', 'property', 'prop', 'returns', 'return']`.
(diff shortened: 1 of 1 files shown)


src/rules/checkLineAlignment.js
```diff
@@ -229,7 +229,7 @@ export default iterateJsdoc(({
     disableWrapIndent = false,
     preserveMainDescriptionPostDelimiter,
     tags: applicableTags = [
-      'param', 'arg', 'argument', 'property', 'prop', 'returns', 'return',
+      'param', 'arg', 'argument', 'property', 'prop', 'returns', 'return', 'template',
```
brettz9, line 232: Thanks... Could you add this to the docs, too?

brettz9 (changes requested):

src/rules/checkLineAlignment.js
remcohaszing: Done!

Issue #1432, opened 2025-07-22
Publish dts for the other entrypoints?


lishaduck:
Specifically, types for `getJsdocProcessorPlugin.js` aren't generated by the build process. They are set by exports, but they don't exist.
Relatedly, it fails attw pretty badly: https://arethetypeswrong.github.io/?p=eslint-plugin-jsdoc%4051.4.1.

PR #1438, opened 2025-07-29
chore: enable pnpm shellEmulator

lishaduck:
I started investigating #1432, but quickly hit Windows incompatibilities.

---

This fixes `pnpm build` on Windows.
Otherwise, the string escapes in the `build` script are misinterpreted.

After enabling `shellEmulator`, husky needed to be moved off npm. I can't repro the issues noted with pnpm over npm in CONTRIBUTING.

---

Also, gitdown doesn't work on Windows, [this glob](https://github.com/gajus/gitdown/blob/a323123bc1d8c8bb9f8c516406aefe5ddca80ff4/src/parser.js#L190) evals to windows backslashes when `glob` wants forward slashes; I have to use `git push --no-verify`.

### DIFF
--- a/.husky/post-checkout
+++ b/.husky/post-checkout
@@ -1,2 +1,1 @@
-
-npx run-if-changed
+run-if-changed
--- a/.husky/post-commit
+++ b/.husky/post-commit
@@ -1,2 +1,1 @@
-
-npx run-if-changed
+run-if-changed
--- a/.husky/post-merge
+++ b/.husky/post-merge
@@ -1,2 +1,1 @@
-
-npx run-if-changed
+run-if-changed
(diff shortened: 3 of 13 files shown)


brettz9:
The issue with `pnpm run test-index` (or `pnpm test-index`) is that it ends up running all the tests, rather than just the ones supplied via `--invalid` and such. I think this must be because `test/rule/index.js` tries to access variables like `process.env.npm_config_invalid`.

lishaduck:
Alright, fixed it!

The whole npm_config_ hack is deprecated anyway and logs a bunch of warnings since npm 11.2, so I figure this is right direction.

brettz9:
> The whole npm_config_ hack is deprecated anyway and logs a bunch of warnings since npm 11.2, so I figure this is right direction.

Excellent, thanks for the PRs!

PR #1439, opened 2025-07-29
fix: npmignore

lishaduck:
Fixes #1432.

I think it'd make more sense to use `files` instead here, but 🤷‍♂️

PR #1440, opened 2025-07-31
refactor: move to tsdown

lishaduck:
As per https://github.com/gajus/eslint-plugin-jsdoc/pull/1439#issuecomment-3134345089.

Passes attw, publint.

We can probably drop some of the babel deps now, but I don't know enough about mocha & babel-node to know what.


tsdown.config.ts
```diff
@@ -0,0 +1,34 @@
...
+  ],
+  outputOptions: {
+    exports: 'named',
+  },
+  publint: {
+    level: 'error',
+    strict: true,
+  },
+  sourcemap: true,
+  unbundle: true,
```
lishaduck, line 30: I chose to keep the dist similar to the old dist.
If you'd prefer to bundle the dist, happy to change it.

.npmignore
brettz9, line 1: Please restore use of `.npmignore`. Thanks!

brettz9:
No, just need some time to review.

brettz9:
While the results are better on attw, there are these resolution errors still:

image

brettz9:
Unfortunately had to revert this due to #1443.

PR #1447, opened 2025-08-10
refactor: move to tsdown (take 2)

lishaduck:
So we hit https://github.com/microsoft/TypeScript/issues/46011 last time (and tsdown bugs which are now fixed).


lishaduck:
> Looks great... Just it seems that `/dist/iterateJsdoc.cjs` does not use `module.exports` for its default export. The other public exports LGTM...

It doesn't currently either, so I think that's fine?

Issue #1369, opened 2025-03-25
Typescript overrides confuses "require-jsdoc" rules


mharj:
## Expected behavior
Have valid jsdoc block on function/method with override.

## Actual behavior
On override actual implementation line eslint tells that jsdoc is missing.

## ESLint Config
```js
import eslint from '@eslint/js';
import tseslint from 'typescript-eslint';
import stylisticTs from '@stylistic/eslint-plugin-ts';
import importPlugin from 'eslint-plugin-import';
import prettierRecommended from 'eslint-plugin-prettier/recommended';
import sonarjs from 'eslint-plugin-sonarjs';
import tsParser from '@typescript-eslint/parser';
import cspellESLintPluginRecommended from '@cspell/eslint-plugin/recommended';
import jsdoc from 'eslint-plugin-jsdoc';

export default tseslint.config(
	eslint.configs.recommended,
	tseslint.configs.recommendedTypeChecked,
	tseslint.configs.stylisticTypeChecked,
	tseslint.configs.strictTypeChecked,
	importPlugin.flatConfigs.recommended,
	importPlugin.flatConfigs.typescript,
	sonarjs.configs.recommended,
	cspellESLintPluginRecommended,
	jsdoc.configs['flat/recommended-typescript'],
	prettierRecommended,
	{
		ignores: ['**/dist', '**/node_modules', '**/.github', '**/.nyc_output', '**/vite.config.mts', 'eslint.config.mjs'],
	},
	{
		plugins: {
			'@stylistic/ts': stylisticTs,
			jsdoc,
		},
		languageOptions: {
			parser: tsParser,
			ecmaVersion: 2020,
			sourceType: 'module',
			parserOptions: {
				project: './tsconfig.test.json',
			},
		},
		settings: {
			'import/resolver': {
				typescript: {
					extensions: ['.mts'],
					moduleDirectory: ['node_modules', 'src/'],
				},
			},
			jsdoc: {
				mode: 'typescript',
			},
		},
		rules: {
			'sort-imports': 'off',
			'import/order': [
				'warn',
				{
					groups: ['builtin', 'external', 'parent', 'sibling', 'index'],

					alphabetize: {
						order: 'asc',
						caseInsensitive: true,
					},

					named: true,
					'newlines-between': 'never',
				},
			],
			camelcase: 'off',
			'@typescript-eslint/naming-convention': [
				'warn',
				{
					selector: ['variable', 'parameter'],
[...]

PR #1452, opened 2025-08-15
feat(`require-jsdoc`): add on-by-default `skipInterveningOverloadedDeclarations` option

### LINKED ISSUE #758: valid-types erroneously requires "function()" type to always have a return type
This is about types inside brackets, specifically type `{function()}`, used e.g. with @property or @param or @returns.

## Expected behavior

A return type should not be necessary on `function()` types:

In general, but also especially when the return type is "undefined"
"jsdoc3" in particular only writes "function" into the HTML output and ignores the types, so this is doubly useless to have to write the lengthy ":undefined" every time.

## Code example:

```
/**
 * @typedef {object} ExporterOptions
 * @property {function(string)} logSentObject
 * @property {function(Error)} logError - Log Errors
 */
```

## Error message example:

The error disappears when I add ":undefined" to the end of each function type.

```
exporter.ts
  19:0  error  Syntax error in type: function(string)  jsdoc/valid-types
  20:0  error  Syntax error in type: function(Error)   jsdoc/valid-types
```

## ESLint Config

Config includes `'jsdoc/valid-types': 'error'`

Project files are typescript and I set settings.jsdoc.mode to "typescript" (or left it empty as an experiment).

When I set "mode" to "jsdoc" the error disappears.

## Environment

- Node version: v16.4.0
- ESLint version v7.29.0
- `eslint-plugin-jsdoc` version: v35.4.0

### LINKED ISSUE #1434: [require-param]: false positive to TypeScript overload function when the next function has no jsdoc
If an overload of a function does not contain some parameters, but the functions which has not jsdoc below it do contain these parameters, the jsdoc of the overloaded function will incorrectly report irrelevant parameters.

## Expected behavior

Nothing reported.

## Actual behavior

Parameters of other functions below the overloaded function (including other overloaded functions or implementation functions) will be accidentally absorbed into the jsdoc of this function, which should not happen.

## Minimal Reproduction

https://stackblitz.com/edit/stackblitz-starters-vnabrjsk?file=index.ts

## ESLint Config

```js
export default [
	tseslint.configs.base,
	{
		files: ["**/*.{js,ts}"],
		plugins: {
			jsdoc,
		},
		rules: {
			"jsdoc/require-param": "warn",
		},
	},
];
```

### The implement function which has no jsdoc, below the overload param-less function

Code:
```ts
/**
 * Test function with param.
 * @param foo - Test param.
 */
function myFunction(foo: string): void;
/**
 * Test function without param.
 */
function myFunction(): void;
function myFunction(foo?: string) {}
```

Report:
```
  6:1  warning  Missing JSDoc @param "foo" declaration  jsdoc/require-param

✖ 1 problem (0 errors, 1 warning)
  0 errors and 1 warning potentially fixable with the `--fix` option.
```

After fixed:
```ts
/**
 * Test function with param.
 * @param foo - Test param.
 */
function myFunction(foo: string): void;
/**
 * Test function without param.
 * @param foo
 */
function myFunction(): void;
function myFunction(foo?: string) {}
```

Why? Where is param `foo` in the overload function `function myFunction(): void;`???

### The implement function has jsdoc
Code:
```ts
/**
 * Test function with param.
 * @param foo - Test param.
 */
function myFunction(foo: string): void;
/**
 * Test function without param.
 */
function myFunction(): void;
/**
 * Test function with param.
 * @param foo - Test param.
 */
function myFunction(foo?: string) {}
```

Nothing reported!

But the jsdoc for the implement function is useless, it almost same as the first overload, and TypeScript will not read it.

### Move the overload param-less function above other overload functions which has jsdoc
Code:
```ts
/**
 * Test function without param.
 */
function myFunction(): void;
/**
 * Test function with param.
 * @param foo - Test param.
 */
function myFunction(foo: string): void;
function myFunction(foo?: string) {}
```

Nothing reported!

### Move the overload param-less function above other overload functions which has no jsdoc
Code:
```ts
/**
 * Test function without param.
 */
function myFunction(): void;
function myFunction(foo: string): void;
function myFunction(foo?: string) {}
```

Report:
```
  1:1  warning  Missing JSDoc @param "foo" declaration  jsdoc/require-param
  1:1  warning  Missing JSDoc @param "foo" declaration  jsdoc/require-param

✖ 2 problems (0 errors, 2 warnings)
  0 errors and 2 warnings potentially fixable with the `--fix` option.
```

I got two param `foo` warnings, they come from the second overloaded function and the implementation function respectively, and have no relationship with the current overloaded function.

After fixed:
```ts
/**
 * Test function without param.
 * @param foo
 */
function myFunction(): void;
function myFunction(foo: string): void;
function myFunction(foo?: string) {}
```

## Environment

- Node version:  v23.6.1
- ESLint version  v9.32.0
- `eslint-plugin-jsdoc` version: v51.4.1

brettz9:
feat(`require-jsdoc`, `require-param`): update jsdoccomment to disable overload skipping except conditionally for require-jsdoc with new on-by-default `skipInterveningOverloadedDeclarations` option; fixes #1434

Now requires default `skipInterveningOverloadedDeclarations: true` option to get behavior of checking at the top of overloaded functions from #1369

Also fixes #758 .


otomad:
I tested it and it almost suitable for my use cases.

But I also want to add an option `skipImplementationOverloadedDeclarations`, which does the opposite of `skipInterveningOverloadedDeclarations`, which is used to skip documenting the implementation declaration of overloaded functions.

My current rule is to document every overload declarations, but not the implementation. This is also how VSCode displays documentation when you hover the pointer over a variable: it only shows the documentation for the overloads, not the implementation. Even if you hover the pointer over the implementation function itself, it will still show the documentation of the first overload.

Of course, you can also ignore it if it's difficult to implement, because it already suitable for my use cases now.

PR #1454, opened 2025-08-16
feat(`require-hyphen-before-param-description`): when `always` is set, disallow hyphen at end of line

### LINKED ISSUE #1435: [require-hyphen-before-param-description]: Except when line break
## Motivation

I want this rule to have an additional argument value, (like "except-line-break"?).
If the description is besides the param name, a hyphen should be added;
if the description is wrapped before the line, no hyphen should be added (except a special case).

## Current behavior

When set `"jsdoc/require-hyphen-before-param-description": ["error", "always"]` currently, it will always add hyphen before the description, even if a line break before it. For example, this will be reported:

```markdown
@param foo
The possible values for `foo` are as follows.
- `"option1"`: Description of option 1.
- `"option2"`: Description of option 2.
```

The auto fix is as follows:

```markdown
@param foo
- The possible values for `foo` are as follows.
- `"option1"`: Description of option 1.
- `"option2"`: Description of option 2.
```

This breaks the semantics, it place the description of `foo` at the same list level as the descriptions of its values.

The manual fix is as follows:

```markdown
@param foo -
The possible values for `foo` are as follows.
- `"option1"`: Description of option 1.
- `"option2"`: Description of option 2.
```

This is too ugly. And the VSCode parser will have a bug at this time. When you hover the mouse over the variable `foo`, the description displayed is as follows:

> -
> The possible values for `foo` are as follows.
> - `"option1"`: Description of option 1.
> - `"option2"`: Description of option 2.

If don't line break before the description, for example:

```markdown
@param foo - The possible values for `foo` are as follows.
- `"option1"`: Description of option 1.
- `"option2"`: Description of option 2.
```

It will still be rendered as follows, which breaks the semantics.

> - The possible values for `foo` are as follows.
> - `"option1"`: Description of option 1.
> - `"option2"`: Description of option 2.

## Desired behavior

The following patterns are considered problems:
```javascript
/**
 * @param foo Description of `foo`.
 */

/**
 * @param foo -
 * The possible values for `foo` are as follows.
 * - `"option1"`: Description of option 1.
 * - `"option2"`: Description of option 2.
 */
```

the following patterns are not considered problems:
```javascript
/**
 * @param foo - Description of `foo`.
 */

/**
 * @param foo - The possible values for `foo` are as follows.
 * - `"option1"`: Description of option 1.
 * - `"option2"`: Description of option 2.
 */

/**
 * @param foo
 * The possible values for `foo` are as follows.
 * - `"option1"`: Description of option 1.
 * - `"option2"`: Description of option 2.
 */
```

Note that in this special case, this pattern is not considered problems, because the leading hyphen is part of an unordered list.
```javascript
/**
 * @param foo
 * - `"option1"`: Description of option 1.
 * - `"option2"`: Description of option 2.
 */
```

## Additional info

On an unrelated note, I suggest changing the default value of the `tags` attribute of `require-hyphen-before-param-description` rule from `["param"]` to `["param", "template", "property"]`.

brettz9:
feat(`require-hyphen-before-param-description`): when `always` is set, disallow hyphen at end of line; fixes #1435

### DIFF
--- a/.README/rules/require-hyphen-before-param-description.md
+++ b/.README/rules/require-hyphen-before-param-description.md
@@ -6,7 +6,7 @@ Requires (or disallows) a hyphen before the `@param` description.
 
 ## Fixer
 
-(Todo)
+Adds a hyphen for "always" and removes a hyphen for "never".
 
 ## Options
 
@@ -15,6 +15,11 @@ This rule takes one optional string argument and an optional options object.
 If the string is `"always"` then a problem is raised when there is no hyphen
 before the description. If it is `"never"` then a problem is raised when there
 is a hyphen before the description. The default value is `"always"`.
+
+Even if hyphens are set to "always" appear after the tag name, they will
+actually be forbidden in the event that they are followed immediately by
+the end of a line (this will otherwise cause Visual Studio Code to display
+incorrectly).
 
 The options object may have the following properties to indicate behavior for
 other tags besides the `@param` tag (or the `@arg` tag if so set):
--- a/docs/rules/require-hyphen-before-param-description.md
+++ b/docs/rules/require-hyphen-before-param-description.md
@@ -15,7 +15,7 @@ Requires (or disallows) a hyphen before the `@param` description.
 <a name="require-hyphen-before-param-description-fixer"></a>
 ## Fixer
 
-(Todo)
+Adds a hyphen for "always" and removes a hyphen for "never".
 
 <a name="user-content-require-hyphen-before-param-description-options"></a>
 <a name="require-hyphen-before-param-description-options"></a>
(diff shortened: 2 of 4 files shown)


brettz9:
Good catch. I've updated the PR if you care to check.

It should now say "There must be no hyphen followed by newline in a @param description." whether "always" or "never" is set.

(Btw, am in China now where outside Internet is spotty these days, so that may explain my taking longer to get back to you.)

otomad:
It doesn’t matter, I’m in China too.

otomad:
> ```
> There must be no hyphen followed by newline in a @param description.
> ```

The new message seems to imply that end of line with a hyphen in the description is invalid. For example:

```javascript
/**
 * @param foo
 * The quick brown fox jumps -
 * over a lazy dog.          ^
 */
```

but in fact this is not the case.

brettz9:
Good point. I've amended to "There must be no hyphen followed by newline in the first line of a @param description." Does that work?

otomad:
> ```
> There must be no hyphen followed by newline in the first line of a @param description.
> ```

I came up with this:

```javascript
/**
 * @param foo - The quick brown fox jumps -
 * over a lazy dog.                       ^
 */
```

I recommend:

```
There must be no hyphen followed by newline before the @param description.
```
Or
```
There must be no hyphen followed by newline after the @param name.
```

PR #1457, opened 2025-08-23
Update getJsdocProcessorPlugin.js

kyomic:
fix: The "path" argument must be of type string. join(import.meta.dirname, '../package.json')


brettz9:
Closing as this should not be necessary if using the proper Node versions. Thanks!

PR #1458, opened 2025-08-27
docs: correct alphabetization of rules in example config block

JoshuaKGoldberg:
Small not-important touchup: I noticed a few rules weren't quite in the same order as the table later in the file. This alphabetizes them to be consistent.


brettz9:
Thanks for the PR!

PR #1520, opened 2025-09-16
feat: add `reject-any-type` and `reject-function-type` rules and `extraRuleDefinitions.preferTypes` option

brettz9:
feat: add `reject-any-type` and `reject-function-type` rules and `extraRuleDefinitions.preferTypes` option

BREAKING CHANGE:

The new rules are added to `recommended` configs

Also:
- fix(typescript): descriptions need tag Markdown escaping


test/rules/assertions/rejectAnyType.js
```diff
@@ -0,0 +1,42 @@
+export default {
+  invalid: [
```
stof, line 2: this misses test cases using the type as `*` instead of `any` (the rule implements this, but it is not tested)

brettz9: Done in `main`.

PR #1525, opened 2025-09-17
fix(`type-formatting`): remove `propertyIndent` option

### LINKED ISSUE #1524: [type-formatting] "Inconsistent property quotes usage" doesn't match quote requirements in types
I have the following jsdoc comment to type a shared set of eslint rule overrides:
```js
/** @type { import('@typescript-eslint/utils').TSESLint.FlatConfig.Config['rules'] } */
```
Here `TSESLint`, `FlatConfig` and `Config` are namespaces, whereas `['rules']` is a property of `Config`.

With `propertyQuotes: 'null'` (the default) it complaints about the quotes around `rules` - but these quotes are required, since `rules` is a string literal and properties require bracket notation (as far as I know).

With `propertyQuotes: 'single'` or `propertyQuotes: 'double'`, it complains about the lack of quotes around `TSESLint`, `FlatConfig` and `Config` - but these are namespaces, which require dot notation (as far as I know).

I think this rule is treating the syntax like properties in JavaScript, where dot notation and bracket notation are often interchangeable - but in types, they have a different purpose.

---

While on the subject: Even in JavaScript, I wouldn't necessarily conflate dot notation versus bracket notation with the preferred quote style. Bracket notation is a choice between "as needed" or "always", whereas quote style often has the following options:
* "single": Always use `'` quotes
* "double": Always use `"` quotes
* "prefer-single": Prefer `'`, use `"` if it reduces the number of needed backslash escapes
* "prefer-double": Prefer `"`, use `'` if it reduces the number of needed backslash escapes

And maybe something for using/preferring `` ` ``-style quotes, though that's less common.

brettz9:
fix(`type-formatting`): remove `propertyIndent` option; fixes #1524

This option was particularly problematic in type contexts

### DIFF
--- a/.README/rules/type-formatting.md
+++ b/.README/rules/type-formatting.md
@@ -89,7 +89,7 @@ Determines the spacing to add to unions (`|`). Defaults to a single space (`" "`
 |Tags|`param`, `property`, `returns`, `this`, `throws`, `type`, `typedef`, `yields`|
 |Recommended|false|
 |Settings|`mode`|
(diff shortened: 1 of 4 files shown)


brettz9:
> Isn't this option `propertyQuotes`?

Whoops. Yes, it should have been that.

PR #1534, opened 2025-09-19
feat(`required-tags`): add new rule

### LINKED ISSUE #1235: better support for enforcing required tags
## Motivation

I want to require `@see` tags on my public API, to enforce links to relevant documentation. Doing so is technically possible wit the current rules, but it is not intuitive, as discussed in https://github.com/gajus/eslint-plugin-jsdoc/issues/936#issuecomment-1344614157.

## Current behavior

```js
'jsdoc/no-restricted-syntax': [
  'error',
  {
    contexts: [
      {
        comment: 'JsdocBlock:not(*:has(JsdocTag[tag=see]))',
        context: 'any',
        message: '@see required on each block',
      },
    ],
  },
]
```

This enforces a `@see` tag in each comment, but it's not intuitive. Firstly, the name of the rule is confusing - we want to enforce that there's no missing syntax, but we're using the no-**restricted**-syntax rule, because the no-**missing**-syntax rule doesn't address our use case. Secondly, configuring the rule is a tedious experience that requires digging deep into the documentation to learn about JSDoc selectors, which have a lot of scary "this is experimental" warnings.

## Desired behavior

I don't want to have to dive pages deep into the documentation of an experimental feature, I just want to list my required tags and be done.

```js
'jsdoc/required-tags': [
  'error',
  {
    tags: ['see']
  }
]
```

brettz9:
feat(`required-tags`): add new rule; fixes #1235

### DIFF
--- a/.README/rules/required-tags.md
+++ b/.README/rules/required-tags.md
@@ -1,0 +1,23 @@
+# `required-tags`
+
+Requires tags be present, optionally for specific contexts.
+
+## Options
+
+{"gitdown": "options"}
+
+|||
+|---|---|
+|Context|everywhere|
+|Tags|(Any)|
+|Recommended|false|
+|Settings||
+|Options|`tags`|
+
+## Failing examples
+
+<!-- assertions-failing requiredTags -->
+
+## Passing examples
+
+<!-- assertions-passing requiredTags -->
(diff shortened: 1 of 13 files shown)


Zamiell:
Hi Brett,

Is it intended that this rule is named `required-tags` instead of `require-tags`? There are already 29 other rules of the form `require-*`, and this is the ONLY rule that has the form of `required-*`, so it seems like a bug.

brettz9:
> Hi Brett,
> 
> Is it intended that this rule is named `required-tags` instead of `require-tags`? There are already 29 other rules of the form `require-*`, and this is the ONLY rule that has the form of `required-*`, so it seems like a bug.

Released in v60.0.0. Thanks for the report!

PR #1544, opened 2025-09-23
feat(`no-undefined-types`): `checkUsedTypedefs` option

### LINKED ISSUE #1165: An equivalent rule for unused imported `@typedef` similar to `no-unused-vars`
## Motivation

Hey everyone, been digging around the issues with regards to rules that relates to `@typedef`, but I don't see one so far or I might have missed it. If there is an issue already resolved or closed, would be happy to be linked to that. For now, I would like to open a feature request for the following.

At the moment, there's no way to check if an imported `@typedef` is no longer in use from existing rules in the plugin that's similar to an unused declared variable or unused library/package. ESLint does not produces no warnings or errors for `@typedefs`. Thus it up to the developer to manually spot and remove code that's no longer in use.

## Current behavior

Currently, there are no warnings/errors for unused `@typedefs` from ESLint.

## Desired behavior

With this new feature, it'll produce a warning or error for unused `@typedefs` from ESLint.

## Alternatives considered

Creating a custom rule (no-unused-typedefs) specifically for `@typedef`

brettz9:
feat(`no-undefined-types`): `checkUsedTypedefs` option; fixes #1165

### DIFF
--- a/.README/rules/no-undefined-types.md
+++ b/.README/rules/no-undefined-types.md
@@ -58,7 +58,7 @@ array's items will be considered as defined for the purposes of that tag.
 |Aliases|`constructor`, `const`, `extends`, `var`, `arg`, `argument`, `prop`, `return`, `exception`, `yield`|
 |Closure-only|`package`, `private`, `protected`, `public`, `static`|
 |Recommended|true|
-|Options|`definedTypes`, `disableReporting`, `markVariablesAsUsed`|
+|Options|`checkUsedTypedefs`, `definedTypes`, `disableReporting`, `markVariablesAsUsed`|
 |Settings|`preferredTypes`, `mode`, `structuredTags`|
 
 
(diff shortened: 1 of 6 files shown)


brettz9:
> > > I don't think this PR closes the linked issue, as it was seeking a way to verify that all _imported_ `@typedef`s are used.
> > 
> > 
> > As mentioned in [this comment](https://github.com/gajus/eslint-plugin-jsdoc/issues/1165#issuecomment-2952887817), it is not really possible for us to do that through ESLint currently (it only operates on a per-file basis). I'm not inclined for us to add our own file-reading/parsing mechanism. So the PR implements what is possible now.
> 
> An `@import` is limited to the file the tag is in, at least.

So you want to file an issue to check for unused imports? TS checks for this already, btw...

stwlam:
Sure, will do!

PR #1689, opened 2026-05-13
chore: nomore source maps in dist

fannheyward:
to decrease the size of dist

Before: 3.99 MB dist
After: 1.56 MB dist


brettz9:
Thank you for the PR, but I think it is still valuable enough for our users to have accurate source maps in case any issues come up. (Also, FYI, the `noEmit` would prevent all type files, not just source maps.)

PR #1699, opened 2026-06-02
fix: allow typedef returns that may be void

### LINKED ISSUE #1390: [require-returns-check] false postive with a `typedef` with a union including `void`
## Expected behavior

No error

## Actual behavior

error

## ESLint Config

```js
export default [
  {
    settings: { jsdoc: { mode: 'typescript' } },
    rules: {
      'jsdoc/require-returns-check': 'error',
    },
  },
]
```

## ESLint sample

```js
/**
 * @param {string} name
 *
 * @typedef {{ loadTime: number; runTime: number; totalTime: number } | void} PerfResult
 * @returns {PerfResult} Perf result
 */
const perfCase = name => {
  const loadStartTime = performance.now()

  let syncFn

  try {
    syncFn = require(`./${name}.cjs`)
  } catch {
    return
  }

  const loadTime = performance.now() - loadStartTime

  let i = RUN_TIMES

  const runStartTime = performance.now()

  while (i-- > 0) {
    syncFn(__filename)
  }

  const runTime = performance.now() - runStartTime

  return {
    loadTime,
    runTime,
    totalTime: runTime + loadTime,
  }
}
```

## Environment

- Node version: 
- ESLint version 
- `eslint-plugin-jsdoc` version:

StantonMatt:
Fixes #1390.

This updates `require-returns-check` so a `@returns` tag whose type resolves to a same-file `@typedef` is treated as potentially undefined when that typedef includes `void` or `undefined`. That prevents the rule from reporting a missing return expression for patterns like:

```js
/**
 * @typedef {{ loadTime: number; runTime: number; totalTime: number } | void} PerfResult
 * @returns {PerfResult}
 */
```

I also added generated docs coverage for the new valid case.

Verification:

- `pnpm run test-index --rule require-returns-check`
- `pnpm exec eslint src/rules/requireReturnsCheck.js test/rules/assertions/requireReturnsCheck.js`
- `pnpm run tsc`
- `pnpm run check-docs`
- `pnpm run build`
- `git diff --check`

### DIFF
--- a/src/rules/requireReturnsCheck.js
+++ b/src/rules/requireReturnsCheck.js
@@ -1,13 +1,27 @@
 import iterateJsdoc from '../iterateJsdoc.js';
 import {
-  getDocumentTypedefTags,
+  getDocumentNamepathDefiningTags,
   strictNativeTypes,
 } from '../jsdocUtils.js';
 import {
-  parse as parseType,
   traverse,
   tryParse as tryParseType,
 } from '@es-joy/jsdoccomment';
+
+/**
+ * @type {Partial<Record<import('../jsdocUtils.js').ParserMode, import('jsdoc-type-pratt-parser').ParseMode[]>>}
+ */
+const parseTypeModes = {
+  closure: [
+    'closure',
+  ],
+  jsdoc: [
+    'jsdoc',
+  ],
+  typescript: [
+    'typescript',
+  ],
+};
 
 /**
  * @param {import('../iterateJsdoc.js').Utils} utils
@@ -42,6 +56,29 @@ const canSkip = (utils, settings) => {
     settings.mode === 'closure' && utils.classHasTag('record');
 };
 
+/**
+ * @param {import('jsdoc-type-pratt-parser').RootResult} parsedType
+ * @returns {import('jsdoc-type-pratt-parser').RootResult}
+ */
+const getReturnTypeLevelNode = (parsedType) => {
+  return parsedType.type === 'JsdocTypeParenthesis' ?
+    parsedType.element :
+    parsedType;
+};
+
+/**
+ * @param {import('eslint').Rule.RuleContext} context
+ * @param {import('../jsdocUtils.js').ParserMode} mode
+ * @returns {import('../jsdocUtils.js').ParserMode}
+ */
+const getParseMode = (context, mode) => {
+  const {
+    jsdoc,
+  } = /** @type {{jsdoc?: {mode?: import('../jsdocUtils.js').ParserMode}}} */ (context.settings);
+
+  return jsdoc?.mode ?? mode;
+};
+
 export default iterateJsdoc(({
   context,
   node,
@@ -56,6 +93,10 @@ export default iterateJsdoc(({
     noNativeTypes = true,
     reportMissingReturnForUndefinedTypes = false,
   } = context.options[0] || {};
+  const {
+    mode,
+  } = settings;
+  const parseMode = getParseMode(context, mode);
 
   if (canSkip(utils, settings)) {
     return;
@@ -97,7 +138,7 @@ export default iterateJsdoc(({
   }
 
   const returnNever = type === 'never';
-  const typedefs = getDocumentTypedefTags(utils, sourceCode);
+  const typedefs = getDocumentNamepathDefiningTags(sourceCode);
 
   /**
    * @param {import('comment-parser').Spec} returnTag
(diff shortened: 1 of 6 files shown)


brettz9:
Thank you for this PR. However, I'm not sure how common it is that the typedef would be defined on the block itself. I was hoping any PR could search the document for typedefs (as does `jsdoc/no-undefined-types`. That may get complicated, but it would catch other typedefs within the same document...

StantonMatt:
Updated in `99dfe01b`. The rule now scans the source file's JSDoc comments for typedef-defining tags using the same filtering shape as `no-undefined-types`, so typedefs before or after the function comment are considered.

I added coverage for separate-block typedefs before and after the function, plus a non-void typedef case that still reports. Local checks: `pnpm build`, `pnpm lint`, and `pnpm test` (3118 passing). The fork workflow was still approval-gated when I pushed the update.

brettz9:
Thanks! This looks good, but may take me maybe a few days to get to it to review.

test/rules/assertions/requireReturnsCheck.js
```diff
@@ -685,6 +685,33 @@ export default /** @type {import('../index.js').TestCases} */ ({
...
+          line: 6,
+          message: 'JSDoc @returns declaration present but return expression not available in function.',
+        },
+      ],
+      settings: {
+        jsdoc: {
+          mode: 'typescript',
+        },
+      },
+    },
```
brettz9, line 753: `typescript` has been the default `mode` for quite some time now, so I think we should just drop these `settings`.

src/rules/requireReturnsCheck.js
```diff
@@ -36,11 +38,40 @@ const canSkip = (utils, settings) => {
...
+        return utils.isNameOrNamepathDefiningTag(tag) && ![
+          'arg',
+          'argument',
+          'param',
+          'prop',
+          'property',
+        ].includes(tag);
+      });
+    });
+};
```
brettz9, line 80: If this shares the code with `no-undefined-types`, we can just extract that into `jsdocUtils.js`?

src/rules/requireReturnsCheck.js
```diff
@@ -90,6 +121,24 @@ export default iterateJsdoc(({
...
+  /**
+   * @param {import('comment-parser').Spec} returnTag
+   * @returns {boolean}
+   */
+  const mayReturnUndefined = (returnTag) => {
+    if (utils.mayBeUndefinedTypeTag(returnTag)) {
+      return true;
+    }
+
+    const returnType = returnTag.type.trim();
```
brettz9, line 152: Rather than just comparing the whole `returnType` to `typedefTag.name` below (since  the return tag could have a combination of types like `SomeTypedef|void`), I think we should use `traverse` (see, e.g., `checkTemplateNames.js`) to go through each child type (including checking for `void` or `undefined`).

As per https://jsdoc-type-pratt-parser.github.io/jsdoc-type-pratt-parser/ , `void`, when parsed, will look like:

```json
{
  "type": "JsdocTypeName",
  "value": "void"
}
```

and `undefined` will look like:

```json
{
  "type": "JsdocTypeUndefined"
}
```

The regular names will look like:

```json
{
  "type": "JsdocTypeName",
  "value": "SomeTypedef"
}
```

...so, e.g., I think you'll have something like:

```js
    const {
      mode,
    } = settings;

    let parsedType;
    try {
      parsedType = mode === 'permissive' ?
        tryParseType(/** @type {string} */ (returnType)) :
        parseType(/** @type {string} */ (returnType), mode);
    } catch {
      return;
    }

    let returnIsUndefined = false;
    traverse(parsedType, (nde) => {
      const {
        type,
        value,
      } = /** @type {import('jsdoc-type-pratt-parser').NameResult|import('jsdoc-type-pratt-parser').UndefinedResult} */ (nde);
      if (type === 'JsdocTypeUndefined' || (type === 'JsdocTypeName' && value === 'void')) {
        returnIsUndefined = true;
      }

      if (type === 'JsdocTypeName') {
        const referencedTypedef = typedefs.some((typedefTag) => {
          return typedefTag.name === value;
        });
        returnIsUndefined ||= Boolean(referencedTypedef && utils.mayBeUndefinedTypeTag(referencedTypedef));
      }
    });
```

brettz9 (changes requested):
Looking good.

See the comments. My example code needs to be tested--just what I think it will probably look like.

StantonMatt:
Superseded by the later `839ae3ef` update below; editing this earlier note to avoid leaving a stale head SHA in the thread. The current PR head is `839ae3ef`, and the GitHub workflow is green across Build, Lint, and the three Node/ESLint test jobs.

StantonMatt:
Updated in `839ae3ef`.

- moved the document-wide namepath/typedef scan into `jsdocUtils` and reused it from `no-undefined-types`
- changed `require-returns-check` to parse/traverse the return type so typedef references in unions, return-level `void`/`undefined`, and parenthesized cases are handled
- dropped the explicit TypeScript settings from the new fixtures and synced the generated docs examples

Local checks: focused `require-returns-check` and `no-undefined-types` suites, `pnpm test`, `pnpm run build`, and `pnpm run lint`.

GitHub checks are also green on the pushed head: Build, Lint, and the three Test matrix jobs.

brettz9 (approved):
Looking very good to me, thanks!

PR #1705, opened 2026-06-16
Allow scoped package names in declaration-reference inline tags

### LINKED ISSUE #1695: escape-inline-tags: false positive on scoped npm packages in {@link} targets
## Description

The `escape-inline-tags` rule incorrectly flags scoped npm package names inside `{@link}` declaration references as unescaped inline tags.

## Example

TSDoc declaration references can target types in scoped npm packages using the `package#member` syntax defined in the [TSDoc spec](https://tsdoc.org/pages/spec/overview/):

```ts
/**
 * Use this CodeMirror {@link @codemirror/state#StateField} to get a reference to the EditorView.
 */
```

The rule sees `@codemirror` inside the `{@link}` block and reports it as an unescaped inline tag, suggesting it should be escaped to `\@codemirror`. However, escaping breaks API Extractor / TSDoc resolution, which requires the literal `@` to identify the scoped package name.

## Current workaround

Adding the scope name to `allowedInlineTags`:

```js
'jsdoc/escape-inline-tags': ['error', {
  allowedInlineTags: ['codemirror']
}]
```

This works but is a hack — every new scoped package referenced in `{@link}` targets requires a new entry.

## Expected behavior

The rule should not flag `@scope` as an unescaped inline tag when it appears inside a `{@link}` (or `{@inheritDoc}`, etc.) target, since `@scope/package#member` is valid TSDoc declaration reference syntax.

## Environment

- eslint-plugin-jsdoc: 62.9.0
- Node: 25.9.0
- ESLint: 10.4.0 (flat config)

morgan-coded:
`escape-inline-tags` was reporting scoped npm package names inside declaration-reference inline tags, like `{@link @codemirror/state#StateField}`.
This allows scoped package references inside `link`, `linkcode`, `linkplain`, and `inheritDoc` inline tags while keeping bare `@scope` text and non-reference inline tag bodies reportable. I checked it with the focused rule tests passing 31 tests, generated docs check, changed-file ESLint, TypeScript, full Node 22 coverage at 3,132 passing with 100% coverage, build, and `git diff --check`.
Fixes #1695

### DIFF
--- a/docs/rules/escape-inline-tags.md
+++ b/docs/rules/escape-inline-tags.md
@@ -224,6 +224,11 @@ The following patterns are not considered problems:
  */
 
 /**
+ * A description with a multi-line inline tag {@link
+ * @scope/pkg#Member} reference.
+ */
+
+/**
  * @example
  * Here are some unescaped tags: @yearly, @monthly
  */
--- a/docs/rules/escape-inline-tags.md
+++ b/docs/rules/escape-inline-tags.md
@@ -112,6 +112,30 @@ The following patterns are considered problems:
 // Message: Unexpected inline JSDoc tag. Did you mean to use {@yearly}, \@yearly, or `@yearly`?
 
 /**
+ * Some description @scope and {@link @scope/pkg#Member}
+ */
+// "jsdoc/escape-inline-tags": ["error"|"warn", {"enableFixer":true}]
+// Message: Unexpected inline JSDoc tag. Did you mean to use {@scope}, \@scope, or `@scope`?
+
+/**
+ * Some description {@example @scope/pkg}
+ */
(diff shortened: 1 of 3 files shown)


brettz9:
Sorry, you'd be welcome to submit the PR again if you can get those Co-pilot issues fixed...

morgan-coded:
I rebased and reopened this with both review points addressed.

The multiline declaration-reference case is covered now, the escaper helper is hoisted, and the generated docs are refreshed. The focused rule suite and full test suite are passing.

brettz9:
Sorry, could you rebase again? Your PR is now priority.

morgan-coded:
Rebased and pushed with no other changes; local tests are passing.

brettz9 (approved):
LGTM! Thanks!

PR #1717, opened 2026-06-24
fix(check-template-names): keep commas inside @template default type values

RobinBol:
## Expected behavior

A `@template` whose default value contains a comma inside a type expression, e.g. `@template [T=Record<string, unknown>]`, declares a single template named `T`.

## Actual behavior

`parseClosureTemplateTag` splits the tag on every comma, so the comma inside `Record<string, unknown>` is treated as a separator. This yields a bogus second "template name" (`unknown>]`), producing spurious reports such as `@template unknown>] not in use` (`check-template-names`) and namepath syntax errors (`valid-types`).

## ESLint sample (eslint-plugin-jsdoc 63.0.7)

```js
/** @template [T=Record<string, unknown>] */
function quux() {}
// check-template-names -> "@template unknown>] not in use"
// valid-types          -> "Syntax error in namepath: unknown>]"
```

## Cause

```js
const parseClosureTemplateTag = (tag) => {
  return tag.name
    .split(',')   // splits commas inside Record<string, unknown> too
    .map((type) => type.trim().replace(/^\[?(?<name>.*?)=.*$/v, '$<name>'));
};
```

The `,` split is meant to separate multiple template names (`@template S, T`), but it is applied to the raw string, breaking type expressions in default values.

## Fix

Split only on commas that are not nested within `<`, `(` or `{` (generic arguments, function params, object types). Square brackets are deliberately not counted, since `[...]` is the optional/default wrapper in which commas legitimately separate entries (`[T=string, U=number]` -> `T`, `U` as before).

Affects the three rules that consume `parseClosureTemplateTag`: `check-template-names`, `require-template`, `valid-types`.

Added a `valid-types` test for the `Record<string, unknown>` default; updated the one existing test on the intentionally-malformed input `@template T<~, R` (the reported namepath now includes the full unparseable segment, still flagged as a syntax error); docs regenerated.

### DIFF
--- a/docs/rules/valid-types.md
+++ b/docs/rules/valid-types.md
@@ -409,7 +409,7 @@ parseArray = function(parser) {
     };
 };
 // Settings: {"jsdoc":{"mode":"closure"}}
-// Message: Syntax error in namepath: T<~
+// Message: Syntax error in namepath: T<~, R
 
 /**
  * @template T, R<~
@@ -1009,5 +1009,10 @@ const MY_BITMASK_CONSTANT = {
     BITMASK_VALUE_A: 1 << 4,
     BITMASK_VALUE_B: somePrivateVariableHere
 };
+
+/**
+ * @template [T=Record<string, unknown>]
+ */
+function quux () {}
 ````
 
--- a/src/jsdocUtils.js
+++ b/src/jsdocUtils.js
@@ -1578,6 +1578,42 @@ const isInlineTag = (tag) => {
 */
 
 /**
+ * Splits a `@template` tag's names on commas that separate template entries,
+ * ignoring commas nested within a type expression so that default values such
+ * as `[T=Record<string, unknown>]` are not split apart.
+ *
+ * Only `<`, `(` and `{` are treated as nesting: a comma within them is part of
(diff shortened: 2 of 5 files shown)


brettz9:
Thank you for the PR. It looks like it's working but needs full test coverage.

RobinBol:
Done!

brettz9:
One final thing... I noticed in `src/rules/validTypes.js` that there is this block:

```js
        if (!tryParsePathIgnoreError(
          // May be an issue with the commas of
          //   `utils.parseClosureTemplateTag`, so first try a raw
          //   value; we really need a proper parser instead, however.
          tag.name.trim().replace(/^\[?(?<name>.*?)=.*$/v, '$<name>'),
          mode,
        )) {
          for (const namepath of utils.parseClosureTemplateTag(tag)) {
            validNamepathParsing(namepath);
          }
        }
```

Could you try removing the `replace` (and the comment above it) and see if the tests still pass? I'm wondering if that part of this block is now obsolete due to your introducing a better parser...

RobinBol:
You are right, after removing the `replace` the tests still pass!

PR #1725, opened 2026-07-18
feat: add normalize-see-links rule

### LINKED ISSUE #1631: Rule suggestion: enforce proper `@see` + `@link` syntax for hyperlinks
## Motivation

As of filing, I know of two syntaxes for saying a user can `@see` a link:

* At-Linked: `@see {@link https://example.com|Example}`
* Markdown: `@see [Example](https://example.com)`

For consistency's sake, I think it would be nice to have a lint rule to prefer one and auto-fix the other to it.

## Current behavior

No lint rule enforces either.

## Desired behavior

I'm not sure. The Markdown variant is fewer characters, but the At-Linked one seems to be more of a "proper" link through JSDoc `@link`, so perhaps that?

Both show up well in at least TypeScript's native intellisense. The Markdown variant isn't perfectly supported by VS Code's native syntax highlighter, but that strikes me as a bug and not a motivator for choosing one over the other.

Screenshot of the two syntaxes. The Markdown syntax has a red '(' and cyan 'https:' as a highlighting bug.

## Alternatives considered

🤷

morgan-coded:
Closes #1631 by adding a fixable `normalize-see-links` rule for labeled links in `@see` tags. It converts Markdown `[label](url)` and prefix `[label]{@link url}` links to `{@link url|label}` and leaves that pipe form unchanged. Malformed or delimiter-colliding links, empty labels, code-span content, and scoped-package targets report without a fix. Bare URLs, namepaths, prose, and non-`@see` tags remain no-ops, with bare-URL wrapping deferred as an opt-in because it is "fine as an option if people want it." The default canonical form is pipe, while the `canonicalForm` option also allows prefix if that form is preferred.

### DIFF
--- a/src/rules/normalizeSeeLinks.js
+++ b/src/rules/normalizeSeeLinks.js
@@ -1,0 +1,275 @@
+import iterateJsdoc from '../iterateJsdoc.js';
+
+const markdownLinkRegex = /(?<!\\)\[([^`\]\r\n]+)\]\(([^`\(\)\s]+)\)/gv;
+const prefixLinkRegex = /(?<!\\)\[([^`\]\r\n]+)\]\{@link\s+([^\}\s\|]+)\}/gv;
+const pipeLinkRegex = /\{@link\s+([^\}\s\|]+)\s*\|\s*([^\}\r\n]+)\}/gv;
+
+const markdownLinkAttemptRegex = /(?<!\\)\[[^\]\r\n]*\]\([^\r\n]*(?:\)|$)/v;
+const prefixLinkAttemptRegex = /(?<!\\)\[[^\]\r\n]*\]\{@link[^\}\r\n]*(?:\}|$)/v;
+const pipeLinkAttemptRegex = /\{@link[^\}\r\n]*\|[^\}\r\n]*(?:\}|$)/v;
+
+const scopedPackageNameRegex = /^@[\w.\-]+\/[\w.\-]+/v;
+const markdownCodeSpanRegex = /(?<!\\)(`+)(?!`)[\s\S]*?(?<!`)\1(?!`)/gv;
+
+/**
+ * @param {'pipe'|'prefix'} canonicalForm
+ * @param {string} label
+ * @param {string} target
+ * @returns {boolean}
+ */
+const cannotSafelyFormatLink = (canonicalForm, label, target) => {
+  if (!label.trim()) {
+    return true;
+  }
+
+  if (canonicalForm === 'pipe') {
+    return target.includes('|') || /[\}\|]/v.test(label);
+  }
+
+  return label.includes(']') || /[\}\|]/v.test(target);
+};
+
+/**
+ * @param {string} description
+ * @param {number} index
+ * @returns {boolean}
+ */
+const isInsideMarkdownCodeSpan = (description, index) => {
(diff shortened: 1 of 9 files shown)


test/rules/assertions/normalizeSeeLinks.js
```diff
@@ -0,0 +1,718 @@
...
+  parser as typescriptEslintParser,
+} from 'typescript-eslint';
+
+export default {
+  invalid: [
+    {
+      code: `
+        /**
+         * @see [Docs](https://api.example.com/v1?ids=1|2|3)
+         */
```
brettz9, line 11: I'm wondering if we should perhaps allow it to be run through `encodeURI` so it can be safely normalized... It may otherwise be frustrating to be left with a form that some users might not know how to fix.

I wondered if `\|` would work, but Gemini says no. Gemini instead suggests space-delimiting after the URL like this: `{@link https://api.example.com/v1?ids=1|2|3 Docs}`. But rather than introducing yet another form, perhaps `encodeURI` is the easiest.

test/rules/assertions/normalizeSeeLinks.js
```diff
@@ -0,0 +1,718 @@
...
+          message: '@see link cannot be safely normalized.',
+        },
+      ],
+      output: null,
+    },
+    {
+      code: `
+        /**
+         * @see [a}b](https://example.com)
+         */
```
brettz9, line 45: In this case, rather than `encodeURI`, since we're dealing with the text portion, it appears that `\}` is the acceptable escape sequence per https://jsdoc.app/about-block-inline-tags .

src/rules/normalizeSeeLinks.js
```diff
@@ -0,0 +1,275 @@
...
+      description: 'Normalizes labeled links in `@see` tags to a canonical `{@link}` form.',
+      url: 'https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/normalize-see-links.md#repos-sticky-header',
+    },
+    fixable: 'code',
+    schema: [
+      {
+        additionalProperties: false,
+        properties: {
+          canonicalForm: {
+            description: 'The canonical `{@link}` form. Defaults to `"pipe"`.',
```
brettz9: I think we should outline a sample of what the formats here ("pipe" and "prefix") look like.

brettz9 (changes requested):
It's looking good to me. If you could just address the comments. I'd hope we can avoid unresolveable errors.

morgan-coded:
Addressed all three. URL-side delimiter collisions now autofix through `encodeURI` (existing `%HH` escapes are preserved so they aren't double-encoded), and the `canonicalForm` option now shows a sample of each form.

On the label side, when I tried the `\}` escape from the spec I found `{@link https://example.com|a\}b}` came back with the label truncated to `a\`, and I couldn't get `[a\]b]{@link ...}` to parse at all (jsdoccomment 0.88.0). So rather than emit a fix that doesn't round-trip, I left those two report-only for now. The safe label cases do autofix: extra `|` in the text, a literal `}` in a prefix label, and a bare `{` in a pipe label all parse back cleanly. Happy to enable `\}`/`\]` if those become parseable later. Empty labels stay report-only since there's nothing to normalize.

brettz9 (approved):

brettz9:
LGTM, thanks for the PR!

PR #1732, opened 2026-07-24
feat(check-param-names): add order fixer and suggestions

### LINKED ISSUE #47: `jsdoc/check-param-names` should be fixable if order is wrong
I saw that another eslint plugin detects merely incorrect parameter order. It would be nice if this could be automatically fixed. You would have to ensure object properties are kept with param, and it could also (if this is preferred) move the `@returns` to the end.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/50732118-jsdoc-check-param-names-should-be-fixable-if-order-is-wrong?utm_campaign=plugin&utm_content=tracker%2F23037809&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23037809&utm_medium=issues&utm_source=github).
</bountysource-plugin>

morgan-coded:
`check-param-names` reports out-of-order flat `@param` names as a general name mismatch and cannot put their whole blocks in signature order. Following [brettz9's confirmed direction](https://github.com/gajus/eslint-plugin-jsdoc/issues/47#issuecomment-5071725925), this distinguishes the order case and, with `enableFixer`, reorders whole non-nested tag blocks while preserving multiline descriptions and trailing text attached to the last tag; destructured and nested parameters remain report-only in this first pass. `badParamOrder` and `duplicateParams` default to `true` and can disable their reports individually, while `extraParams` and `badParamNames` default to `false` and opt into removal and rename suggestions instead of automatic fixes. Suggestion support is new to the shared JSDoc reporter; when the two opt-in options are off, the existing reports stay unchanged. Fixes #47

### DIFF
--- a/.README/rules/check-param-names.md
+++ b/.README/rules/check-param-names.md
@@ -7,11 +7,22 @@ the function declaration.
 
 ## Fixer
 
-Auto-removes `@param` duplicates (based on identical names).
+With `enableFixer`, reorders non-nested `@param` definitions whose names match
+the function signature but appear in a different order. The whole tag block is
+moved, including multiline descriptions and any text after the final tag.
+Destructured and nested parameters are reported without an automatic fix.
 
-Note that this option will remove duplicates of the same name even if
-the definitions do not match in other ways (e.g., the second param will
-be removed even if it has a different type or description).
+The fixer also auto-removes `@param` duplicates (based on identical names).
+This can be disabled with `duplicateParams`. Duplicate removal does not compare
+the rest of the definitions, so the second tag is removed even if it has a
+different type or description.
+
+## Suggestions
+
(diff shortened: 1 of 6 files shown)


brettz9:
I think with the suggestions (`extraParams` and `badParamNames`), we can just automatically apply them since the user can ignore suggestions, no? I intended the flags to be for auto-applying the *fixer* instead of adding a suggestion.

morgan-coded:
Made the suggestions always-on and moved the two flags to gate the autofix instead — `extraParams` and `badParamNames` now always attach their suggestions, and enabling a flag auto-applies the corresponding fix, defaulting off. Docs and tests cover both flag states.

brettz9 (approved):
LGTM! Nice work!

PR #1734, opened 2026-07-25
feat(normalize-see-links): autofix delimiter-colliding labels via escape round-trip

morgan-coded:
Follow-up to #1725, which kept delimiter-colliding labels report-only because jsdoccomment 0.88.0 could not round-trip `\}` in pipe labels or `\]` in prefix labels. es-joy/jsdoccomment#26 (shipped in 0.89.0; this repo is on 0.90.0) made escaped inline-tag text round-trip, so the fixer now emits those labels with minimal escaping instead of only reporting: source-form escapes are decoded first, then the destination form escapes its own closing delimiter, and a backslash is doubled only where it would otherwise merge with a following backslash, the delimiter, or the end of the label. Pre-escaped labels normalize without double escaping, and the new fixtures assert fixpoint behavior (a second pass fixes and reports nothing) across pipe, prefix, mixed-delimiter, and trailing-backslash cases. Empty labels, space-format tags, code-span content, scoped-package targets, and malformed attempts keep their existing report-only behavior.

### DIFF
--- a/.README/rules/normalize-see-links.md
+++ b/.README/rules/normalize-see-links.md
@@ -22,9 +22,6 @@ For prefix form, the fixer escapes `]` in the label:
 /** @see [A\]B]{@link https://example.com} */
 ```
 
-This behavior relies on `@es-joy/jsdoccomment` 0.89.0 or later, which
-round-trips context-specific escaped label delimiters.
-
 ## Options
 
 {"gitdown": "options"}
--- a/.README/rules/normalize-see-links.md
+++ b/.README/rules/normalize-see-links.md
@@ -7,6 +7,24 @@ only a lowercase `http:` or `https:` URL in a plain, no-label `{@link}` tag.
 This option defaults to `false`, and its output is independent of
 `canonicalForm` because there is no label to position.
 
+Labels that contain the destination form's closing delimiter are fixable. For
+pipe form, the fixer escapes `}` in the label:
+
+```js
+/** @see [See {options}](https://example.com/api) */
+/** @see {@link https://example.com/api|See {options\}} */
+```
+
+For prefix form, the fixer escapes `]` in the label:
+
+```js
+/** @see {@link https://example.com|A]B} */
+/** @see [A\]B]{@link https://example.com} */
+```
+
+This behavior relies on `@es-joy/jsdoccomment` 0.89.0 or later, which
+round-trips context-specific escaped label delimiters.
+
 ## Options
 
 {"gitdown": "options"}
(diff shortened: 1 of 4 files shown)


.README/rules/normalize-see-links.md
```diff
@@ -7,6 +7,24 @@ only a lowercase `http:` or `https:` URL in a plain, no-label `{@link}` tag.
...
+
+For prefix form, the fixer escapes `]` in the label:
+
+```js
+/** @see {@link https://example.com|A]B} */
+/** @see [A\]B]{@link https://example.com} */
+```
+
+This behavior relies on `@es-joy/jsdoccomment` 0.89.0 or later, which
+round-trips context-specific escaped label delimiters.
```
brettz9: Let's just remove these last 3 lines (including the empty line) of implementation details.

brettz9 (changes requested):
Besides the doc changes, LGTM!

Just to double-check though... Are there any backslash sequences we can use in place of `encodeURI` which don't cause problems for TS/VSC? Of course, a backslash sequence is more visually elegant than a percent-encoded sequence.

morgan-coded:
Removed the three doc lines and regenerated `docs/`.

On backslash sequences in place of `encodeURI`: I tested `\}` in the URL position (pipe and prefix forms) against both consumers and couldn't find a sequence that survives.

- jsdoccomment 0.90.0 round-trips escapes only in label context — in `parseInlineTags` the label groups accept escape pairs, but `namepathOrURL` is `[^\}\s\|]*` — so `{@link https://example.com/a\}b|Label}` parses with `namepathOrURL` `https://example.com/a\` and the label is lost.
- TypeScript (5.9.3) does no escape processing inside the link tag at all: the link node ends at the first `}`, so the same input becomes a truncated link ending in a backslash, with `b|Label}` leaking into plain hover text; the percent-encoded form stays one intact link node.
- Even with parser support on both sides, the decoded target would still hold a raw `}`, which RFC 3986 requires percent-encoding in a URI anyway.

Agreed the backslash form reads better — the labels use it now that 0.89.0 round-trips them. The URL looks like the one position where percent-encoding is the native escape; `encodeURI` here also preserves already-percent-encoded input rather than double-encoding it.

brettz9:
Ok, thanks.

I think we could look to replace `encodeURI` with just the following (preserving %XX escapes as is):

- unescaped % to %25
- `|` to %7C
- `{` to %7B
- `}` to %7D

But this is good for now.

brettz9 (approved):

PR #1735, opened 2026-07-26
fix(normalize-see-links): minimally escape link targets instead of encodeURI

morgan-coded:
Implements your suggestion in [#1734 (comment)](https://github.com/gajus/eslint-plugin-jsdoc/pull/1734#issuecomment-5082670960): `encodeLinkTarget` now applies only the four fixed mappings — unescaped `%` → `%25` (existing `%XX` escapes preserved as-is, case-insensitive), `|` → `%7C`, `{` → `%7B`, `}` → `%7D` — instead of `encodeURI`, so characters that never collide with `{@link}` parsing (non-ASCII, `^`, `"`, …) stay readable in autofixed targets.

The unescaped-`%` pass runs before the delimiter mappings, so the `%` introduced by `%7C`/`%7B`/`%7D` is never re-escaped.

All existing delimiter and percent-preservation assertions pass unchanged; two new cases pin the delta (literal `ä` and `^` preserved through the autofix). Full suite 3282 passing at 100% coverage; `check-docs` clean — no generated-doc changes.

### DIFF
--- a/src/rules/normalizeSeeLinks.js
+++ b/src/rules/normalizeSeeLinks.js
@@ -47,7 +47,11 @@ const escapeLinkLabel = (canonicalForm, label, sourceForm) => {
  * @returns {string}
  */
 const encodeLinkTarget = (target) => {
-  return encodeURI(target).replaceAll(/%25(?=[\dA-F]{2})/giv, '%');
+  return target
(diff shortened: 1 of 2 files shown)


brettz9 (approved):
Great, thanks!  Plan to merge tomorrow, as we try not to add releases more than once a day.

PR #1738, opened 2026-07-30
feat: ESM only

### LINKED ISSUE #1626: Go ESM only?
## Motivation

Smaller package size, significantly simpler packaging story.

## Current behavior

eslint-plugin-jsdoc ships both ESM & CJS sources.

## Desired behavior

eslint-plugin-jsdoc would only ship ESM sources.

## Additional info

Our engines already support `require(ESM)`, so this shouldn't be breaking at all, but it'd probably be wise to release a major.

## Alternatives considered

Status quo

brettz9:
feat: ESM only; fixes #1626

BREAKING CHANGE:

- Requires Node ^22.22.2 || >=24.15.0
- Drops CJS exports; should be able to use `require(esm)` with newly supported Node versions


package.json
```diff
@@ -88,38 +84,20 @@
     "*.js": "eslint --fix"
   },
   "type": "module",
-  "main": "./dist/index.cjs",
+  "main": "./src/index.js",
```
sxzz: We're not supporting Node.js 12.6 and below, so we can also remove the `main` field as all tooling would read from "exports" only and skip "main".

See also: https://github.com/publint/publint/issues/24

PR #1753, opened 2026-08-31
fix(`convert-to-jsdoc-comments`): handle "any" context and avoid extra line break when there is already multiline text

### LINKED ISSUE #1747: 64.2.1: `convert-to-jsdoc-comments` ignores `contexts: ['any']`?
Search https://github.com/gajus/eslint-plugin-jsdoc/issues?q=contexts revealed some related discussion but nothing on this specific rule.

I'm trying to avoid hardcoding explicit node type arrays, when I really want literally "everything" to match with `contexts: ['any']`.

Surprisingly, `contexts: ['*']` works to hit everything, but has a side-effect of duplicate hits per site.

EDIT `:not(Program)` selector is somewhat less bad, maybe useful as a data point?

What am I missing?

## Expected behavior
I would expect a regular block comment above `VariableDeclaration` node type to get flagged

## Actual behavior
Block comment above `VariableDeclaration` node type does NOT get flagged

## ESLint Config
```
import jsdoc from 'eslint-plugin-jsdoc';

...

{
    ...
    plugins: {
      jsdoc,
    },
    rules: {
      'jsdoc/convert-to-jsdoc-comments': [
        2,
        {
          contexts: [ 'any' ],
          lineOrBlockStyle: 'block',
        },
      ],
    }
}
```

## ESLint sample

```js
/*
 * Seniority levels that participate in distribution, in display
 * order. `custom` is never a real distribution key. Every test shown in this
 * modal can recruit each of these levels, so all three are always editable.
 */
const SENIORITY_ORDER = ['senior', 'middle', 'specialist'];
```

## Environment

- Node version: 24.19
- ESLint version: 9.39.4
- `eslint-plugin-jsdoc` version: 64.2.1

### LINKED ISSUE #1748: 64.2.1: `no-blank-block-descriptions` fails to remove a leading empty line appearing before description content
Related #982 

## Expected behavior
Remove leading empty line appearing before description content 

## Actual behavior
Empty line stays

## ESLint Config
```js
import jsdoc from 'eslint-plugin-jsdoc';

{
    plugins: {
      jsdoc,
    },
    rules: {
      'jsdoc/no-blank-block-descriptions': 2,
    }
}
```

## ESLint sample

```js
/**
 *
 * Seniority levels that participate in distribution, in display
 * order. `custom` is never a real distribution key. Every test shown in this
 * modal can recruit each of these levels, so all three are always editable.
 */
const SENIORITY_ORDER = ['senior', 'middle', 'specialist'];
```

For some reason I don't see this case tested at https://github.com/gajus/eslint-plugin-jsdoc/blob/v64.2.1/docs/rules/no-blank-block-descriptions.md

Side quest: this is a result of `convert-to-jsdoc-comments` auto-fix, if there's bandwidth to look at that.

> The fixer unconditionally prepends /**\n${indent}* to the raw, unmodified comment.value. 
> comment.value is literally "\n * Seniority levels...\n */" (i.e. it already starts with \n * ). So the fixer glues /**\n *  + \n * Seniority levels... together — producing two newline+* sequences back to back, which is exactly the empty \n *\n line you're seeing. It never strips/dedupes the original leading \n * before re-prefixing its own.

## Environment

- Node version: 24.19
- ESLint version: 9.39.4
- `eslint-plugin-jsdoc` version: 64.2.1

brettz9:
fix(`convert-to-jsdoc-comments`): handle "any" context and avoid extra line break when there is already multiline text; fixes #1747

Also:
- fix(`no-blank-block-descriptions`): strip out initial line breaks when text is present (with or without tags); fixes #1748

### DIFF
--- a/docs/rules/convert-to-jsdoc-comments.md
+++ b/docs/rules/convert-to-jsdoc-comments.md
@@ -216,6 +216,15 @@ class TestClass {
 var a = []; // Test comment
 // "jsdoc/convert-to-jsdoc-comments": ["error"|"warn", {"contextsBeforeAndAfter":["VariableDeclarator"]}]
 // Message: Line comments should be JSDoc-style.
+
+/*
+ * Seniority levels that participate in distribution, in display
+ * order. `custom` is never a real distribution key. Every test shown in this
+ * modal can recruit each of these levels, so all three are always editable.
+ */
+const SENIORITY_ORDER = ['senior', 'middle', 'specialist'];
+// "jsdoc/convert-to-jsdoc-comments": ["error"|"warn", {"contexts":["any"],"lineOrBlockStyle":"block"}]
+// Message: Block comments should be JSDoc-style.
 ````
 
 
--- a/docs/rules/no-blank-block-descriptions.md
+++ b/docs/rules/no-blank-block-descriptions.md
@@ -48,6 +48,23 @@ function functionWithClearName(x) {}
  */
 function functionWithClearName() {}
 // Message: There should be no extra blank lines in block descriptions not followed by tags.
+
+/**
+ *
+ * Some text
+ * @param {number} x
+ */
+function functionWithClearName(x) {}
+// Message: There should be no blank lines in block descriptions followed by tags.
+
+/**
+ *
+ * Seniority levels that participate in distribution, in display
+ * order. `custom` is never a real distribution key. Every test shown in this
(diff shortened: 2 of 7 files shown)


lkraav (commented):
🦸‍♂️ 

I see `any` is now equalized with `:not(Program)` at core.

It probably has some side-effects, but thus far in my config it has worked well, so maybe it's the endgame move after all.

PR #1755, opened 2026-09-01
fix(`require-asterisk-prefix`): avoid crashing on custom languages

### LINKED ISSUE #1754: `require-asterisk-prefix` crashes when enabled for Markdown files
## Expected behavior

When `jsdoc/require-asterisk-prefix` is enabled for a Markdown file using ESLint’s native Markdown language, the rule should safely ignore the incompatible syntax tree rather than crash ESLint.

The rule requires JavaScript-specific token and comment APIs, which a Markdown `SourceCode` implementation does not provide.

## Actual behavior

ESLint terminates with the following error:

```text
TypeError: sourceCode.getTokenBefore is not a function
Occurred while linting example.md:1
Rule: "jsdoc/require-asterisk-prefix"
    at findJSDocComment (...)
    at getJSDocComment (...)
    at *:not(Program) (.../src/iterateJsdoc.js:2362:31)
```

This also break tools such as [XO](https://github.com/xojs/xo) when a configuration enables the rule for Markdown files (XO uses this plugin through [eslint-config-xo](https://github.com/xojs/eslint-config-xo)).

## ESLint Config

```js
import markdown from '@eslint/markdown';
import jsdoc from 'eslint-plugin-jsdoc';

export default [
  {
    files: ['**/*.md'],
    language: 'markdown/gfm',
    plugins: {
      jsdoc,
      markdown,
    },
    rules: {
      'jsdoc/require-asterisk-prefix': 'error',
    },
  },
];
```

## ESLint sample

`example.md`:
```md
# Heading

Text.
```

Running ESLint against this file causes the exception even though it contains no JSDoc comments.

## Environment

* Node version: v24.19.0
* ESLint version: v10.8.1
* `eslint-plugin-jsdoc` version: v64.3.3
* `@eslint/markdown` version: v8.0.3

oscarmarcelo:
Closes #1754.

Prevents `require-asterisk-prefix` from crashing when it is enabled for a non-JavaScript language such as Markdown.

Native ESLint languages provide their own `SourceCode` implementations. The Markdown implementation does not expose the JavaScript-specific `getTokenBefore()` method, but the shared `iterateAllJsdocs` visitor previously called it for every non-`Program` syntax node.

Because the guard is in the shared `iterateAllJsdocs` path, it also prevents the same underlying failure in other rules that use this traversal mode.

### DIFF
--- a/src/iterateJsdoc.js
+++ b/src/iterateJsdoc.js
@@ -2338,6 +2338,13 @@ const iterateAllJsdocs = (iterator, ruleConfig, contexts, additiveCommentContext
         // @ts-expect-error ESLint < 10
         sourceCode = context.getSourceCode(),
       } = context;
+
+      // Custom languages provide their own `SourceCode` implementations.
+      // JSDoc traversal requires the JavaScript token API.
+      if (typeof sourceCode.getTokenBefore !== 'function') {
+        return {};
+      }
+
       settings = getSettings(context);
       if (!settings) {
         return {};
--- a/test/iterateJsdoc.js
+++ b/test/iterateJsdoc.js
@@ -85,6 +85,20 @@ describe('iterateJsdoc', () => {
       });
     });
   });
+  describe('custom language compatibility', () => {
+    it('does not iterate without the JavaScript token API', () => {
+      const rule = iterateJsdoc(() => {}, {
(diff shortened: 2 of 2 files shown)


brettz9:
LGTM, thanks!

PR #1759, opened 2026-09-09
fix(`getJsdocProcessorPlugin`): key positions by file; drop suggestions

morgan-coded:
Editor positions after the first file pointed at an earlier file's example block, and leaked example suggestions could damage the example tag.

The processor now keeps block positions in a filename-keyed `Map` and drops example suggestions alongside fixes, while host-file fixes and suggestions remain available.

The position and suggestion regressions fail before the fix and pass with it; the full suite passes at the existing 100% coverage gate.

The proper fix is mapping example fix and suggestion ranges back to the host file; I can take that as a follow-up if wanted.

### DIFF
--- a/src/getJsdocProcessorPlugin.js
+++ b/src/getJsdocProcessorPlugin.js
@@ -151,16 +151,22 @@ export const getJsdocProcessorPlugin = (options = {}) => {
   }
 
   /**
-   * @type {{
+   * @typedef {{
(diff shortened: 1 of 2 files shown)


brettz9:
We generally do only one release per day, so if you wanted to build on this PR to get it out in the same day, let me know. Otherwise, I'll just go ahead and release for today.

PR #1769, opened 2026-09-15
fix: specify typescript as an optional peer dependency

### LINKED ISSUE #1768: Cannot disable erroneous "typescript is not installed" warning
## Expected behavior

* `jsdoc/no-unnecessary-type-assertion` does not give the "typescript is not installed" warning if typescript is installed.
* Disabling `jsdoc/no-unnecessary-type-assertion`, as documented, should disable the warning.

## Actual behavior

* `eslint-plugin-jsdoc` relies on hoisting to find `typescript`, rather than declaring it as an optional peer dependency, causing it to fail to find it in some environments.
* The warning is issued as soon as the package is imported, even if the plugin isn't enabled in the config at all.

## ESLint Config

As this depends on hoisting and the structure of a monorepo, I've created a minimal repo: [test.zip](https://github.com/user-attachments/files/32254287/test.zip)

## ESLint sample

* Unzip [test.zip](https://github.com/user-attachments/files/32254287/test.zip)
* Run `pnpm install && pnpm exec eslint` in the extracted `test/` directory.

## Environment

- Node version: v24.21.0
- ESLint version v10.10.0
- `eslint-plugin-jsdoc` version: 64.5.0

brettz9:
fix: specify typescript as an optional peer dependency; fixes #1768


package.json
```diff
@@ -123,7 +123,13 @@
     "statements": 100
   },
   "peerDependencies": {
-    "eslint": "^7.0.0 || ^8.0.0 || ^9.0.0 || ^10.0.0"
+    "eslint": "^7.0.0 || ^8.0.0 || ^9.0.0 || ^10.0.0",
+    "typescript": ">= 6.0.0"
```
SchroederSteffen, line 127: This is a breaking change for anyone on typescript <6.
- https://github.com/gajus/eslint-plugin-jsdoc/issues/1768#issuecomment-5695349031

Why does it need `6`?
Could `*` be enough or another somewhat low version?

brettz9: @SchroederSteffen : I've set the version to any with another patch release (v64.5.2) (per #1770 ). I don't know the version that corresponds with our API usage, but this will at least hopefully avoid the breaking aspect of the change.

SchroederSteffen: perfect, thanks!
I couldn't find a versioned TS API doc either. 🙁

